PR #5, opened 2020-04-21
pull down the SDK binary for the architecture on which it is going to be run

jmazzitelli:
This is mainly here to help the ARM support effort going on right now.


jmazzitelli:
closing this in favor of https://github.com/kiali/kiali-operator/pull/6 which has the same change but also runs a travis build to show that the operator image can be built on both archs.

PR #12, opened 2020-04-23
disable arm64 travis test

jmazzitelli:
the comment being committed provides a description of what this is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,10 @@ os: linux
 
 arch:
   - amd64
-  - arm64
+  # It takes a very long time to spin up an ARM machine on travis.
+  # Because of this, it takes a long time before we can merge PRs.
+  # So this is disabled for now. We can re-enable if we need this.
+  #- arm64
 
 language: python
 addons:


jmazzitelli:
> I guess this is fine although what, now we have to remember to turn this on "occasionally"?

I suppose so. I don't have access to an ARM machine to periodically run it locally.

But realize that this test is NOT testing anything at runtime (i.e. it is not like we are running molecule or other integration tests here). This only tests that the Operator SDK has an ARM build that can be used to BUILD the operator on an ARM machine. So, it is a nice test, but is no where near anything like a comprehensive test since it doesn't even run the operator to see that it works on ARM.

.travis.yml
```diff
@@ -3,7 +3,10 @@ os: linux
 
 arch:
   - amd64
-  - arm64
+  # It takes a very long time to spin up an ARM machine on travis.
+  # Because of this, it takes a long time before we can merge PRs.
+  # So this is disabled for now. We can re-enable if we need this.
+  #- arm64
```
josejulio, line 9: Someone could mark this arm64 as [allow_failures](https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail) and setup travis to [fast-finishing](https://docs.travis-ci.com/user/build-matrix/#fast-finishing). In that way, the build will report success without waiting for this arm64 build to finish.

jshaughn: cc @jmazzitelli ^

jmazzitelli: that sounds like a fine PR someone in the community can contribute if they want this feature.

PR #13, opened 2020-04-24
Adding config flag for istio status feat in the masthead

xeviknal:
This task is to complement the new flag introduced in https://github.com/kiali/kiali/pull/2598

Once the flag is set to false, the istio components check feat won't be performed. Default value is set to true.

Needs https://github.com/kiali/kiali/pull/2598

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -429,12 +429,14 @@ spec:
 #      url: ""
 #
 # **Istio-specific settings:
+# istio_status_enabled: Enable/Disable of istio component status into masthead indicator. It defaults to true.
 # istio_identity_domain: The annotation used by Istio to identify domains.
 # istio_sidecar_annotation: The pod annotation used by Istio to identify the sidecar.
 # url_service_version: The Istio service used to determine the Istio version.
 #                      If empty, assumes it is in the istio namespace at the URL "http://istio-pilot.<istio_namespace>:8080/version"
 #    ---
 #    istio:
+#      istio_status_enabled: true
 #      istio_identity_domain: "svc.cluster.local"
 #      istio_sidecar_annotation: "sidecar.istio.io/status"
 #      url_service_version: ""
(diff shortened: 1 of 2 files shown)


lucasponce:
I think is missing the defaults value, where all values should be explicitly listed.

But @jmazzitelli can confirm here if there is any change in that.

xeviknal:
@jmazzitelli besides the config.go, where are the default values?

I have this one set in here: https://github.com/kiali/kiali/pull/2598/files/ee1b79cc092840b5dee07c0f93d1c64f04955aaf#diff-3baf47c64847a8fb8aaa8cc2e088513bR287

Thanks

lucasponce:
I think it's this:

https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml

lucasponce:
Thanks @xeviknal I guess that's the missing one, as this is new and it shouldn't need to be added in other place.

But I think @jmazzitelli should confirm because probably I'm missing other checks (OLM/Manifest or perhaps other duplication of this in some yaml that I don't remember right now).

deploy/kiali/kiali_cr.yaml
```diff
@@ -431,10 +431,12 @@ spec:
 # **Istio-specific settings:
 # istio_identity_domain: The annotation used by Istio to identify domains.
 # istio_sidecar_annotation: The pod annotation used by Istio to identify the sidecar.
+# istio_status_enabled: Enable/Disable of istio component status into masthead indicator. It defaults to true.
 # url_service_version: The Istio service used to determine the Istio version.
 #                      If empty, assumes it is in the istio namespace at the URL "http://istio-pilot.<istio_namespace>:8080/version"
 #    ---
 #    istio:
+#      istio_status_enabled: true
```
jmazzitelli, line 439: keep this list alphabetized - move it down to line 441 (after istio_sidecar_annotaiton).

jmazzitelli:
FYI: If this needs to be backported to the production (1.12) builds, you also need to update the playbooks/roles under the v1.12 folder.

lucasponce:
@xeviknal this feature doesn't need to be backported to 1.12 unless I missed something.

I mean, it's going to add something in the config that won't have a counterpart in the 1.12 branch.

PR #14, opened 2020-04-24
Document that query params are allowed in grafana URL

jotak:
Depends on https://github.com/kiali/kiali/pull/2687


jotak:
DNM: to merge only when dependent is merged

PR #16, opened 2020-04-27
Truncate version label to the k8s maximum of 63 when appropriate.

jmazzitelli:
This will be required when the version label is based on an image version that is a SHA.

backport PR: https://github.com/kiali/kiali-operator/pull/17


jmazzitelli:
This now has a molecule test showing the truncation working.

PR #17, opened 2020-04-27
[v1.12] Truncate version label to the k8s maximum of 63 when appropriate.

jmazzitelli:
This will be required when the version label is based on an image version that is a SHA.


jmazzitelli:
molecule test added and shows it working

PR #18, opened 2020-04-27
Truncate version label using the old jinja API

jmazzitelli:
Because we need to use the old jinja API, we have to truncate to 60.
The trailing "XXX" adds three more to make the k8s maximum of 63.
We can't use elipses ("...") because k8s label values must end with an alphanum char.


jmazzitelli:
backport: https://github.com/kiali/kiali-operator/pull/19

PR #22, opened 2020-05-01
convert to new olm bundle format and use the new RELATED_IMAGE_xyz env var

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2031

Wait until OLM can support this new stuff before merging - should happen the week of May 18.
UPDATE: going to wait until end of next week when full support will be available (by May 29)
UPDATE2: won't be until end of June at the earliest
UPDATE3: This should be ready now (July 6, 2020). Going to ask to get this merged now.


jmazzitelli:
not sure why I put the backport label on this. I don't think this requires backporting to the v1.12 branch. but if I remember why, I'll put the label back on :)

jmazzitelli:
I added this to the current sprint - this should be able to be completed soon.

jmazzitelli:
> not sure why I put the backport label on this. I don't think this requires backporting to the v1.12 branch. but if I remember why, I'll put the label back on :)

Now I remember :) The master branch has added a bunch of things that are not supported in the Kiali v1.12 server such as iter8 and other things. So I need to backport these changes to the v1.12 branch.

PR #26, opened 2020-05-04
don't use jinja api - use array notation for easier and more predictable results

jmazzitelli:


jmazzitelli:
I was seeing odd behavior with this truncate function (and we can't even use the new version of that API anyway). Then I realized we don't need to use the jinja truncate function - using just array notation does what we need. This fixes the molecule test also - shows this passing.

jmazzitelli:
@mattmahoneyrh this needs approval - it fixes that problem we had where the SHA is used as a label but is too long (over 63 chars) for k8s labels and needs to be truncated. I need to change the implementation of this. I was seeing the molecule test failing for some reason. I'm removing the use of that truncate API (there are two versions of that, depending on the base ansible operator image in use - and the behavior is different between the two). Instead I am just using a simpler array notation to truncate the string. The molecule test is now passing again.

PR #31, opened 2020-05-07
Provide option to create operator namespace or not.

jmazzitelli:
This is needed if, say, you want to install directly inside istio-system - see https://github.com/istio/istio/pull/22762


jmazzitelli:
> I guess you can't just use it if it exists and otherwise create it, Fine.

No, because this is a standalone script with no requirement that a cluster is running (or that you even have access to one or that you even have "oc" or "kubectl" in PATH.

PR #33, opened 2020-05-09
pilot version endpoint is gone - use istiod now

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2754

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 molecule/maistra-e2e/assets/secret.yaml
 _output
+
+# Do not commit any compiled Ansible filter plugins
+*.pyc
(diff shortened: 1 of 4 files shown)


jmazzitelli:
Ran this with a 1.6 dev build, and it is working. This should be backwards compatible with Istio 1.5, as well. Who ever reviews this should check that by running this PR with Istio 1.5.

image

jmazzitelli:
ConfigMap will now have this when running with Istio 1.6+:

```
data:
  config.yaml: |
  ...
    external_services:
      ...
      istio:
        url_service_version: http://istiod.istio-system:15014/version
```

PR #34, opened 2020-05-10
make sure accessible namespaces is correct if kiali not installed in istio-system

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2756

[this backend PR](https://github.com/kiali/kiali/pull/2757) needs to go with this - that backend code change is just for some Makefile enhancements so you can test installing Kiali somewhere other than istio-system.

### DIFF
--- a/deploy/kiali/kiali_cr_dev.yaml
+++ b/deploy/kiali/kiali_cr_dev.yaml
@@ -5,6 +5,7 @@ metadata:
   annotations:
     ansible.operator-sdk/verbosity: "1"
 spec:
+  istio_namespace: ${ISTIO_NAMESPACE}
   version: "${KIALI_CR_SPEC_VERSION}"
   auth:
     strategy: $AUTH_STRATEGY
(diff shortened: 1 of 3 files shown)


jmazzitelli:
This will require a kiali/kiali PR just to change some makefiles to make it easy to test installing kiali in a  namespace that is not the istio control  plane.

jmazzitelli:
To test, Kiali CR needs 4 things:

1) deployment.namespace needs to be some other namespace other than istio-system
2) deployment.accessible_namespaces needs to be something other than ** (an empty list is fine)
3) istio_namespace needs to be set to "istio-system" (assuming that is your control plane namespace)
4) auth.strategy should be something that will use the Kiali service account (`anonymous` or `login`)

If you have the kiali/kiali repo cloned, you can run this make command to create such a Kiali CR:

```
kubectl create ns testns; ACCESSIBLE_NAMESPACES="" AUTH_STRATEGY="anonymous" NAMESPACE="testns" make kiali-create
```

jmazzitelli:
@mattmahoneyrh I made you are reviewer because this might be something you want to add to the test suite you have. This fixes a problem with someone wants to install Kiali in a namespace other than the control plane namespace (e.g. istio-system). This isn't an issue with Service Mesh/Maistra because that isn't how Kiali is deployed, but it is allowed by the operator so the community can be deploying Kiali in that fashion. So we should probably test for this to catch any other issues that might occur.

PR #37, opened 2020-05-15
deploy script can allow for updates for idempotency

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2561

This introduces a new option to the deploy script (`--allow-updates true` aka `-au true`). When set to `true` this option allows the script to continue even if an operator or a Kiali instance is found to already be installed.

This should work under most circumstances, but I hesitate to say this will work 100% of the time, particularly if the YAMLs change between invocations of the script (there could be changes made to, say, a Deployment yaml that may not be conducive to updates). In the cases where an update does not work (for example, across different versions it may not work), this is not to be considered a bug - you simply must run the script without `--allow-updates true` and let the script uninstall the old installation and re-install the new one.

### DIFF
--- a/deploy/deploy-kiali-operator.sh
+++ b/deploy/deploy-kiali-operator.sh
@@ -25,6 +25,16 @@
 # -----------
 # Environment variables that affect the overall behavior of this script:
 #
+# ALLOW_UPDATES
+#    This script's purpose is to be an installer for the operator. It may not be able to
+#    update existing operator or Kiali instances. This is why the default behavior of this
+#    script is to uninstall existing installations before attempting to install new ones.
+#    However, you can force this script to attempt to update existing installations by setting
+#    this to true. When true, any existing operator or Kiali instances will remain and this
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Because this updates the deploy script (a public-facing script that gets released as soon as it is merged to master) a good review is needed to make sure this doesn't break anything.

jmazzitelli:
I will put the DNM label on this PR - I don't want this going in this sprint (which ends today). I'll postpone merging to give time for testing and review. We'll include it next sprint.

jmazzitelli:
I'm going to log this only because it was a question @mattmahoneyrh had...

Note that you CANNOT expect to uninstall/reinstall only the operator and expect Kiali to remain untouched as-is. When you upgrade an operator (even if no changes are made), the operator upon restart will perform initial processing of any existing Kiali CR - at this point, it may be that the Kiali deployment itself will change or redeploy its pod.

PR #38, opened 2020-05-15
correct value for ansible boolean is "yes"

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2792

Grrr... Ansible booleans require "yes" instead of "true". 

"true" evaluates to "false" :/


jmazzitelli:
At least its another win for molecule tests - found it while running them with Service Mesh installed in the cluster.

PR #39, opened 2020-05-15
do a URL check to make sure github.com is accessible

jmazzitelli:
Fixes: https://github.com/kiali/kiali/issues/2795

This should return a better error message than before.

If github is not accessible, the task with the name "Confirm the cluster can access github.com when it needs to determine the last release of Kiali" will fail and be associated with an error such as this (as logged by the operator):

```
    "msg": "Failed to connect to api.github.com at port 443: [Errno 111] Connection refused", 
```


jmazzitelli:
Not sure how to test this - the cluster needs to not have access to the Internet (or more specifically, not have access to github.com) for the error to show up.

PR #41, opened 2020-05-18
re-enable arm64 buids 

jmazzitelli:
But do not require them to pass to allow PRs to be merged quicker (i.e. only need amd64 builds to pass for PRs to be mergeable). This is because arm64 builds sometimes take minutes to complete. Because ARM64 builds are not currently required or supported, we don't need to require these builds to pass.

Goes with https://github.com/kiali/kiali/issues/1091

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,7 @@ os: linux
 
 arch:
   - amd64
-  # It takes a very long time to spin up an ARM machine on travis.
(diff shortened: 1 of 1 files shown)


jmazzitelli:
This PR works. Travis shows "green" when the AMD64 build passes, even when the ARM64 build is still in-progress, as you see here:

image

PR #42, opened 2020-05-19
Add support for authentication through an OpenId provider

israel-hdez:
* Add "OpenId" as new strategy
  * Documentation for OpenId auth in the example CR file
  * Add default values for OpenId auth
  * Add 'openid' as a valid strategy for Kubernetes cluster but not for OpenShift
  * Add OpenId config validations in the playbook

Additionally, remove `localhost:3000` from the valid redirectURIs for the openshift authentication.

Part of kiali/kiali#2056

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -691,8 +691,8 @@ spec:
 #    ---
 #    port: 20001
 #
-# Defines the public domain where Kiali is being served. This the "domain" part
-# of the URL (usually it's a full-qualified domain name).
+# Defines the public domain where Kiali is being served. This is the "domain" part
+# of the URL (usually it's a fully-qualified domain name).
 # For example, "kiali.example.org".
 # When empty, Kiali will try to guess this value from HTTP headers.
 #    ---
@@ -710,4 +710,3 @@ spec:
 # When empty, Kiali will try to guess this value from HTTP headers.
 #    ---
 #    web_schema: ""
-
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -188,7 +188,7 @@ spec:
 # the issued token.
 #
 # When users are logging in, Kiali will wait at most 5 minutes after the "Login" button is pressed
-# in Kiali and redirected to the OpenID provider. If users don't login withing 5 minutes, Kiali will
+# in Kiali and redirected to the OpenID provider. If users don't login within 5 minutes, Kiali will
 # reject login even if the user authenticates successfully. You can change this time by adjusting
 # the 'authentication_timeout' option.
 #
@@ -202,7 +202,7 @@ spec:
 #      client_id: ""
 #      insecure_skip_verify_tls: false
 #      issuer_uri: ""
-#      scopes: ["profile", "email"]
+#      scopes: ["openid", "profile", "email"]
 #      username_claim: "sub"
 
 ##########
@@ -691,10 +691,23 @@ spec:
 #    ---
 #    port: 20001
 #
+# Defines the public domain where Kiali is being served. This the "domain" part
+# of the URL (usually it's a full-qualified domain name).
+# For example, "kiali.example.org".
+# When empty, Kiali will try to guess this value from HTTP headers.
+#    ---
+#    web_fqdn: ""
+#
 # Defines the context root path for the Kiali console and API endpoints and readiness probes.
 # When providing a context root path that is not "/", do not add a trailing slash.
 # For example, use "/kiali" not "/kiali/".
 # When empty, will default to "/" on OpenShift and "/kiali" on Kubernetes.
 #    ---
 #    web_root: ""
+#
+# Defines the public HTTP schema used to serve Kiali. This can only take
+# two possible values: either "http" or "https".
+# When empty, Kiali will try to guess this value from HTTP headers.
+#    ---
+#    web_schema: ""
 
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -163,7 +163,6 @@ spec:
 #      ldap_user_filter: "(cn=%s)"
 #      ldap_user_id_key: "cn"
 #
-#    ---
 # When "openid" strategy is chosen for authentication, you will need to fill some
 # required configurations in the following openid section. The "openid" authentication strategy
 # assumes that your Kubernetes cluster is properly configured to accept OpenID connect
@@ -188,12 +187,13 @@ spec:
 # suit your needs, you can change the 'scopes' setting as long as your cluster will accept
 # the issued token.
 #
+#    ---
 #    openid:
-#      issuer_uri: ""
+#      authorization_endpoint: ""
 #      client_id: ""
+#      issuer_uri: ""
 #      scopes: ["profile", "email"]
 #      username_claim: "sub"
-#      authorization_endpoint: ""
 
 ##########
 #  ---
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -187,10 +187,20 @@ spec:
 # suit your needs, you can change the 'scopes' setting as long as your cluster will accept
 # the issued token.
 #
+# When users are logging in, Kiali will wait at most 5 minutes after the "Login" button is pressed
+# in Kiali and redirected to the OpenID provider. If users don't login withing 5 minutes, Kiali will
+# reject login even if the user authenticates successfully. You can change this time by adjusting
+# the 'authentication_timeout' option.
+#
+# If your OpenID provider is using a certificate signed by an unknown CA, adjust the 'insecure_skip_verify_tls'
+# to disable the certificate verification. Remember this should be for testing purposes only.
+#
 #    ---
 #    openid:
+#      authentication_timeout: 300
 #      authorization_endpoint: ""
 #      client_id: ""
+#      insecure_skip_verify_tls: false
 #      issuer_uri: ""
 #      scopes: ["profile", "email"]
 #      username_claim: "sub"
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -135,6 +135,9 @@ spec:
 #  the individual's  RBAC roles in OpenShift. Not valid for non-OpenShift environments.
 # Choose "ldap" to enable LDAP based authentication. There are additional configurations for
 #  LDAP auth strategy that are required. See below for the additional LDAP configuration section.
+# Choose "openid" to enable OpenID connect based authentication. Your cluster is required to
+#  be configured to accept the tokens issued by your IdP. There are additional required
+#  configurations for this strategy. See below for the additional OpenID configuration section.
 # When empty, its value will default to "openshift" on OpenShift and "login" on Kubernetes.
 #    ---
 #    strategy: ""
@@ -159,6 +162,38 @@ spec:
 #      ldap_use_ssl: false
 #      ldap_user_filter: "(cn=%s)"
 #      ldap_user_id_key: "cn"
+#
+#    ---
+# When "openid" strategy is chosen for authentication, you will need to fill some
+# required configurations in the following openid section. The "openid" authentication strategy
+# assumes that your Kubernetes cluster is properly configured to accept OpenID connect
+# tokens and delegates token validation to the cluster's API Server. Thus, "openid" strategy
+# won't work without the proper configuration on your Kubernetes cluster
+# (see https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server).
+#
+# The "openid" strategy should NOT be used when installing Kiali in OpenShift. In this case, you
+# still should use "openshift" strategy and, in turn, setup OpenShift to use your IdP. For
+# OpenShift 4.4, you can find related docs here: https://docs.openshift.com/container-platform/4.4/authentication/identity_providers/configuring-oidc-identity-provider.html
+#
+# Kiali requires that your IdP allows the Implicit Flow grant. The 'issuer_uri' and 'client_id' attributes
+# are required values and must match your OpenId-Connect Kubernetes configuration. The 'username_claim'
+# is optional, but must match the value configured in your cluster (default value is "sub").
+#
+# You can, optionally, specify an 'authorization_endpoint' which is the URI where users will
+# be redirected to login and authorize access to Kiali. If you don't specify it, Kiali will try
+# to discover it using the 'issuer_uri'.
+#
+# By default, Kiali will ask for "openid" (forced), "profile" and "email" scopes when asking
+# for authentication to the OpenID provider. These are standard OpenID scopes. If this doesn't
+# suit your needs, you can change the 'scopes' setting as long as your cluster will accept
+# the issued token.
+#
+#    openid:
+#      issuer_uri: ""
+#      client_id: ""
+#      scopes: ["profile", "email"]
+#      username_claim: "sub"
+#      authorization_endpoint: ""
 
 ##########
 #  ---
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -43,6 +43,12 @@ kiali_defaults:
       ldap_use_ssl: false
       ldap_user_filter: "(cn=%s)"
       ldap_user_id_key: "cn"
+    openid:
+      issuer_uri: ""
+      client_id: ""
+      scopes: ["profile", "email"]
+      username_claim: "sub"
+      authorization_endpoint: ""
     strategy: ""
 
   deployment:
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -49,7 +49,7 @@ kiali_defaults:
       client_id: ""
       insecure_skip_verify_tls: false
       issuer_uri: ""
-      scopes: ["profile", "email"]
+      scopes: ["openid", "profile", "email"]
       username_claim: "sub"
     strategy: ""
 
@@ -181,7 +181,9 @@ kiali_defaults:
     metrics_enabled: true
     metrics_port: 9090
     port: 20001
+    web_fqdn: ""
     web_root: ""
+    web_schema: ""
 
 # These variables are outside of the kiali_defaults. Their values will be
 # auto-detected by the role and are not meant to be set by the user.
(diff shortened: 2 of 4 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -159,6 +162,38 @@ spec:
...
+# for authentication to the OpenID provider. These are standard OpenID scopes. If this doesn't
+# suit your needs, you can change the 'scopes' setting as long as your cluster will accept
+# the issued token.
+#
+#    openid:
+#      issuer_uri: ""
+#      client_id: ""
+#      scopes: ["profile", "email"]
+#      username_claim: "sub"
+#      authorization_endpoint: ""
```
jmazzitelli, line 196: sort alphabetically please ;)

```suggestion
#      authorization_endpoint: ""
#      client_id: ""
#      issuer_uri: ""
#      scopes: ["profile", "email"]
#      username_claim: "sub"
```

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -43,6 +43,12 @@ kiali_defaults:
       ldap_use_ssl: false
       ldap_user_filter: "(cn=%s)"
       ldap_user_id_key: "cn"
+    openid:
+      issuer_uri: ""
+      client_id: ""
+      scopes: ["profile", "email"]
+      username_claim: "sub"
+      authorization_endpoint: ""
```
jmazzitelli, line 51: sort alphabetically please ;)

```suggestion
      authorization_endpoint: ""
      client_id: ""
      issuer_uri: ""
      scopes: ["profile", "email"]
      username_claim: "sub"
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -159,6 +162,38 @@ spec:
 #      ldap_use_ssl: false
 #      ldap_user_filter: "(cn=%s)"
 #      ldap_user_id_key: "cn"
+#
+#    ---
```
jmazzitelli, line 166: Move this marker so it appears immediately above the new section's settings. The idea is that the `---` marker appears just above the actual section settings so the user's eyes are drawn to the values and knows what has to be uncommented to set the values. See the comment in lines 16 thru 19 that explains this.

```suggestion
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -159,6 +162,38 @@ spec:
...
+# You can, optionally, specify an 'authorization_endpoint' which is the URI where users will
+# be redirected to login and authorize access to Kiali. If you don't specify it, Kiali will try
+# to discover it using the 'issuer_uri'.
+#
+# By default, Kiali will ask for "openid" (forced), "profile" and "email" scopes when asking
+# for authentication to the OpenID provider. These are standard OpenID scopes. If this doesn't
+# suit your needs, you can change the 'scopes' setting as long as your cluster will accept
+# the issued token.
+#
+#    openid:
```
jmazzitelli, line 191: Move the `---` marker to here. See the comment in lines 16 thru 19 that explains this.

```suggestion
#    ---
#    openid:
```

roles/default/kiali-deploy/defaults/main.yml
jmazzitelli, line 51: sort:
```suggestion
      authorization_endpoint: ""
      client_id: ""
      issuer_uri: ""
      scopes: ["profile", "email"]
      username_claim: "sub"
```

roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -333,19 +333,26 @@
...
+  - kiali_vars.auth.strategy != 'openid'
 - name: Confirm ldap configuration when auth strategy is 'ldap'
   fail:
     msg: "Invalid configuration for LDAP! The mandatory parameters should be provided like `ldap_host', 'ldap_port', 'ldap_bind_dn', and 'ldap_base'"
   when:
   - kiali_vars.auth.strategy == "ldap"
   - kiali_vars.auth.ldap.ldap_host == "" or kiali_vars.auth.ldap.ldap_port <= 0 or kiali_vars.auth.ldap.ldap_bind_dn == "" or kiali_vars.auth.ldap.ldap_base == ""
+- name: Confirm OpenID configuration when auth strategy is 'openid'
+  fail:
+    msg: "Invalid configuration for OpenID connect! The mandatory parameters should be provided: `issuer_uri', 'client_id'. Also, the 'username_claim' cannot be set to the empty string."
```
jmazzitelli, line 352: This is kinda weird - not sure why that backtick character is in there. Change that backtick to a single-quote. I see you just copy-n-pasted from the ldap msg, so can you change that one, too?

```suggestion
    msg: "Invalid configuration for OpenID connect! The mandatory parameters should be provided: 'issuer_uri', 'client_id'. Also, the 'username_claim' cannot be set to the empty string."
```

roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -333,19 +333,26 @@
...
   when:
   - is_k8s == True
   - kiali_vars.auth.strategy != 'login'
   - kiali_vars.auth.strategy != 'anonymous'
   - kiali_vars.auth.strategy != 'ldap'
   - kiali_vars.auth.strategy != 'token'
+  - kiali_vars.auth.strategy != 'openid'
 - name: Confirm ldap configuration when auth strategy is 'ldap'
   fail:
     msg: "Invalid configuration for LDAP! The mandatory parameters should be provided like `ldap_host', 'ldap_port', 'ldap_bind_dn', and 'ldap_base'"
```
jmazzitelli, line 346: I know you didn't change this.. but.. see below.
```suggestion
    msg: "Invalid configuration for LDAP! The mandatory parameters should be provided like 'ldap_host', 'ldap_port', 'ldap_bind_dn', and 'ldap_base'"
```

roles/default/kiali-deploy/templates/openshift/oauth.yaml
```diff
@@ -8,6 +8,5 @@ metadata:
     version: {{ kiali_vars.deployment.version_label }}
 redirectURIs:
   - {{ kiali_route_url }}
-  - localhost:3000
```
jmazzitelli, line 11: This isn't part of the OpenID PR right? Are we sure this isn't needed and this won't break anything?

deploy/kiali/kiali_cr.yaml
israel-hdez: Sorted :)

roles/default/kiali-deploy/defaults/main.yml
israel-hdez: Also it's now sorted.

deploy/kiali/kiali_cr.yaml
israel-hdez: OK, it's moved down.

roles/default/kiali-deploy/tasks/main.yml
israel-hdez: My eyes didn't see that! Ok, I replaced the backtick.

roles/default/kiali-deploy/tasks/main.yml
israel-hdez: And fixed this backtick too

roles/default/kiali-deploy/templates/openshift/oauth.yaml
israel-hdez: It's not related to the OpenID work. This seems to be residuary code from previous dev-cycles. This seems to be part of a dev-setup and I think it shouldn't be here for better security.

israel-hdez:
Promoting this PR, because IMO it's good enough

israel-hdez:
@jshaughn Just FYI. Review if you have the time. If you don't, it's OK.

deploy/kiali/kiali_cr.yaml
```diff
@@ -159,6 +162,48 @@ spec:
...
+# be redirected to login and authorize access to Kiali. If you don't specify it, Kiali will try
+# to discover it using the 'issuer_uri'.
+#
+# By default, Kiali will ask for "openid" (forced), "profile" and "email" scopes when asking
+# for authentication to the OpenID provider. These are standard OpenID scopes. If this doesn't
+# suit your needs, you can change the 'scopes' setting as long as your cluster will accept
+# the issued token.
+#
+# When users are logging in, Kiali will wait at most 5 minutes after the "Login" button is pressed
+# in Kiali and redirected to the OpenID provider. If users don't login withing 5 minutes, Kiali will
```
jshaughn, line 191: typo: withing -> within

deploy/kiali/kiali_cr.yaml
```diff
@@ -159,6 +162,48 @@ spec:
...
+# reject login even if the user authenticates successfully. You can change this time by adjusting
+# the 'authentication_timeout' option.
```
jshaughn, line 193: Does it give any indication of a timeout or how to refresh the timer, or does it look like a credential failure?

deploy/kiali/kiali_cr.yaml
```diff
@@ -135,6 +135,9 @@ spec:
 #  the individual's  RBAC roles in OpenShift. Not valid for non-OpenShift environments.
 # Choose "ldap" to enable LDAP based authentication. There are additional configurations for
 #  LDAP auth strategy that are required. See below for the additional LDAP configuration section.
+# Choose "openid" to enable OpenID connect based authentication. Your cluster is required to
+#  be configured to accept the tokens issued by your IdP. There are additional required
+#  configurations for this strategy. See below for the additional OpenID configuration section.
 # When empty, its value will default to "openshift" on OpenShift and "login" on Kubernetes.
```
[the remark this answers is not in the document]
jmazzitelli: remember, if we default to "token" - people will have to go through some hoops to find out how to login (do YOU know how to get a valid login token?).. so for demo purposes or people just getting started with kiali, they may not know how to login when presented with the login screen.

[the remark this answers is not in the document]
jshaughn: If we are going to deprecate `login` then we can't use `login` as a default for more than maybe one more release.  Perhaps we should just default to `anonymous`, I mean it's about as useful as `login` and for demo's why even have a login?  Any prod install will want to use a real auth strategy.  On OS we can still default to `openshift`.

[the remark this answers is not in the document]
jmazzitelli: +1 on defaulting to anonymous. Would get new users up and running faster and they won't have to worrying about obtaining any credentials.

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -43,6 +43,14 @@ kiali_defaults:
       ldap_use_ssl: false
       ldap_user_filter: "(cn=%s)"
       ldap_user_id_key: "cn"
+    openid:
+      authentication_timeout: 300
+      authorization_endpoint: ""
+      client_id: ""
+      insecure_skip_verify_tls: false
+      issuer_uri: ""
+      scopes: ["profile", "email"]
```
jmazzitelli, line 52: I know that in the server code you forcibly add "openid" if it doesn't exist, but for clarity, shouldn't you just put "openid" in here explicitly? That way people seeing this will know, "ah, ok, this does have openid scope" - otherwise, they will be asking the question, "Why isn't openid in here? Do I need to add it?" I realize they can read it in the above comments, but I think it is best to also explicitly set it here to. Again, that just makes it clear.

deploy/kiali/kiali_cr.yaml
```diff
@@ -159,6 +162,48 @@ spec:
...
+# to disable the certificate verification. Remember this should be for testing purposes only.
+#
+#    ---
+#    openid:
+#      authentication_timeout: 300
+#      authorization_endpoint: ""
+#      client_id: ""
+#      insecure_skip_verify_tls: false
+#      issuer_uri: ""
+#      scopes: ["profile", "email"]
```
jmazzitelli, line 205: See my comment below about adding "openid" explicitly here.

deploy/kiali/kiali_cr.yaml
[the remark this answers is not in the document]
jmazzitelli: BTW: I would do this "deprecating the login strategy" in a separate PR so as not to make this more complicated.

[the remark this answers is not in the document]
jmazzitelli: I am going to backtrack and say the default should NOT be anonymous. We need to have true auth here... so default to "token". Upstream Istio people are worried that even the demo profile comes with out of box admin/admin credentials that they don't like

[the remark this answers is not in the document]
israel-hdez: I agree and prefer deprecation to be done in another PR.
And, IMO, "token" should be the default for kubernetes. Even if we wouldn't deprecate anything, like Istio folks I don't like too much the fact of leaving the doors open with `anonymous` nor partially open with `login` with default credentials.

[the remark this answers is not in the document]
jshaughn: @israel-hdez @jmazzitelli OK, we're in agreement to deprecate Login and LDAP in 1.19.0 (assuming this makes it into 1.19.0).  And the new default will be Token.  This will also handle the recent complaint about the admin/admin login account.  Mazz, perhaps you can let them know that we're responding in our upcoming release.

cc @lucasponce FYI about auth direction in case you have any comment/concern.

deploy/kiali/kiali_cr.yaml
israel-hdez: Fixed. openid is added to the list.

roles/default/kiali-deploy/defaults/main.yml
israel-hdez: Fixed. openid is added to the list.

deploy/kiali/kiali_cr.yaml
israel-hdez: It will look like a credential failure.

deploy/kiali/kiali_cr.yaml
israel-hdez: Fixed.

israel-hdez:
In the last commit, I fixed feedback and added configs `web_fqdn` and `web_schema` because of Joel's feedback on the back-end PR (about that GuessKialiURL function).

deploy/kiali/kiali_cr.yaml
```diff
@@ -691,10 +691,23 @@ spec:
 #    ---
 #    port: 20001
 #
+# Defines the public domain where Kiali is being served. This the "domain" part
+# of the URL (usually it's a full-qualified domain name).
```
jmazzitelli, line 695: Missing the word "is" and missing "y" in the term "fully-qualified":

```suggestion
# Defines the public domain where Kiali is being served. This is the "domain" part
# of the URL (usually it's a fully-qualified domain name).
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -691,10 +691,23 @@ spec:
...
 # For example, use "/kiali" not "/kiali/".
 # When empty, will default to "/" on OpenShift and "/kiali" on Kubernetes.
 #    ---
 #    web_root: ""
+#
+# Defines the public HTTP schema used to serve Kiali. This can only take
+# two possible values: either "http" or "https".
+# When empty, Kiali will try to guess this value from HTTP headers.
+#    ---
+#    web_schema: ""
```
jmazzitelli, line 712: We do not need this setting at all. You already know this (and we do this kind of check already), by looking at the `identity` settings. In other words, there is no need to ask an admin user to set a setting like this explicitly when by virtue of the admin user setting `identity` you already know if kiali will be using http or https. See: https://github.com/kiali/kiali-operator/blob/v1.18.0/deploy/kiali/kiali_cr.yaml#L518-L525

If identity.cert_file and identity.private_key_file are explicitly left empty, then you know the schema is 'http' by definition. If these are set, then you know the schema is 'https' (again, by definition). Giving Kiali an identity means you have an identity cert that will identify the server via https. We use this very knowledge when setting up the http/https endpoint here: https://github.com/kiali/kiali/blob/05d65e5757ba908332b7b7eecd902644a255822b/server/server.go#L66-L73

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -181,7 +181,9 @@ kiali_defaults:
     metrics_enabled: true
     metrics_port: 9090
     port: 20001
+    web_fqdn: ""
     web_root: ""
+    web_schema: ""
```
jmazzitelli, line 186:
```suggestion
```

Delete this - we do not need it. Use `identity`.

roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -1,6 +1,14 @@
...
     msg: INSTALLING KIALI
 
+- name: Validate web_schema configuration
+  fail:
+    msg: "Invalid server.web_schema [{{ kiali_vars.server.web_schema }}]! Must be empty, or one of either 'http' or 'https'"
+  when:
+  - kiali_vars.server.web_schema != ''
+  - kiali_vars.server.web_schema != 'https'
+  - kiali_vars.server.web_schema != 'http'
+
```
jmazzitelli, line 11:
```suggestion
```
Delete all changes here in this file - not needed.

deploy/kiali/kiali_cr.yaml
israel-hdez: That's partially true.

When you set the identity option, Kiali will serve via https **within** the cluster. But the ingres, loadbalancer, or whatever exposure method you choose, may use another *public-facing* protocol.

You can try in your minikube setup. Kiali should be listening in plain HTTP because there is no identity by default. But if you use the ingress endpoint, you can access Kiali either through HTTP or HTTPS. Try going to either **https**://{your_minikube_ip}/kiali or **http**://{your_minikube_ip}/kiali and both endpoints should respond adequately as plain/secure http respectively. The other case could also be true: you could fill the `identity` configuration and Kiali pod will be serving with HTTPS, but the ingress/loadbalancer/whatever may be exposing Kiali publicly through plain HTTP.

So, I think the `identity` option is unreliable regarding the *public-facing* protocol and that's the reason I added this option. But well... We could try the "identity" route if you think it's OK enough.

jmazzitelli: hmmm. ok. lemme think about that.

roles/default/kiali-deploy/tasks/main.yml
jmazzitelli, line 11: OK, if we are keeping this, can you move this down? I would like this away from the top.. move this down to where we also do some checking of the web_root. So these lines should go on about line 286 (put it after the web_root processing). So put this about here: https://github.com/kiali/kiali-operator/blob/v1.18.1/roles/default/kiali-deploy/tasks/main.yml#L286

israel-hdez: It's moved down.

israel-hdez:
Passing to QA.

PR #45, opened 2020-05-26
make openshift console links in masthead more specific

jmazzitelli:
fixes kiali/kiali#2828


jmazzitelli:
My one question here is - should we use the deployment.namespace or the istio_namespace?

Technically, it should be using the istio_namespace (which is the control plane namespace). In Service Mesh, these are one and the same, so the end result is the same.

jmazzitelli:
What this will show now is links in the masthead that make it more obvious which Kiali instance the links will take you to - the namespace where kiali is installed - typically the control plane namespace - is shown in brackets like this:

image

PR #47, opened 2020-06-04
do not create ApplicationMenu consolelinks

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2861

cc @maltron - I assume we want this, even though I know the design is not fully fleshed out (i.e. we know we do not want these application menu links in the OS masthead).


jshaughn:
Adding DNM until this is ties to 4.6 console Jira.

PR #48, opened 2020-06-05
Deprecate "login" strategy

israel-hdez:
This is moving from "login" to "token" as the default strategy for
non-OpenShift clusters.

A message about what can be used to log in to Kiali is also added after
Kiali is deployed.

Also "openid" is added to the list of valid strategies.

Part of kiali/kiali#2862

### DIFF
--- a/deploy/deploy-kiali-operator.sh
+++ b/deploy/deploy-kiali-operator.sh
@@ -149,7 +149,9 @@
 #     based on the individual's RBAC roles in OpenShift.
 #    Choose "ldap" to enable LDAP based authentication. There are additional configurations for
 #     LDAP auth strategy that are required. See below for the additional LDAP configuration section.
-#    Default: "openshift" (when using OpenShift), "login" (when using Kubernetes)
+#    Choose "openid" to enable openid based authentication. There are additional configurations for
+#     openid strategy that are required. See below for the additional configuration section.
+#    Default: "openshift" (when using OpenShift), "token" (when using Kubernetes)
 #
 # CREDENTIALS_CREATE_SECRET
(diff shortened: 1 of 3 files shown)


deploy/deploy-kiali-operator.sh
```diff
@@ -1246,3 +1247,19 @@ EOF
...
+  echo
+  echo "You could use the secrets of the Kiali ServiceAccount. To find the names of the secrets run:"
+  echo "  ${CLIENT_BASENAME} get serviceaccount -n ${NAMESPACE} kiali-service-account -o jsonpath='{.secrets[*].name}'"
+  echo
+  echo "If there are several secrets associated with the Kiali ServiceAccount, you will need to find which"
+  echo "one is accepted at login."
+  echo
+  echo "Get the bearer token of the secrets using the command:"
+  echo "  ${CLIENT_BASENAME} get secret -n ${NAMESPACE} {secret_name} -o jsonpath='{.data.token}'"
+fi
```
jmazzitelli, line 1265: I wouldn't put any of this in here. This should be in the docs. Because if we spit this out for "token", why not also spit out messages for "openid" (e.g. "You are using openid strategy. You will need to provide a valid OpenID Connect implemention in your cluster or things will not work. Yadda yadda yadda, etc, etc, etc").

So rather than have this long and loud output of this script (which only shows up once - if the user misses it, they miss it completely and its lost to them), we should put this in our docs.

israel-hdez: OK, I removed all this text. I'll do that change of the kiali/kiali repo.

PR #54, opened 2020-06-16
Remove component namespaces config

jmazzitelli:
This is the operator-side fixes: https://github.com/kiali/kiali/issues/2891


jmazzitelli:
at this time, we are not going to do this.

PR #56, opened 2020-06-18
Use new OpenShift interface for service serving certificates

israel-hdez:
OpenShift is now requiring new annotations for the service serving certificates interface. This is doing the needed changes to use the new interface.

Part of kiali/kiali#2917.


israel-hdez:
@jmazzitelli Although this moves away from the deprecated OS interface, I'm not very sure this is the way to go. Merging this would mean that changes will be required on maistra operator to pass the right `ca_file` config.

(I haven't tried this... Test is pending)

roles/default/kiali-remove/tasks/main.yml
```diff
@@ -164,6 +164,7 @@
   - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='OAuthClient', resource_name='kiali-' + kiali_vars.deployment.namespace, api_version='oauth.openshift.io/v1') if is_openshift == True else [] }}"
   - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='Route', resource_name='kiali', api_version='route.openshift.io/v1') if is_openshift == True else [] }}"
   - "{{ query('k8s', kind='ConsoleLink', label_selector='kiali.io/home=' + kiali_vars.deployment.namespace)}}"
+  - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='ConfigMap', resource_name='kiali-cabundle', api_version='v1') }}"
```
jmazzitelli, line 167: can you add ` if is_openshift == True else []` to this? While you are here, can you also add it to line 166 - it is missing there (I wondered why I was seeing warning message while on minikube in here!! :))

israel-hdez: OK, both are fixed.

PR #57, opened 2020-06-19
do not regenerate a signing key secret if we already have one

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2921


jmazzitelli:
my testing shows this working. and it fixes the problem where it triggers the reconciliation in an infinite loop if I turn on watchDependentResources in the operator.

jmazzitelli:
To test this - install operator, install a basic Kiali CR (do not define signing_key, which you do not typical set anyway).

Look at the kiali-signing-key resource created in istio-system namespace. Look at the key value.

Now touch the Kiali CR (just change, say, deployment.verbose_mode - just change anything basic).

Now look at the secret again - make sure the signing key value does not change.

PR #62, opened 2020-06-28
be able to run the Molecule tests on platforms that only have podman - pass DORP to the molecule config

jmazzitelli:
this should be paired with https://github.com/kiali/kiali/pull/2945 - it is necessary to run tests on systems that only have podman (e.g. default Fedora32 systems).


jmazzitelli:
the ldap molecule test has been broken since, I think, the new openid code went in. The ldap feature is going away soon (already deprecated) so rather than me try to debug and fix this test, this PR removes it. I believe it is just the test that is broke, not the ldap feature itself.

PR #64, opened 2020-07-01
add test to see cabundle is created/destroyed appropriately

jmazzitelli:
cc @israel-hdez


jmazzitelli:
Because this test is specific to OpenShift (cabundle CM is not created on minikube, for example), I need to add a conditional here to check only when appropriate

jmazzitelli:
To run this test, install OpenShift cluster with Istio (do not install Kiali). Then push your dev images via `make cluster-push`. Finally run the test:

`MOLECULE_USE_DEV_IMAGES=true MOLECULE_SCENARIO="default-namespace-test" make molecule-test`

jmazzitelli:
@israel-hdez this passes on my box

PR #74, opened 2020-07-13
Molecule tests should dump the logs on failure

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2984

do not squash merge this - the one commit moves to molecule 3 (a prerequisite). The next commit implements the fix.


jmazzitelli:
Ignore the error - there is a regression that was introduced in operator-courier 2.1.8 that was just released which is causing the failure here.

PR #77, opened 2020-07-14
Add helm chart

naphta:
As mentioned in https://github.com/kiali/kiali/issues/195 this is the chart which I've been working on.

I've done a little bit more work so that the chart can be created and packaged from the Makefile. I'm not sure if that's necessarily where you want this logic but I was expecting that this would be best to publish on Helm hub or just directly as a release artifact.

Ultimately it's not a lot different from the chart @jmazzitelli wrote originally, I did add in support for creating ingress resources from an Istio point of view (gateway/virtualservice) for now but I could see that being something the operator generally supports going forward.


Makefile
```diff
@@ -30,6 +30,32 @@ help: Makefile
...
+	    os=$$(uname -s | tr '[:upper:]' '[:lower:]') ;\
+	    arch="" ;\
+	    case $$(uname -m) in \
+	        i386)   arch="386" ;; \
+	        i686)   arch="386" ;; \
+	        x86_64) arch="amd64" ;; \
+	        arm)    dpkg --print-architecture | grep -q "arm64" && arch="arm64" || arch="arm" ;; \
+	    esac ;\
+	    mkdir -p "${OUTDIR}/helm-install" ;\
+		cd "${OUTDIR}/helm-install" ;\
```
jmazzitelli, line 49: you should not have to do this since it was already done on line 35 above. Also adjust indentation on the `cd` line.
```suggestion
	    cd "${OUTDIR}/helm-install" ;\
```

deploy/charts/kiali-operator/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v2
+name: kiali-operator
+description: Kiali is an open source project for service mesh observability, refer to https://www.kiali.io for details.
+version: 1.0.0
+appVersion: 1.0.0
```
jmazzitelli, line 5: don't we want the appVersion of Kiali here? 1.0.0 implies the (very old) Kiali. I think we'd want some kind of ${xyz} environment variable replacement here so we can build the charts for a specific version of Kiali.

deploy/charts/kiali-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,311 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
jmazzitelli, line 1: Here's what I suggest for this file. Because its essentially identical to our other clusterrole.yaml, you should indent it the same way; e.g. put the list items on the same level as the parent. 

That way we can do a diff and determine what, if anything, changed between one versus the other (this lets us more easily keep them in sync).

For example,

```
resources:
- configmaps
```

as opposed to:

```
resources:
  - configmaps
```

This way one can do a "diff" on this file and the other clusterrole.yaml and at least get a better view of what changed (otherwise, every line will look changed with the different indentation).

deploy/charts/kiali-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,82 @@
+apiVersion: apps/v1
```
jmazzitelli, line 1: Same comment here regarding indentation. Keep it looking at identical as possible to the other deployment.yaml.

That way we can do a diff and determine what, if anything, changed between one versus the other (this lets us more easily keep them in sync).

deploy/charts/kiali-operator/values.yaml
```diff
@@ -0,0 +1,59 @@
+nameOverride: ""
+fullnameOverride: ""
+
+image:
+  repo: ${$OPERATOR_QUAY_NAME}
+  tag: ${$OPERATOR_QUAY_TAG}
+  pullPolicy: Always
+  pullSecrets: []
+  # - name: xyz
```
jmazzitelli, line 9: Should delete this superfluous line.
```suggestion
```

deploy/charts/kiali-operator/templates/istio-destinationrule.yaml
```diff
@@ -0,0 +1,16 @@
+{{ if eq .Values.istio.ingress.enabled true }}
```
jmazzitelli, line 1: Regarding the destination rule/gateway/virtual service, we can leave this here since it is the "istio way" of exposing the Kiali server, however, the operator does provide a k8s-native way (using an Ingress) and an OpenShift-native way (using Route) - that's what this setting does:

https://github.com/kiali/kiali-operator/blob/v1.18.0/deploy/kiali/kiali_cr.yaml#L230-L233

Perhaps in the future the operator could be told to create an ingress using the "istio" way, but that doesn't need to hold up this PR.

jmazzitelli:
Quick run of `make build-helm-chart` results in:

```
....
Building helm chart for Kiali operator
/bin/bash: helm: command not found
make: *** [build-helm-chart] Error 127
```

Makefile
jmazzitelli: I include this in my diff down below when I fixed the issue with the helm executable not being found

deploy/charts/kiali-operator/templates/istio-destinationrule.yaml
naphta: Completely agree, it's quite new for Istio to recommend it to be done this way. In the older versions Kubernetes ingress resources were also just created. 

I can open up an issue to add in that feature perhaps?

deploy/charts/kiali-operator/values.yaml
naphta: Ah, I tend to do this (as do many other chart creators) to show an example of how to use that value. Happy to remove it but it seems to be a commonly trodden path. [An example here.](https://github.com/helm/charts/blob/master/stable/nginx-ingress/values.yaml#L127)

deploy/charts/kiali-operator/Chart.yaml
naphta: During the chart artifact creation I set the appVersion there, I was thinking the less which needs to be updated the better. This way the artifact is always inline with the released version but nobody needs to come in here and make a change.

jmazzitelli: Ah.. OK. if its changed automagically on chart creation, you can ignore me.

deploy/charts/kiali-operator/templates/istio-destinationrule.yaml
jmazzitelli: > I can open up an issue to add in that feature perhaps?

Feel free to add a future enhancement request issue.

jmazzitelli:
I am contemplating switching over our current "non-OperatorHub" install mechanism (bash) to this helm chart, versus what we have today which is this bash script : https://kiali.io/getLatestKialiOperator.

This will involve a lot of coordination and work (it will involve updating docs, changing the molecule tests, updating the Makefile, and notifying users). I'll need to discuss this with the Kiali team first before committing to this. But I'm starting to think about it. I'd rather have a helm chart and OperatorHub as the two main installation mechanisms rather than have a bash script and OperatorHub plus a third (helm) sitting off to the side.

deploy/charts/kiali-operator/values.yaml
```diff
@@ -0,0 +1,59 @@
+nameOverride: ""
+fullnameOverride: ""
+
+image:
+  repo: ${$OPERATOR_QUAY_NAME}
+  tag: ${$OPERATOR_QUAY_TAG}
```
jmazzitelli, line 6: I don't see how this could ever work - this looks wrong with `${$`. I think this should be `${OPERATOR...`

jmazzitelli:
FYI: I'm taking this PR and building off it to get it to work and with additional stuff. I will most likely close this PR and create a new one.

deploy/charts/kiali-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+          {{- end }}
+          volumeMounts:
+            - mountPath: /tmp/ansible-operator/runner
+              name: runner
+              readOnly: true
+          {{- if .Values.resources.ansible }}
+          resources: {{- toYaml .Values.resources.ansible | nindent 12 }}
+          {{- end }}
+        - name: operator
+          image: "{{ .Values.image.hub }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
```
jmazzitelli, line 52: uses ".hub" but the values.yaml has ".repo" So this doesn't resolve to anything.
Also I don't see .name defined in values.yaml

jmazzitelli:
Closing this in favor of https://github.com/kiali/kiali-operator/pull/93
I took the commit in this PR and built on it to get a working helm chart. Will continue to work on it.

PR #78, opened 2020-07-14
be able to tell what dashboards to create

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2989
Users probably won't want all of them, and they just cause the Kiali CR processing to be slower than need be.
The default is backward compat to previous behavior (all monitoring dashboard resources are created) but you now have the ability to not install those you do not want. e.g., in the Kiali CR:

```
deployment:
  custom_dashboards:
    includes: ['go*', 'kiali*']
```

This will install the Go and Kiali monitoring dashboards but no others.

```
deployment:
  custom_dashboards:
    includes: ['vertx*']
    excludes: ['vertx-server*']
```

This will install all the vertx dashboards except the server.

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -237,6 +237,19 @@ spec:
 #      pod: {}
 #      pod_anti: {}
 #
+# Names of the out-of-box custom monitoring dashboards that are to be installed.
+# The custom monitoring dashboards are defined in yaml files located within the operator.
+# Consult the operator templates for the custom monitoring dashboard yaml files available.
+# For example, see this for the current list of yaml files available:
+# https://github.com/kiali/kiali-operator/tree/master/roles/default/kiali-deploy/templates/dashboards
+# These settings will determine the additional metric graphs that you will see within the Kiali UI.
+# You can specify an includes and excludes list, the excludes list takes precedence.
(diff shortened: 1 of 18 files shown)


jmazzitelli:
You would think the default excludes can be an empty string but you will notice the default is `"_"` - this is because if it is empty, [a bug in ansible find task](https://github.com/ansible/ansible/issues/70640) will exclude everything.

jmazzitelli:
This PR includes a molecule test that shows this working.

jmazzitelli:
@jotak what are we calling the "monitoring dashboards" today? Is it still "monitoring dashboards"?

This PR introduces a new section in the Kiali CR under `deployments` called `dashboards` - I want to make sure I name that new section correctly. I was thinking `monitoring_dashboards` but I do not know if that is the official name for these things.

For details of this new section, see [the kiali_cr.yaml comments](https://github.com/kiali/kiali-operator/pull/78/files#diff-6ed3eb450a43cd770b902d09d55e3824R240-R251).

jotak:
@jmazzitelli they are still 'monitoringdashboards' resources, although if I had to name them today perhaps I would probably choose 'customdashboards'.
I'm not sure if this setting should be named just 'dashboard', to me it's a very broad term, for many users Kiali itself is a dashboard. I fear it could cause confusion.

jmazzitelli:
I'll rename to `custom_dashboards`

PR #82, opened 2020-07-17
Configurable thresholds for error rates

aljesusg:
Related with https://github.com/kiali/kiali/pull/2807

UI: https://github.com/kiali/kiali-ui/pull/1831
Backend: https://github.com/kiali/kiali/pull/2915

Fix https://github.com/kiali/kiali/issues/1424

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -609,7 +609,6 @@ spec:
 #      namespace_selector: true
 #      url: ""
 #      whitelist_istio_system: ["jaeger-query", "istio-ingressgateway"]
-#
 ##########
 #  ---
 #  health_config:
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -779,3 +779,43 @@ spec:
 # When empty, Kiali will try to guess this value from HTTP headers.
 #    ---
 #    web_schema: ""
+
+##########
+#  ---
+#  health_config:
+#
+#
+# rate: A list of configurations health that Kiali apply. Each item contains:
+#   namespace: The name of the namespace to apply. Kiali match the value like a regular expression
+#   kind: The type of the resource to apply. Kiali match the value like a regular expression
+#   name: Tha name of the resource to apply. Kiali match the value like a regular expression
+#   tolerance: A list of tolerances for this configuration. Each item contains:
+#     protocol: The protocol to apply the tolerance (grpc | http | ...) Kiali match the value like a regular expression
+#     direction: The direction to apply the tolerance (inbound | outbound | ...).Kiali match the value like a regular expression
+#     code: The status code to apply the tolerance. Kiali match the value like a regular expression
+#     degraded: The value in % to show a degraded status
+#     failure: The value in % to show a failure status,
+#    ---
+#    rate:
+#      - namespace: "default"
+#        tolerance:
+#          - request_protocol: "http"
+#            direction: "inbound"
+#            code: "400"
+#            degraded: 5
+#            failure: 10
+#      - namespace: default
+#        tolerance:
+#          - request_protocol: "grpc"
+#            code: "1"
+#            degraded: 10
+#            failure: 15
+#      - namespace: bookinfo
+#        kind: app
+#        name: reviews
+#        tolerance:
+#          - request_protocol: "http"
+#            code: "400"
+#            degraded: 0
+#            failure: 5
+#
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -526,12 +526,12 @@ spec:
 #      name: ".*"
 #      namespace: ".*"
 #      tolerance:
-#      - code: "/^[4-5]\\d\\d$/"
+#      - code: "^[4-5]\\d\\d$"
 #        degraded: 0.1
 #        direction: ".*"
 #        failure: 20
 #        protocol: "http"
-#      - code: "/^[1-9]$|^1[0-6]$/"
+#      - code: "^[1-9]$|^1[0-6]$"
 #        degraded: 0.1
 #        direction: ".*"
 #        failure: 20
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -505,24 +505,6 @@ spec:
 #      in_cluster_url: "http://grafana.istio-system:3000"
 #      url: ""
 #
-# **Health Configuration**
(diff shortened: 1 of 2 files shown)


roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -189,7 +189,22 @@ kiali_defaults:
...
+          - code: "/^[4-5]\\d\\d$/"
+            protocol: "http"
+            direction: ".*"
+            degraded: 0.1
+            failure: 20
+          - code: "/^[1-9]$|^1[0-6]$/"
+            protocol: "grpc"
+            direction: ".*"
+            degraded: 0.1
+            failure: 20
```
jmazzitelli, line 206: Those commas at the end of rate.namespace, rate.kind, and rate.name look invalid. Remove those commas (we don't put commas at the end of settings, even if it somehow is valid yaml).

Also, make the indentation consistent as all of our other yaml files. We do not indent two spaces for each list item. Put the "-" on the same level as the parent.

Also, alphabetic sort all items in the same group.

So this is my recommended changes:

```suggestion
    rate:
    - kind: ".*"
      name: ".*"
      namespace: ".*"
      tolerance:
      - code: "/^[4-5]\\d\\d$/"
        degraded: 0.1
        direction: ".*"
        failure: 20
        protocol: "http"
      - code: "/^[1-9]$|^1[0-6]$/"
        degraded: 0.1
        direction: ".*"
        failure: 20
        protocol: "grpc"
```

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -189,7 +189,22 @@ kiali_defaults:
...
+
```
jmazzitelli, line 207: Delete this extra blank line.
```suggestion
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -723,3 +723,43 @@ spec:
 # When empty, Kiali will try to guess this value from HTTP headers.
 #    ---
 #    web_schema: ""
+
+##########
+#  ---
+#  health_config:
+#
+#
```
jmazzitelli, line 731: Delete this extra blank line.
```suggestion
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -723,3 +723,43 @@ spec:
...
+#            degraded: 10
+#            failure: 15
+#      - namespace: bookinfo
+#        kind: app
+#        name: reviews
+#        tolerance:
+#          - request_protocol: "http"
+#            code: "400"
+#            degraded: 0
+#            failure: 5
```
jmazzitelli, line 764: The values here must document the actual defaults (do not put in arbitrary examples). Please [see these comments at the top of this kiali_cr.yaml file that explain this](https://github.com/kiali/kiali-operator/blob/v1.21/deploy/kiali/kiali_cr.yaml#L16-L19) to the readers. So when someone reads this kiali_cr.yaml file, they will know what the defaults are by simply looking at the examples here in the comments. So in other words, just replace these example settings with the actual defaults you have in the defaults/main.yaml in this PR.

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -188,6 +188,24 @@ kiali_defaults:
     web_root: ""
     web_schema: ""
 
+  health_config:
```
jmazzitelli, line 191: Please put all settings in alphabetical sort order. So this "health_config" should go up after the `external_services` section and before the `identity` section ([see line 148 here](https://github.com/kiali/kiali-operator/blob/f8eb903fa270352bdba671d711419bee4e98bff6/roles/default/kiali-deploy/defaults/main.yml#L148))

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -188,6 +188,24 @@ kiali_defaults:
...
+    rate:
+      - namespace: ".*",
```
jotak, line 193: it doesn't work for me, shouldn't be comas at the end of those lines.
I get the following error:

```
ERROR! Syntax Error while loading YAML.
  expected <block end>, but found ','

The error appears to be in '/opt/ansible/roles/default/kiali-deploy/defaults/main.yml': line 193, column 24, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    rate:
      - namespace: ".*",
                       ^ here

```

jotak:
@aljesusg  @jmazzitelli correct me if I'm wrong ... should there be some new entry in variables file `roles/default/kiali-deploy/vars/main.yml` ?

The reason I'm asking is that, even when I fix the comma issue, I don't see this new health config being populated in the resulting config map ... so in the end whatever I put there gets ignored, and I'm not sure why.

jmazzitelli:
> @aljesusg @jmazzitelli correct me if I'm wrong ... should there be some new entry in variables file `roles/default/kiali-deploy/vars/main.yml` ?

Oh, good catch! Yes, as the [comments at the top of the defaults/main.yml](https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L8) indicate, if you create a whole new top-level group of settings, you need to add that to vars/main.yml.

aljesusg:
@jmazzitelli Should I add this to roles/default/kiali-deploy/vars/main.yml too?

deploy/kiali/kiali_cr.yaml
```diff
@@ -505,6 +505,24 @@ spec:
...
+#   name: Tha name of the resource to apply. Kiali match the value like a regular expression
+#   tolerance: A list of tolerances for this configuration. Each item contains:
+#     protocol: The protocol to apply the tolerance (grpc | http | ...) Kiali match the value like a regular expression
+#     direction: The direction to apply the tolerance (inbound | outbound | ...).Kiali match the value like a regular expression
+#     code: The status code to apply the tolerance. Kiali match the value like a regular expression
+#     degraded: The value in % to show a degraded status
+#     failure: The value in % to show a failure status,
+#    ---
+#    rate: []
+#
```
jmazzitelli, line 525: This looks to be in the wrong spot. Is this "health_config" part (a child of) `external_services`? Because that's where this is showing it to be.

This was a brand new top-level group. It belongs here, right before the top-level `identity` section: https://github.com/kiali/kiali-operator/blob/v1.23.0/deploy/kiali/kiali_cr.yaml#L612

jmazzitelli (changes requested):
I went through and updated the docs in the comment. See what you think. I think it makes things more clear and corrects some of the language a bit.

jmazzitelli (approved):

PR #86, opened 2020-07-18
be able to run molecule test "metrics-test" in Service Mesh environment.

jmazzitelli:
This ensures this test can run with either upstream Istio or Service Mesh.

fixes: https://github.com/kiali/kiali/issues/3003


jmazzitelli:
This is ready for merge. The test actually has a legitimate failure, but we'll fix that in another PR. This PR is simply to be able to get the molecule tests to run when inside a Service Mesh Environment, which they now do.

PR #87, opened 2020-07-19
Service Mesh doesn't scrape operator so ignore those test errors

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3006

Don't know why, but Service Mesh doesn't scrape operator pod for metrics. Upstream Istio does. It might be due to an older version of Prometheus that ships with Service Mesh 1.1 (that's just pure speculation, I don't know).

This ignores the test failures when testing with SM.

### DIFF
--- a/molecule/asserts/configmap_asserts.yml
+++ b/molecule/asserts/configmap_asserts.yml
@@ -23,10 +23,12 @@
     that:
     - kiali_configmap.external_services.prometheus.custom_metrics_url == "http://prometheus.{{ istio.control_plane_namespace }}:9090"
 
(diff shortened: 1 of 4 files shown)


jmazzitelli:
The changes made to get things to run with Service Mesh broke some other tests.. additional commits attempt to fix those.

jotak:
The prometheus scrape jobs have been changed upstream, indeed.
The diff that you're showing in the ticket are what I was talking about in this ticket: https://github.com/istio/istio/issues/24075

If I remember correctly, the previous config (which is still used in Service Mesh) requires some annotation like `prometheus.io/scheme: http` ... it needs to be checked

PR #90, opened 2020-07-21
fix operator due to SDK upgrade

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2611
fixes: https://github.com/kiali/kiali/issues/2073

Now that we are able to bump up the SDK to a newer version on both upstream and downstream, we need to update the operator.


jmazzitelli:
molecule testing with the 4.5 downstream ansible base image shows this all working.

jmazzitelli:
@jshaughn this is needed to move to the newest productized ansible operator SDK base image. We'll want this so we can keep up with the base image SDK and its CVE fixes that they release weekly.

jmazzitelli:
> I'll assume the ansible operator SDK is updated outside of this PR. Otherwise it's missing.

will be done in productization git repo (not here).

PR #92, opened 2020-07-22
Modify 3scale config. Add Patch verb to workloads

lucasponce:
Needed for https://github.com/kiali/kiali/pull/2936

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -191,6 +191,7 @@ rules:
   verbs:
   - get
   - list
+  ${OPERATOR_ROLE_PATCH}
   - watch
 - apiGroups: ["extensions", "apps"]
   resources:
@@ -200,6 +201,7 @@ rules:
   verbs:
   - get
   - list
+  ${OPERATOR_ROLE_PATCH}
   - watch
 - apiGroups: ["autoscaling"]
   resources:
@@ -215,6 +217,7 @@ rules:
   verbs:
   - get
   - list
+  ${OPERATOR_ROLE_PATCH}
   - watch
 - apiGroups:
   - config.istio.io
@@ -256,6 +259,7 @@ rules:
   verbs:
   - get
   - list
+  ${OPERATOR_ROLE_PATCH}
   - watch
 - apiGroups: ["project.openshift.io"]
   resources:
--- a/roles/default/kiali-deploy/templates/kubernetes/role.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/role.yaml
@@ -21,6 +21,7 @@ rules:
   - get
   - list
   - watch
+  - patch
 - apiGroups: ["extensions", "apps"]
   resources:
   - deployments
@@ -30,6 +31,7 @@ rules:
   - get
   - list
   - watch
+  - patch
 - apiGroups: ["autoscaling"]
   resources:
   - horizontalpodautoscalers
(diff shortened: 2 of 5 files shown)


roles/default/kiali-deploy/templates/kubernetes/role.yaml
```diff
@@ -21,6 +21,7 @@ rules:
   - get
   - list
   - watch
+  - patch
```
jmazzitelli, line 24: going to need to add all of these to the OLM metadata. I'll add a new github issue for this since we don't have the new manifest directories for the new versions.

jmazzitelli: Need to add "patch" here to the operator role, line 194: https://github.com/kiali/kiali-operator/blob/v1.21/deploy/role.yaml#L194

roles/default/kiali-deploy/templates/kubernetes/role.yaml
```diff
@@ -45,6 +47,7 @@ rules:
...
```
jmazzitelli, line 50: Need to add "patch" here to the operator role, line 218: https://github.com/kiali/kiali-operator/blob/v1.21/deploy/role.yaml#L218

roles/default/kiali-deploy/templates/openshift/role.yaml
```diff
@@ -86,6 +89,7 @@ rules:
   - get
   - list
   - watch
+  - patch
```
jmazzitelli, line 92: Need to add "patch" here to the operator role, line 259: https://github.com/kiali/kiali-operator/blob/v1.21/deploy/role.yaml#L259

roles/default/kiali-deploy/templates/kubernetes/role.yaml
```diff
@@ -30,6 +31,7 @@ rules:
...
```
jmazzitelli, line 34: Need to add "patch" here to the operator role, line 203: https://github.com/kiali/kiali-operator/blob/v1.21/deploy/role.yaml#L203

jmazzitelli:
added https://github.com/kiali/kiali/issues/3020 to remind us to add these new perms to the OLM metadata.

jmazzitelli:
@lucasponce Also, put the verbs in alphabetical order to remain consistent throughout - in all verbs sections have "patch" before "watch". So when you add "patch" verbs, put them before "watch" if "watch" is in the list.

deploy/role.yaml
```diff
@@ -191,6 +191,7 @@ rules:
   verbs:
   - get
   - list
+  ${OPERATOR_ROLE_PATCH}
```
lucasponce, line 194: I used this ${OPERATOR_ROLE_PATCH} to be consistent with the other cases.

roles/default/kiali-deploy/templates/kubernetes/role.yaml
lucasponce: I've added into this PR but you are pointing to a v1.21 version in your comment.
Is this PR to master enough or it is needed to patch in other branch ?

jmazzitelli: This is fine. I only linked to the last release to keep the links from growing stale over time. Just master branch is needed.

deploy/role.yaml
jmazzitelli: Ah, yes. Good catch. This is to support view-only for the operator.

PR #96, opened 2020-07-25
PoC: update the helm chart for people to review

jmazzitelli:
Added some messages at the end of the install that will be important for people reviewing the chart to know.

This is just updating the SNAPSHOT helm chart in the helm repo that will be useful for those in the community testing and reviewing the helm chart.


jotak:
This is not great for review, the changes lie in a binary file... isn't there any text file that triggered them? Should I unpack and compare?

jmazzitelli:
> Or is it just temporary for testing purpose and we won't see these kind of opaque PR in normal times?

This is just temporary - it's for people who want to play with installing Kiali using helm without having to git clone the repo and run builds... like this:

```
helm upgrade --install --version v1.22.0-SNAPSHOT --repo https://kiali.org/kiali-operator/charts kiali-operator kiali-operator
```

We already have a test chart published - this just adds a message at the end explaining to people some important info.

jmazzitelli:
> we won't see these kind of opaque PR in normal times?

you won't see it. The chart yamls will be commited regularly (see my kiali-operator PR to see what they look like). These binaries will be pushed at Jenkins build time to the chart repo location.

PR #98, opened 2020-07-27
Adding defaults for new IstioComponentStatus PR

xeviknal:
Default values for https://github.com/kiali/kiali/pull/3040

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -511,12 +511,12 @@ spec:
 #      component_status:
 #        enabled: true
 #        components:
-#          - app_label: istiod
-#            is_core: true
-#          - app_label: istio-ingressgateway
-#            is_core: true
-#          - app_label: istio-egressgateway
-#            is_core: false
+#        - app_label: istiod
+#          is_core: true
+#        - app_label: istio-ingressgateway
+#          is_core: true
+#        - app_label: istio-egressgateway
+#          is_core: false
 #      istio_identity_domain: "svc.cluster.local"
 #      istio_sidecar_annotation: "sidecar.istio.io/status"
 #      url_service_version: ""
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -448,7 +448,7 @@ spec:
 #   ca_file: The certificate authority file to use when accessing Grafana using https. An empty string means no extra
 #       certificate authority file is used. Default is an empty string.
 #   component_status: Grafana info for the Istio Component Status masthead indicator. Only if it is enabled.
-#       name: Grafana deployment name. It defaults to grafana.
+#       app_label: APP label value of the Grafana pod(s). It defaults to grafana.
 #       is_core: Whether the component is core for your deployment. It defaults to false.
 #   insecure_skip_verify: Set true to skip verifying certificate validity when Kiali contacts Grafana over https.
 #   password: Password to be used when making requests to Grafana, for basic authentication. User only requires viewer permissions. May refer to a secret - see note above.
@@ -498,7 +498,7 @@ spec:
 # component_status:
 #   enabled: Enable/Disable of istio component status into masthead indicator. It defaults to true.
 #   components: A list of components that Kiali will check its statuses.
-#     name: Istio component deployment name.
+#     app_label: Istio component pod app label.
 #     is_core: Whether the component is core for your deployment.
 # istio_identity_domain: The annotation used by Istio to identify domains.
 # istio_sidecar_annotation: The pod annotation used by Istio to identify the sidecar.
@@ -508,11 +508,11 @@ spec:
 #      component_status:
 #        enabled: true
 #        components:
-#          - name: istiod
+#          - app_label: istiod
 #            is_core: true
-#          - name: istio-ingressgateway
+#          - app_label: istio-ingressgateway
 #            is_core: true
-#          - name: istio-egressgateway
+#          - app_label: istio-egressgateway
 #            is_core: false
 #      istio_identity_domain: "svc.cluster.local"
 #      istio_sidecar_annotation: "sidecar.istio.io/status"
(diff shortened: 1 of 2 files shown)


xeviknal:
Thanks @jmazzitelli ! The server side has the default setup, so it works without any value in the ConfigMap:  https://github.com/kiali/kiali/pull/3040/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR367

Do you mind confirming if we are good with those changes? thx

jmazzitelli:
> Do you mind confirming if we are good with those changes? thx

I think that will work. Those defaults you set should take effect if those values are not found in the ConfigMap yaml ([this is the code](https://github.com/kiali/kiali/blob/v1.21/config/config.go#L461-L466) that overrides those defaults with the yaml loaded from the ConfigMap). It's been a while since I played with that server-side config code, but I'm pretty sure that's still how it works.

xeviknal:
@jmazzitelli I've updated the defaults/main.yaml adding the defaults there too so not diverging.

deploy/kiali/kiali_cr.yaml
```diff
@@ -447,6 +447,9 @@ spec:
 # auth: authentication settings to connect to Grafana:
 #   ca_file: The certificate authority file to use when accessing Grafana using https. An empty string means no extra
 #       certificate authority file is used. Default is an empty string.
+#   component_status: Grafana info for the Istio Component Status masthead indicator. Only if it is enabled.
+#       app_label: APP label value of the Grafana pod(s). It defaults to grafana.
```
jmazzitelli, line 451: Is this the app label value or the app label name? I assume it is the value of the label whose name is `istio_labels.app_label_name` from here: https://github.com/kiali/kiali-operator/blob/v1.20.0/deploy/kiali/kiali_cr.yaml#L577-L586 ?
You might want to consider calling this setting something more explicit ("app_label_value") or at least add some more detail to this comment to explain what this is. I *think* (but I am not 100% sure) that "This is the value of the app label whose name is defined in the setting 'istio_labels.app_label_name' ".

deploy/kiali/kiali_cr.yaml
```diff
@@ -492,22 +495,37 @@ spec:
...
 #    istio:
+#      component_status:
+#        enabled: true
+#        components:
+#          - app_label: istiod
+#            is_core: true
+#          - app_label: istio-ingressgateway
+#            is_core: true
+#          - app_label: istio-egressgateway
+#            is_core: false
```
jmazzitelli, line 516: Changes here just to remain consistent with how it is really defined in the defaults/main.yml (the order is reversed):
```suggestion
#          - app_label: istio-egressgateway
#            is_core: false
#          - app_label: istio-ingressgateway
#            is_core: true
#          - app_label: istiod
#            is_core: true
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -532,12 +550,18 @@ spec:
 #        type: "none"
 #        use_kiali_token: false
 #        username: ""
+#      component_status:
+#        app_label: ""
+#        is_core: false
```
jmazzitelli, line 555: Again, these values commented here should be exactly what the true default is for documentation purposes.

```suggestion
#      component_status:
#        app_label: "prometheus"
#        is_core: true
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -568,6 +592,9 @@ spec:
 #        type: "none"
 #        use_kiali_token: false
 #        username: ""
+#      component_status:
+#        app_label: ""
```
jmazzitelli, line 596: values commented here should be exactly what the true default is for documentation purposes.
```suggestion
#        app_label: "jaeger"
```

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -97,6 +97,9 @@ kiali_defaults:
     grafana:
       auth:
         ca_file: ""
+        component_status:
+          app_label: "grafana"
+          is_core: false
```
jmazzitelli, line 102: There should be a corresponding commented section at around line 484 in kiali_cr.yaml to document these defaults.

deploy/kiali/kiali_cr.yaml
xeviknal: I'd prefer to change the defaults to be exact as in here. I'd rather go for this order.

deploy/kiali/kiali_cr.yaml
xeviknal: Good!

deploy/kiali/kiali_cr.yaml
xeviknal: I've approached that. Tried to do a mix of your good suggestion and mine. I think it is now more specific wrt the app_label setting and label name.

deploy/kiali/kiali_cr.yaml
jmazzitelli: That's fine. Just so the order in the list is the same for consistency.

deploy/kiali/kiali_cr.yaml
```diff
@@ -492,22 +498,37 @@ spec:
...
 #    istio:
+#      component_status:
+#        enabled: true
+#        components:
+#          - app_label: istiod
+#            is_core: true
+#          - app_label: istio-ingressgateway
+#            is_core: true
+#          - app_label: istio-egressgateway
+#            is_core: false
```
jmazzitelli, line 519: sorry if I missed this earlier... in every yaml used in the operator everywhere, we never indent the list. Keep it at the same level as the parent (this keeps the lists from growing too wide as you get deeper in the hierarchy).

```suggestion
#        components:
#        - app_label: istiod
#          is_core: true
#        - app_label: istio-ingressgateway
#          is_core: true
#        - app_label: istio-egressgateway
#          is_core: false
```

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -116,9 +119,17 @@ kiali_defaults:
...
     istio:
+      component_status:
+        enabled: true
+        components:
+          - app_label: "istiod"
+            is_core: true
+          - app_label: "istio-ingressgateway"
+            is_core: true
+          - app_label: "istio-egressgateway"
+            is_core: false
```
jmazzitelli, line 130:
```suggestion
        components:
        - app_label: "istiod"
          is_core: true
        - app_label: "istio-ingressgateway"
          is_core: true
        - app_label: "istio-egressgateway"
          is_core: false
```

deploy/kiali/kiali_cr.yaml
xeviknal: Didn't realized. good catch. fixed in my latest commit.

PR #100, opened 2020-07-30
Add / reorganize custom dashboards config

jotak:
- New NamespaceLabel field
- Prometheus' Auth now decoupled from main prometheus config
- Reorganized config for custom metrics

Part of https://github.com/kiali/kiali/issues/3055

Kiali PR: https://github.com/kiali/kiali/pull/3056

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -261,12 +261,6 @@
   when:
   - kiali_vars.external_services.prometheus.url == ""
 
-- name: Set default Prometheus Custom Metrics URL
-  set_fact:
-    kiali_vars: "{{ kiali_vars | combine({'external_services': {'prometheus': {'custom_metrics_url': kiali_vars.external_services.prometheus.url}}}, recursive=True) }}"
-  when:
-  - kiali_vars.external_services.prometheus.custom_metrics_url == ""
-
 # Determine some more defaults.
 
 - name: Set default web root based on cluster type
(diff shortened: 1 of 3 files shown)


roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -261,13 +261,13 @@
...
   - kiali_vars.external_services.prometheus.url == ""
 
-- name: Set default Prometheus Custom Metrics URL
+- name: "Set default namespace label for custom dashboards: needed for Istio 1.7 and later"
   set_fact:
-    kiali_vars: "{{ kiali_vars | combine({'external_services': {'prometheus': {'custom_metrics_url': kiali_vars.external_services.prometheus.url}}}, recursive=True) }}"
+    kiali_vars: "{{ kiali_vars | combine({'external_services': {'custom_dashboards': {'namespace_label': 'kubernetes_namespace'}}}, recursive=True) }}"
   when:
-  - kiali_vars.external_services.prometheus.custom_metrics_url == ""
+  - ???
```
jotak, line 268: @jmazzitelli any idea?

jmazzitelli: This is a new feature we would need to add. AND we need to make sure this version checking works on both upstream Istio and ServiceMesh.

IIRC, we have already have a fact variable to tell us if we are on maistra (aka SM) versus upstream istio, but right now we have no way of knowing what version of either we are on. And remember, Service Mesh reports different version numbers versus upstream Istio. So this won't be trivial.

jotak: OK, then I'm just removing this setup, it will have to be manually configured until perhaps we find a solution.

jotak:
thanks @jmazzitelli can you also take a look at the one on kiali? https://github.com/kiali/kiali/pull/3056 thanks :-)

PR #105, opened 2020-08-01
first attempt at kiali helm chart, for those that refuse to use the operator

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/3066

Install with auth.strategy of anonymous for quick testing/demoing:
```
$ cd kiali-operator # where you pulled this PR branch
$ make .build-hem-chart-server
$ _output/helm-install/helm upgrade --install --namespace istio-system  --set deployment.image_version=latest --set auth.strategy=anonymous kiali-server _output/charts/kiali-server-v1.23.0-SNAPSHOT.tgz
```
(note: pass in `--disable-openapi-validation kiali-server` if installing in OpenShift)

Uninstall:
```
_output/helm-install/helm uninstall --namespace istio-system kiali-server
```


jmazzitelli:
I'm gonna hate myself in the morning, but... here's a PR that has a functional helm chart for the Kiali Server.

The values.yaml (with the exception of 3 settings - two of which can remain unset and 1 is OpenShift only) has the same settings as the operator's Kiali CR `spec` section.

This works on minikube.

Should work on OpenShift EXCEPT "openshift" auth strategy won't work out of box because helm needs to know the Route URL to be able to create the OAuthClient, but the Route is created by the helm chart - chicken and egg. The operator is smart enough to wait for the Route it creates to come up and then obtain the URL. Helm might be able to do this, too, but I really don't want to use those hooks or whatever, and I don't know what the best practice is for something like this. You can get it to work if you know what the Route URL will be ahead of time - just pass in `--set kiali_route_url="the...url...here"`. Also, there appears to be problem in OpenShift 4.5's Route OpenAPI schema - I had to pass in `--disable-openapi-validation kiali-server` to get things to run even with `--dry-run`.

jmazzitelli:
Note - if we do this, it will mean yet ANOTHER set of yaml we have to maintain. If you add a permission to the server, you have to add them to the kiali-operator helm chart, and kiali-operator role, and now here, in addition also to the OLM metadata golden copies. What's one more place, right?

jmazzitelli:
Due to https://github.com/kiali/kiali/issues/3069 the route name has to be `kiali` in order to use `openshift` auth strategy. The change is:

```diff
--- a/deploy/charts/kiali-server/templates/route.yaml
+++ b/deploy/charts/kiali-server/templates/route.yaml
@@ -5,7 +5,9 @@
 apiVersion: route.openshift.io/v1
 kind: Route
 metadata:
-  name: {{ include "kiali-server.fullname" . }}
+  # Cannot use any name here other than "kiali" due to https://github.com/kiali/kiali/issues/3069
+  #name: {{ include "kiali-server.fullname" . }}
+  name: kiali
   namespace: {{ .Release.Namespace }}
```

However, I tried to use `openshift` auth strategy with that fix, and now I hit this error I see in the browser (tried both Chrome and Firefox) when I click the "Login With OpenShift" button on the Kiali login page. I don't know why or how to fix:

```
{"error":"unauthorized_client","error_description":"The client is not authorized to request a token using this method."}
```

jmazzitelli:
This PR is ready for review. Helm chart is working across the board now. Minikube, OpenShift, and any auth-strategy (but see previous comment re: `openshift` strategy and the `kiali_route_url` setting you must set).

jmazzitelli:
To install when published (note - only need to mention the image_version because this chart is published as a SNAPSHOT so you have to tell it not to use that SNAPSHOT version):

```
helm upgrade
  --install \
  --namespace istio-system \
  --version v1.23.0-SNAPSHOT \
  kiali-server/kiali-server \
  https://kiali.org/kiali-operator/charts
```

You can optionally add these options if you want:
```
  --debug \
  --atomic \
  --cleanup-on-fail \
```

jmazzitelli:
> One thing missing is to describe the install and UNINSTALL operations.

I updated my earlier comment to include the uninstall commands.

PR #110, opened 2020-08-05
helm charts have moved to a separate repo

jmazzitelli:
The helm charts moved to a separate rep.

Devs - git clone the helm-charts repo into a directory that is a peer to this kiali-operator repo for a better dev experience.

This needs a backend PR to remove the Jenkins call to the now defunct update-helm-repo target. We need Jenkins to do something different to commit helm charts to the new repo.


jmazzitelli:
We should merge this prior to end of sprint. This includes changes to the deploy script (getLatestKialiOperator) which now pulls from the correct helm repo.

jmazzitelli:
I have some more changes that need to be made. DNM yet

PR #113, opened 2020-08-06
Update config with Istio Injection flag/label

lucasponce:
Required for https://github.com/kiali/kiali/pull/3085 in the context of https://github.com/kiali/kiali/issues/2896

I tried to follow the checklist (it's really helpful), let me know if it's ok or something is missing.

One thing that may need a follow up is that this flag probably want to make it enable in Istio community but disable in Maistra due the way namespaces are labeled is different.

But I didn't address that in this PR.


lucasponce:
@jmazzitelli comments addressed.

PR #122, opened 2020-08-28
more molecule test bulletproofing

jmazzitelli:
Molecule tests are failing on slow-performing clusters. Need to bulletproof some of the tests to try to survive some of the intermittent issues that are occurring.


lucasponce (approved):
Code looks OK but I haven't tested it

PR #128, opened 2020-09-15
Add namespace field in each istio component health

xeviknal:
Requires https://github.com/kiali/kiali/pull/3211

It allows users to specify the namespace of each istio component. It defaults to the `istio_namespace`. This way the user doesn't have to specify each namespace for each component.

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -458,6 +458,7 @@ spec:
 #   component_status: Grafana info for the Istio Component Status masthead indicator. Only if it is enabled.
 #       app_label: The value of the Grafana pod(s)'s app label. The name of the app label is defined in the setting 'istio_labels.app_label_name'. It defaults to grafana.
 #       is_core: Whether the component is core for your deployment. It defaults to false.
+#       namespace: The namespace where the grafana is installed in. It defaults to the 'istio_namespace' setting.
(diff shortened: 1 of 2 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -573,6 +576,7 @@ spec:
 #   component_status: Jaeger info for the Istio Component Status masthead indicator. Only if it is enabled.
 #       app_label: The value of the Jaeger pod(s)'s app label. The name of the app label is defined in the setting 'istio_labels.app_label_name'. It defaults to jaeger.
 #       is_core: Whether the component is core for your deployment. It defaults to false.
+#       namespace: The namespace where Jaeger is installed in. It defaults to the 'istio_namespace' setting.
```
jmazzitelli, line 579: You should at least set these four settings to `""` in the defaults/main.yml so the Kiali CR still sets it. The server can default to what it wants if its "" but we should have all defaults explicitly defined (only in rare cases do we leave values completely undefined).

jmazzitelli (changes requested):

jmazzitelli (dismissed):
lgtm

lucasponce (approved):

lucasponce (changes requested):
Rebase needed

xeviknal:
@lucasponce rebase done

lucasponce (approved):

PR #129, opened 2020-09-16
add new supported version of v1.24

jmazzitelli:
this forks the "default" role and names the new copy "v1.24"

fixes: https://github.com/kiali/kiali/issues/3213

Once merged, people need to backport changes made to default role into the v1.24 role (if you want that feature to be in 1.24).


jshaughn (approved):
LOK

PR #135, opened 2020-09-22
sort the list of accessible namespaces

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/3231


jshaughn (approved):
Despite our inability to reproduce I think this is a valid and needed fix.

PR #136, opened 2020-09-22
latest builds will be named with rhel8 now. (#134)

jmazzitelli:


jshaughn (commented):
OK, v1.24 seems like a reasonable branch name, ties it with the version for SM 2.0.

PR #137, opened 2020-09-22
do not rely on the name of the clusterrole to do a can_i check

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/3227

Unfortunately, because of [a bug in the ansible kubernetes collections](https://github.com/ansible-collections/community.kubernetes/issues/234) we can't do [a true "can_i" check](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#checking-api-access).

So this PR will actually attempt to create dummy both a cluster role and cluster role binding. If they fail, the operator will abort like it did before. Therefore, we no longer rely on the naming of resources - we actually see if the operator has the permissions.

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -430,16 +430,70 @@
     role_kind: "{{ 'ClusterRole' if '**' in kiali_vars.deployment.accessible_namespaces else 'Role' }}"
     role_binding_kind: "{{ 'ClusterRoleBinding' if '**' in kiali_vars.deployment.accessible_namespaces else 'RoleBinding' }}"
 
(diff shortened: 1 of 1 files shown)


Baykonur:
A better solution indeed. Thanks for the PR.

jmazzitelli:
I ran all molecule test with this PR branch, and all passed. So this code looks good.

jmazzitelli:
We can merge this. I plan on removing this implementation and replacing it with a true "can_i" check when that k8s bug is fixed - see issue https://github.com/kiali/kiali/issues/3241

jshaughn (approved):
I guess this works because despite ignoring errors the actual check isn't looking for an error but whether or not the "can_i_create_clusterroles" is set and showing success.  LOK.

PR #144, opened 2020-09-25
Remove ServiceMeshPolicy and SMRbacConfig for Maistra installations

xeviknal:
related to https://github.com/kiali/kiali/pull/3256
requires https://github.com/kiali/kiali/pull/3256


jmazzitelli (changes requested):
Don't forget to go through the checklist: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission

In this case, we'll want to at least change the kiali-ossm golden copy: https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L526-L545

We don't have a newer version of the [community](https://github.com/kiali/kiali-operator/tree/master/manifests/kiali-community) and [upstream](https://github.com/kiali/kiali-operator/tree/master/manifests/kiali-upstream) OLM metadata, so there is nothing to do there yet. Can you write up a github issue to remind us to create two new golden copies for new versions of community and upstream and remind us to also remove those role permissions in those new copies?

jmazzitelli (approved):

xeviknal:
Would it be a good idea to have one folder with the in-progress changes? Like the staging folder we have in the kiali.io.
This way, we could be updating all the changes as we go and the day we need to bump a version for the upstream or community, there won't be the need to track all the changes from the latest version.
If that has been discussed and I've missed that (pretty likely), let's keep it this way. Just in case, the promised issue is here: https://github.com/kiali/kiali/issues/3258

jmazzitelli:
> Would it be a good idea to have one folder with the in-progress changes?

That is hopefully how it will work when the community operator repo moves to the new bundle format. This is why you can make the changes you did in kiali-ossm - because that is essentially one file with "in-progress" changes -- when we build the new kiali-ossm version (via a script we use) it will take that in-progress file as the source for the new version metadata.

But because we can't use that yet for community/upstream (because they don't accept new bundle format), we have to create individual versions of them. It is annoying. But hopefully in at most a couple months this will change. So I don't think we should make any changes now since it will all change again once community/upstream moves to the new bundle version. Then we'll just do it like we do it for kiali-ossm.

lucasponce (approved):

xeviknal:
@jmazzitelli the backend/frontend for this will have to be backported to 1.24. Then, should we remove the permissions in the roles for v1.24 too?
1. https://github.com/kiali/kiali-operator/blob/master/roles/v1.24/kiali-deploy/templates/openshift/role-viewer.yaml#L59-L72
2. https://github.com/kiali/kiali-operator/blob/master/roles/v1.24/kiali-deploy/templates/openshift/role.yaml#L65-L84

I am missing something else in here?

jmazzitelli (changes requested):
> Then, should we remove the permissions in the roles for v1.24 too?

Yes, exactly. You need to change those roles too.

xeviknal:
@jmazzitelli changes approached. thanks

jmazzitelli (approved):
LGTM

lucasponce (approved):

PR #148, opened 2020-09-25
forgot the backport needed to change the v1.24 playbook

jmazzitelli:
the backport PR #143 forgot to change the v1.24 playbook


jshaughn (approved):
probably not needed but LOK

PR #150, opened 2020-09-25
fix component_status

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3261


xeviknal (approved):
LGTM! Grafana component is the only one misplaced in the configs. Namespaces are now consistent too.

PR #151, opened 2020-09-28
[v1.24] updated kiali-ossm CSV metadata

jmazzitelli:
This was already put in master, but we need it in v1.24 branch.


jshaughn (approved):
LOK

PR #159, opened 2020-10-12
Revert "Change kiali-ossm manifest to remove support to maistra crds"

jmazzitelli:
This reverts commit a3697e5f7ad5c7f2c1821b75f68bbfe3da45b0bb.

This is needed because we still want to support installing the older v1.12 which supports the older Istio installations that still has these resources. Without this revert, the operator will fail to install Kiali due to the missing permissions when running with the v1.12 role.

The backport PR: https://github.com/kiali/kiali-operator/pull/160


jshaughn (approved):
Visually Fine.

jshaughn (approved):
Visually Fine.

PR #161, opened 2020-10-13
more notes about developing operator

jmazzitelli:
We hit a problem recently where we broke backwards compatibility with the older supported version of v1.12. This is because I forgot to mention that any changes to the golden copies of the metadata need to be backward compatible with those older versions. In the case that just happened, a permission is no longer needed by the latest Maistra, but is still required by older Maistra versions - so if we install Kiali v1.12 to run with that older Maistra version, we still need the operator to have those permissions so it can install Kiali with those older permissions that are still needed.

Therefore, this PR is adding some notes about this. It also is adding some notes about backporting to older supported versions of the Ansible roles.

### DIFF
--- a/DEVELOPING.adoc
+++ b/DEVELOPING.adoc
@@ -79,11 +79,15 @@ If new Kiali Server resources are added or removed, you must make appropriate ch
 
 Below are checklists that developers should consult when making various kinds of changes.
 
+#### Maintain Backward Compatibility
+
+NOTE: Remember to maintain backward compatiblity for all supported versions. Do **not** make any changes if those changes would break the older supported versions. For example, if you are removing permissions from the operator roles/rolebindings within the golden copies, and if those permissions are required by the older supported versions such as link:./roles/v1.24[`v1.24`], then you must not make those changes. Only when those older versions are no longer supported can you make those changes. In this case, create a github issue as a reminder to make the changes at the appropriate time in the future when those versions are no longer supported.
+
 ### Are You Altering a Kiali Operator Resource?
 
-- [ ] Update the golden copy of the link:./manifests/kiali-upstream[kiali-upstream CSV metadata]
-- [ ] Update the golden copy of the link:./manifests/kiali-community[kiali-community CSV metadata]
-- [ ] Update the golden copy of the link:./manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml[kiali-ossm CSV metadata]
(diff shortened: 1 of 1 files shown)


lucasponce (approved):

DEVELOPING.adoc
```diff
@@ -79,11 +79,15 @@ If new Kiali Server resources are added or removed, you must make appropriate ch
 
 Below are checklists that developers should consult when making various kinds of changes.
 
+#### Maintain Backward Compatibility
```
jotak, line 82: @jmazzitelli just to make sure I understand correctly:
"Backward compatibility" here means having new versions of the **Operator** still working with all supported versions of **Kiali**, right?

You take the example of permissions and roles but I guess the same is true with, say, config changes? If we take the example of what we worked on previously, amending the `custom_dashboard` config: does it mean we should actually have kept the old `custom_prometheus_url` config alive in the CR field (marking it deprecated)? Should we reintroduce it in the documentation?

jmazzitelli: @jotak I used the roles/permissions as an example because that was the one that made me realize this problem. It affected the permissions given to the operator (the operator needed to create a role for the v1.12 Kiali Server, but the operator was no longer given permissions to give out that role because we removed those permissions).

But, in general, yes, this might be something we need to watch out for when changing config. Certain things may not be a problem, but we just have to remember to think about it. And there are certain things we could work around by updating the versioned Ansible roles (i.e. backporting changes to the `v1.0`, `v1.12` and/or `v1.24` Ansible roles as well as changing the `default` one) so long as that version of Kiali Server is also updated to match (so we would need to make changes to the kiali/kiali repo's version branch to fix the server, too). If we backported that `custom_dashboard` work to the v1.12 branch of the kiali/kiali repo, (to fix the Server), then all we would have to do is change the v1.12 Ansible roles to match. Did you backport that custom_dashboard stuff? If not, we should do that OR we should document that custom_prometheus_url is needed in your Kiali CR for older versions.

In that specific case of `custom_prometheus_url`, I think we are OK because the operator itself didn't use or even look at `custom_prometheus_url` in the older playbooks. So the operator didn't need it. And since the operator passes down any config from the Kiali CR to the ConfigMap, you can still put `custom_prometheus_url` directly in your Kiali CR and the Kiali Server will get it in its ConfigMap. So older Kiali Servers that look for it will see it.

PR #164, opened 2020-10-14
turn off dependent resource watching

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3322

backport PR: https://github.com/kiali/kiali-operator/pull/163


jmazzitelli:
> So do we need to update any docs/comments to go with this? This would mean that some changes would not be picked up and could force the user to do what to run a manual reconciliation?

I don't think we ever documented the dependency watching/auto-reconciliation in the first place.

jshaughn (approved):
well ok, then

PR #165, opened 2020-10-14
Add defaults for `disable_rbac` OpenId configuration

israel-hdez:
By default, `disable_rbac` is "off" assuming users usually want RBAC. If users don't want RBAC this is forcing to explicitly turn off RBAC. Basically, we want people to be aware that users logging into Kiali will all share the same privileges.

Related to kiali/kiali#3084

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -139,43 +139,8 @@ spec:
 #    ---
 #    strategy: ""
 #
-# When "openid" strategy is chosen for authentication, you will need to fill some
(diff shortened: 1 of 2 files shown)


jmazzitelli (approved):

cfcosta (approved):

jshaughn:
@mattmahoneyrh We have upstream users waiting for this issue to be resolved and willing to test.  So I am asking @israel-hdez to merge for the 1.26 release.   Does QE have the ability to test OpenID related issues?  It is something we may need to address going forward.

PR #166, opened 2020-10-15
[v1.24] remove buggy checks - these cause problems

jmazzitelli:
Removing is fine, the purpose is just to spit out a more human-friendly error message.

fixes: https://github.com/kiali/kiali/issues/3323


jshaughn (approved):
LOK

PR #170, opened 2020-10-15
[v1.24] convert some vars to bool during conditional checks

jmazzitelli:
This is in case a user specified the bool as a string.

fixes: kiali/kiali#3325


jshaughn (approved):
LOK

PR #174, opened 2020-10-22
k8s kind "extensions/v1beta1 Ingress" is deprecated in favor of "networking.k8s.io/v1beta1 Ingress"

jmazzitelli:
This was reported against the helm chart - but we need to do this in the operator, too.
See https://github.com/kiali/kiali/issues/3343


jshaughn (approved):
LOK

jmazzitelli:
@mattmahoneyrh do you want to review/test this? All you need to test is to install in a k8s cluster (like minikube - NOT OpenShift) and see that an Ingress is installed.

jmazzitelli:
FYI: I tested and it created and deleted the Ingress properly.

PR #175, opened 2020-10-23
Remove Mixer + RBAC + 3Scale deps

lucasponce:
Kiali is going to stop using some old/deprecated APIs so it can be removed from the roles for future versions.
Also, removing 3scale config as this extensions depends on Mixer module.


xeviknal (commented):
I think there are other files to be updated here. Mazz always points to the check-list when a change in this area is done: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission

I think that the helm-charts and the openshift roles needs to be updated too.

hhovsepy (approved):
Verified

roles/default/kiali-deploy/templates/kubernetes/role-viewer.yaml
```diff
@@ -47,10 +47,7 @@ rules:
   - list
   - watch
 - apiGroups:
-  - config.istio.io
   - networking.istio.io
-  - authentication.istio.io
-  - rbac.istio.io
```
jmazzitelli, line 53: I think you also want to do this to the openshift role-viewer.yaml template.  This: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/role-viewer.yaml#L49-L54

roles/default/kiali-deploy/templates/kubernetes/role.yaml
```diff
@@ -50,10 +50,7 @@ rules:
   - patch
   - watch
 - apiGroups:
-  - config.istio.io
   - networking.istio.io
-  - authentication.istio.io
-  - rbac.istio.io
```
jmazzitelli, line 56: I think you also want to do this to the openshift role.yaml template. This: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/role.yaml#L52-L57

jmazzitelli (changes requested):

jmazzitelli (approved):
LGTM

PR #187, opened 2020-10-30
Support logger configuration (format, level)

MarianZoll:


deploy/kiali/kiali_cr.yaml
```diff
@@ -277,6 +277,13 @@ spec:
 #    ---
 #    ingress_enabled: true
 #
+# Determines the logger configuration.
+# log_format supports text and json.
+# log_level supports trace, debug, info, warn, error, fatal.
+#    logger:
```
jmazzitelli, line 283: To match/be consistent with the rest of the settings documented here, put "---" above this line (all settings groups have `---` to make it easier to find the actual settings). So this:
```suggestion
#    ---
#    logger:
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -353,7 +360,7 @@ spec:
 #    ---
 #    tolerations: []
 #
-# Determines which priority levels of log messages Kiali will output.
+# DEPRECATED - Determines which priority levels of log messages Kiali will output.
```
jmazzitelli, line 363: As part of the deprecated message, I would point the reader to the new setting. Something like this:
```suggestion
# DEPRECATED - use the logger.log_level setting.
# Determines which priority levels of log messages Kiali will output.
```

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -58,6 +58,9 @@ kiali_defaults:
     image_pull_secrets: []
     image_version: ""
     ingress_enabled: true
+    logger:
+      log_level: info
+      log_format: text
```
jmazzitelli, line 63: Even though this is valid, for consistency with the rest of the values in this file, please wrap string values in quotes:
```suggestion
      log_level: "info"
      log_format: "text"
```

jmazzitelli (changes requested):
Other than my trivial requests for changes, this looks good from a code review perspective. I have not tried to run this yet.

deploy/kiali/kiali_cr.yaml
jmazzitelli: Also, all the peer settings here are sorted alphabetically. Can you sort so "log_level" is after "log_format".

jmazzitelli (commented):
This looks good. When the other PRs are in their final state, I'll run them for testing.

jmazzitelli:
I got a molecule test failure in the null-cr-values-test - I'm investigating:

```
/tmp/tmp/KIALI-GIT/kiali/operator/molecule/null-cr-values-test/converge.yml:13
fatal: [localhost]: FAILED! => {
   "msg": "The conditional check 'kiali_configmap.deployment.verbose_mode == \"3\"' failed. The error was: error while evaluating conditional (kiali_configmap.deployment.verbose_mode == \"3\"): 'dict object' has no attribute 'verbose_mode'"
```

jmazzitelli:
Right, so verbose_mode has been removed. So we need to clean up the tests as well as some files used by devs:

- [ ] Delete line: https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr_dev_servicemesh.yaml#L20
- [ ] Delete line: https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr_dev.yaml#L25 
- [ ] Delete line: https://github.com/kiali/kiali-operator/blob/master/molecule/null-cr-values-test/converge.yml#L24
- [ ] And replace [that line](https://github.com/kiali/kiali-operator/blob/master/molecule/null-cr-values-test/converge.yml#L24) with checks for the default logger settings.. something like:
```
kiali_configmap.deployment.logger.log_format == "text"
kiali_configmap.deployment.logger.log_level == "info"
```
- [ ] Delete line: https://github.com/kiali/kiali-operator/blob/master/molecule/null-cr-values-test/kiali-cr.yaml#L49
- [ ] And replace [that line](https://github.com/kiali/kiali-operator/blob/master/molecule/null-cr-values-test/kiali-cr.yaml#L49) with setting the logger values to null:
```
logger:
  log_format: null
  log_level: null
```
- [ ] We don't need to set verbose_mode in any of the molecule tests, so remove any mention of it in the individual tests' kiali-cr.yaml files - I found it in two places: [open-id-test](https://github.com/kiali/kiali-operator/blob/master/molecule/openid-test/kiali-cr.yaml#L24), and the [api-test](https://github.com/kiali/kiali-operator/blob/master/molecule/api-test/kiali-cr.yaml#L19)

Finally, there are two places in the main kiali repo we have to clean up. I'll put this in the kiali/kiali PR review, but I'll document it here as well. Delete the setting and usage of VERBOSE_MODE env var in the makefiles. We don't need to set that via env var - devs can manually set it if they need to. So these need to be deleted:

Delete line from make/Makefile.operator.mk: https://github.com/kiali/kiali/blob/master/make/Makefile.operator.mk#L72
Delete line from Makefile: https://github.com/kiali/kiali/blob/master/Makefile#L113

jmazzitelli (changes requested):
See my previous comment with the task list of things that need to be removed and changed.

jmazzitelli (approved):
This is good. Also confirmed molecule test passes. I will put an approval on this PR, but will add a "do not merge" label so we do not merge this until all the other PRs are ready (see https://github.com/kiali/kiali/pull/3382 and https://github.com/kiali/helm-charts/pull/15).

PR #190, opened 2020-11-05
Add web_port to the CR

cfcosta:
Followup to https://github.com/kiali/kiali/pull/3383.

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -738,6 +738,14 @@ spec:
 #    ---
 #    web_fqdn: ""
 #
+# Defines the ingress port where the connections come from. This is usually
+# necessary when the application responds through a proxy/ingress, and it does
+# not forward the correct headers so Kiali can guess the port.
+#
+# When empty, Kiali will try to guess this value from HTTP headers.
+#    ---
+#    web_port: 20001
+#
 # Define the history mode of kiali UI. This can only take
 # two possible values: either "browser" or "hash".
 # When empty, it will always be considered as browser
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -738,6 +738,12 @@ spec:
 #    ---
 #    web_fqdn: ""
 #
+# Define the history mode of kiali UI. This can only take
+# two possible values: either "browser" or "hash".
+# When empty, it will always be considered as browser
+#    ---
+#    web_history_mode: ""
+#
 # Defines the ingress port where the connections come from. This is usually
 # necessary when the application responds through a proxy/ingress, and it does
 # not forward the correct headers so Kiali can guess the port.
@@ -745,12 +751,6 @@ spec:
 # When empty, Kiali will try to guess this value from HTTP headers.
 #    ---
 #    web_port: 20001
-#
-# Define the history mode of kiali UI. This can only take
-# two possible values: either "browser" or "hash".
-# When empty, it will always be considered as browser
-#    ---
-#    web_history_mode: ""
 #
 # Defines the context root path for the Kiali console and API endpoints and readiness probes.
 # When providing a context root path that is not "/", do not add a trailing slash.
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -750,7 +750,7 @@ spec:
 #
 # When empty, Kiali will try to guess this value from HTTP headers.
 #    ---
-#    web_port: 20001
+#    web_port: ""
 #
 # Defines the context root path for the Kiali console and API endpoints and readiness probes.
 # When providing a context root path that is not "/", do not add a trailing slash.
--- a/molecule/null-cr-values-test/kiali-cr.yaml
+++ b/molecule/null-cr-values-test/kiali-cr.yaml
@@ -138,7 +138,7 @@ spec:
     metrics_port: null
     port: null
     web_fqdn: null
-    web_port: null
     web_history_mode: null
+    web_port: null
     web_root: null
     web_schema: null
--- a/molecule/null-cr-values-test/kiali-cr.yaml
+++ b/molecule/null-cr-values-test/kiali-cr.yaml
@@ -138,6 +138,7 @@ spec:
     metrics_port: null
     port: null
     web_fqdn: null
+    web_port: null
     web_history_mode: null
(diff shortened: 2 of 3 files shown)


molecule/null-cr-values-test/kiali-cr.yaml
```diff
@@ -138,6 +138,7 @@ spec:
     metrics_port: null
     port: null
     web_fqdn: null
+    web_port: null
```
israel-hdez, line 141: I think you should move this (and also the other places) to appear after `web_history_mode` to keep alphabetization.
Right, @jmazzitelli ?

israel-hdez (changes requested):

molecule/null-cr-values-test/kiali-cr.yaml
jmazzitelli: correct... a peer settings within the same group should be sorted alphabetically.

cfcosta: Can do, I kept them together because they are intrinsically related (changing both seems to be the correct default), but alphabetizing might be better.

molecule/null-cr-values-test/kiali-cr.yaml
```diff
@@ -8,7 +8,7 @@ spec:
   additional_display_details: null
   installation_tag: null
   istio_component_namespaces: null
-  istio_namespace: {{ istio.control_plane_namespace }}
+  istio_namespace: { { istio.control_plane_namespace } }
```
jmazzitelli, line 11: This file got all screwed up. First, this is incorrect Ansible and won't run. No spaces between the two braces are allowed. Looks like this happened all throughout this file.

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -14,18 +14,18 @@ kiali_defaults:
   version: "default"
 
   additional_display_details:
-  - title: "API Documentation"
-    annotation: "kiali.io/api-spec"
-    icon_annotation: "kiali.io/api-type"
+    - title: "API Documentation"
+      annotation: "kiali.io/api-spec"
+      icon_annotation: "kiali.io/api-type"
```
jmazzitelli, line 19: Something reformated with the wrong indentation. All operator yamls do not "double-indent". All array list items have their "-" marker on the same level as the parent. This bad reformatting is done all throughout this file.

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -45,15 +45,16 @@ kiali_defaults:
     strategy: ""
 
   deployment:
-    accessible_namespaces: ["^((?!(istio-operator|kube.*|openshift.*|ibm.*|kiali-operator)).)*$"]
+    accessible_namespaces:
+      ["^((?!(istio-operator|kube.*|openshift.*|ibm.*|kiali-operator)).)*$"]
```
jmazzitelli, line 49: revert - looks like an IDE wrapped this line.

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -45,15 +45,16 @@ kiali_defaults:
...
     #additional_service_yaml:
     affinity:
       node: {}
       pod: {}
       pod_anti: {}
     custom_dashboards:
-      excludes: ['']
-      includes: ['*']
+      excludes: [""]
+      includes: ["*"]
```
jmazzitelli, line 57: technically OK, but why is this changed?

jmazzitelli (changes requested):
looks like an IDE reformatted this all wrong :)

cfcosta:
@jmazzitelli fixed, weird :(

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -225,6 +225,7 @@ kiali_defaults:
     port: 20001
     web_fqdn: ""
     web_history_mode: ""
+    web_port: 20001
```
jmazzitelli, line 228: Is it true we want to hardcode the default to 20001? I read the description of this value and it says, "When empty, Kiali will try to guess this value from HTTP headers." As I understand, that's the current behavior (attempts to guess the port).

I'm just confirming that we no longer want "guessing" to be the default behavior and that we will now always assume 20001 by default unless the person sets this value to "" in the Kiali CR.

israel-hdez: I think we DO want the default behavior of "guessing" the public port.

I re-checked the changes in the back-end and I see that if "guessing" fails, it fallbacks to the listening port in the pod (which is 20001 by default).

So, I don't think we want a fixed value here. It will break existing installs that are relying on Kiali to "guess" the public port.

deploy/kiali/kiali_cr.yaml
```diff
@@ -744,6 +744,14 @@ spec:
 #    ---
 #    web_history_mode: ""
 #
+# Defines the ingress port where the connections come from. This is usually
+# necessary when the application responds through a proxy/ingress, and it does
+# not forward the correct headers so Kiali can guess the port.
+#
+# When empty, Kiali will try to guess this value from HTTP headers.
+#    ---
+#    web_port: 20001
```
jmazzitelli, line 753:
```suggestion
#    web_port: ""
```

roles/default/kiali-deploy/defaults/main.yml
jmazzitelli, line 228:
```suggestion
    web_port: ""
```

jmazzitelli (changes requested):
default to "guess the port" config value

roles/default/kiali-deploy/defaults/main.yml
jmazzitelli: Note - since this appears to be a numeric value - do we want the default to be `0` or empty string `""`? I don't know what the backend will want.
@israel-hdez @cfcosta

cfcosta:
@jmazzitelli from the code, it's defined as a string, so your suggestions work.

jmazzitelli (approved):
LGTM

israel-hdez:
I'll try this shortly.

israel-hdez (approved):
I'm approving these changes, but rejecting the server side PR.
Sorry about that. See: https://github.com/kiali/kiali/pull/3383#discussion_r522408016

cfcosta:
@israel-hdez no problems! Added a fix there, when it is ok we can merge both together.

PR #194, opened 2020-11-10
[v1.24] add common k8s labels.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3411


jshaughn (approved):
LOK, I hope this was somehow generated and not all by hand.

PR #195, opened 2020-11-11
Remove ComponentStatus from external_services (jaeger, grafana, prom)

xeviknal:
Requires https://github.com/kiali/kiali/pull/3423

It remove the ComponentStatus from jaeger, prom and grafana. It leaves only one param for users tune the severity of the status in the masthead.

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -419,10 +419,7 @@ spec:
 #   use_kiali_token: When true and if auth.type is "bearer", the same OAuth token used for authentication in Kiali will be used for the API calls to Grafana,
 #       and auth.token config is ignored then.
 #   username: Username to be used when making requests to Grafana, for basic authentication. User only requires viewer permissions.
-# component_status: Grafana info for the Istio Component Status masthead indicator. Only if it is enabled.
-#   app_label: The value of the Grafana pod(s)'s app label. The name of the app label is defined in the setting 'istio_labels.app_label_name'. It defaults to grafana.
-#   is_core: Whether the component is core for your deployment. It defaults to false.
-#   namespace: The namespace where Grafana is installed in. It defaults to the 'istio_namespace' setting.
+# core_component: Used in the Components health feature. When true, the unhealthy scenarios will be raised as errors. Otherwise, they will be raised as a warning.
 # dashboards: A list of Grafana dashboards that Kiali can link to. Each item contains:
 #   name: The name of the dashboard in Grafana
 #   variables:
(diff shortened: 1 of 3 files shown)


hhovsepy (approved):

jmazzitelli (changes requested):
Looks like this is removing some settings that the `null-cr-values-test` molecule test is setting.

See: https://github.com/kiali/kiali-operator/blob/master/molecule/null-cr-values-test/kiali-cr.yaml#L72 (and other places). Just make sure that test CR is still valid.

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -152,10 +149,6 @@ kiali_defaults:
       cache_duration: 7
       cache_enabled: true
       cache_expiration: 300
-      component_status:
-        app_label: "prometheus"
-        is_core: true
-        namespace: ""
```
jmazzitelli, line 158: Does Prometheus have a "core_component" field also? If so, it should be specified here and documented in the `kiali_cr.yaml` so users know about it.

xeviknal:
I've updated the molecule test for the nulls replacing the previous flag by the new one.
thank you @hhovsepy and @jmazzitelli.

jmazzitelli (dismissed):
LGTM

deploy/kiali/kiali_cr.yaml
```diff
@@ -419,10 +419,7 @@ spec:
 #   use_kiali_token: When true and if auth.type is "bearer", the same OAuth token used for authentication in Kiali will be used for the API calls to Grafana,
 #       and auth.token config is ignored then.
 #   username: Username to be used when making requests to Grafana, for basic authentication. User only requires viewer permissions.
-# component_status: Grafana info for the Istio Component Status masthead indicator. Only if it is enabled.
-#   app_label: The value of the Grafana pod(s)'s app label. The name of the app label is defined in the setting 'istio_labels.app_label_name'. It defaults to grafana.
-#   is_core: Whether the component is core for your deployment. It defaults to false.
-#   namespace: The namespace where Grafana is installed in. It defaults to the 'istio_namespace' setting.
+# core_component: Used in the Components health feature. When true, the unhealthy scenarios will be raised as errors. Otherwise, they will be raised as a warning.
```
jotak, line 422: I think I prefer with the `is_` prefix to make it clear that a boolean is expected without having to refer to doc

xeviknal:
@jotak @hhovsepy approached this one too. thx

jotak (approved):
LGTM

PR #199, opened 2020-11-18
Fix metrics molecule test (api changed)

jotak:
Fixes https://github.com/kiali/kiali/issues/3443

PS: I could not test; pinging you on irc @jmazzitelli

### DIFF
--- a/molecule/api-test/metrics.test.yaml
+++ b/molecule/api-test/metrics.test.yaml
@@ -18,5 +18,3 @@
   assert:
     that:
     - api_namespaces_metrics_json is defined
-    - api_namespaces_metrics_json.metrics is defined
-    - api_namespaces_metrics_json.histograms is defined


molecule/api-test/metrics.test.yaml
```diff
@@ -18,5 +18,3 @@
   assert:
     that:
     - api_namespaces_metrics_json is defined
-    - api_namespaces_metrics_json.metrics is defined
-    - api_namespaces_metrics_json.histograms is defined
```
jmazzitelli, line 22: There is no other tests you want to add here to test the new format of the metrics results?

jotak: it will depend on what is deployed / what metrics are actually present, I haven't tried to see what's in.
But removing these two lines keep the test equivalent as before, because these "metrics" and "histograms" properties have been merged into the root object. (hence disappeared).

jmazzitelli: ok, will merge this as is. Here's the metrics json the molecule test is seeing for reference: https://github.com/kiali/kiali/issues/3443#issuecomment-729629321

jmazzitelli (approved):
let's merge this as-is to get the tests passing. we can add to it later if we want.

PR #200, opened 2020-11-18
document the fact molecule tests can't run if SELinux is enabled.

jmazzitelli:
It's possible the combination of SELinux and podman 2.x is causing the problem. At the moment, I believe disabling SELinux will work around the problem.


jshaughn (approved):
LOK

PR #201, opened 2020-11-19
let molecule tests empty prometheus data during testing

jmazzitelli:
I noticed a nightly run of the molecule tests failed, and it looked like it might have something to do with some old, extra data that was sitting in Prometheus. This PR allows the molecule tests to purge all data from  Prometheus so a test can start with clean/empty telemetry.

### DIFF
--- a/molecule/api-test/converge.yml
+++ b/molecule/api-test/converge.yml
@@ -13,6 +13,9 @@
       that:
       - kiali_configmap.auth.strategy == "anonymous"
 
+  # Start with empty Prometheus data
+  - import_tasks: ../common/purge-prometheus-data.yml
+
   - import_tasks: create-simple-mesh.yml
 
   - name: Find all API tests
--- a/molecule/common/purge-prometheus-data.yml
+++ b/molecule/common/purge-prometheus-data.yml
@@ -1,0 +1,28 @@
+- name: Purging all data from Prometheus
+  k8s:
+    state: absent
+    api_version: "{{ k8s_item.apiVersion }}"
+    kind: "{{ k8s_item.kind }}"
+    namespace: "{{ k8s_item.metadata.namespace }}"
+    name: "{{ k8s_item.metadata.name }}"
+  with_items:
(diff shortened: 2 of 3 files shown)


jshaughn (approved):
So a pod restart purges Prom data?  Good to know.  I know I've had my prom pod get evicted and restart, Iguess I never noticed that it took the old data with it when it went away.

jmazzitelli:
> So a pod restart purges Prom data?

Look into the Prometheus server pod's `/data` dir (if on minikube, this works; if on OpenShift I believe the container name is just `prometheus` so change the `-c` option appropriately):

```
kubectl exec -n istio-system -it -c prometheus-server $(kubectl get pod -n istio-system -l app=prometheus --no-headers -o custom-columns=N:.metadata.name) -- /bin/ls /data
```

If it has a bunch of data, you will see things like:

```
01EKFZ9SB4VEGJ12Z6CQ1TB95P  01EKSDDTTG2BAA8VQ0976AGD2E  01EMDXA8TK191KNWM12F4ANTGA  chunks_head
01EKHP7Q2N54P3GVVABBQX2JWW  01EKVB7BRY3N5KYGJDWPPSPHXT  01EMDXANY1Y9JZMY46PZ26Q9JH  lock
01EKKM18ENTC9N5Q0X945YGYYC  01EKX90Z4J1A5590VA63WECFBB  01EMEN6M29NT4KSA5XCHQGWHR8  queries.active
01EKNHTR8W4T3W6P2T5A19GK2G  01EKZ6TEW2PE6HTN62B3TXAHHH  01EMEW2BACWJS180T3FVZTV2FT  wal
01EKQFMC2TP4FQ8N8KAY1CNSNX  01EM14KZJ0GJD9CVXPZ8BQAHS6  01EMEW2DQ9Q84NQ0HYPHX7FG0R
```

If it doesn't have data, it will just have:

```
chunks_head  lock  queries.active  wal
```

If you kill the pod like this, you'll see its `/data` goes from that first dir listing to the second:

```
kubectl delete -n istio-system $(kubectl get pods -n istio-system -l app=prometheus -o name)
```

I believe this purges the prometheus data so all timeseries are gone.

PR #208, opened 2020-11-30
[v1.24] disable dependent resource watching

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3482


jmazzitelli:
I tested and can confirm this turns off the multiple reconciliation runs due to dependent resource watching being disabled.

PR #209, opened 2020-11-30
change what the status.accessibleNamespaces field looks like in Kiali CR

jmazzitelli:
Due to bug in ansible task k8s_status, we can't just log the list of namespaces. Use string instead of list for the value type.

fixes: https://github.com/kiali/kiali/issues/3485

If accessible namespaces is less than or equal to 20 namespaces, this will be what the status field looks like:

```
accessibleNamespaces: anamespace,default,istio-system
```

If accessible namespaces is greater than 20 namespaces, this will be what the status field looks like:

```
  accessibleNamespaces: 'Number of accessible namespaces (including control plane
    namespace): 43'
```

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -854,7 +854,11 @@
   - processed_resources.configmap.changed == True
   - processed_resources.configmap.method == "patch"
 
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LOK (small typo in comment, I'll let it go)

PR #211, opened 2020-12-01
pass custom helm --set options to deploy script; turn on profiler in molecule via helm 

jmazzitelli:
part of fix to https://github.com/kiali/kiali/issues/3489


jshaughn (approved):
LOK

PR #214, opened 2020-12-02
speed up monitoring dashboard creation

jmazzitelli:
part of performance enhancement issue https://github.com/kiali/kiali/issues/3469

Quick test on minikube shows this task now takes roughly 3.5 seconds versus originally taking 25s. Much faster.

We need to backport this to v1.24 also.

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -832,8 +832,12 @@
   k8s:
     state: "present"
     namespace: "{{ kiali_vars.deployment.namespace }}"
-    definition: "{{ lookup('template', item) }}"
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
Arcane but LOK

jotak:
waw .. do you know why there's such a difference?
That's what I don't like with the declarative programming, it's sometimes hard to understand what happens under the cover ... and the consequences

jmazzitelli:
> waw .. do you know why there's such a difference?

Each invocation of the k8s task requires a new connection to the k8s master API server. I believe the connection/login for each new resource to create is where the extra time comes in.

(IIRC, the folks in charge of this library might be working on a connection cache to help with a problem like this. But I'm not sure)

PR #217, opened 2020-12-03
Create a filter plugin to speed up the portion of the ansible script that determines the accessible namespaces

jmazzitelli:
part of performance analysis issue https://github.com/kiali/kiali/issues/3469


jshaughn (approved):
LOK, but now that we're escaping to python for optimization I'm afraid things are going to get too complicated.  Perhaps you've done enough :)

PR #220, opened 2020-12-11
fix molecule image to workaround k8s py library bug

jmazzitelli:
There is a bug in the 12.x version of the kubernetes py library. Force downgrade to 11.0.0 which works.

See this github issue explaining the bug: https://github.com/kubernetes-client/python/issues/1333

We will want to revert this when that bug is fixed so we can upgrade the py library. For now, we don't need 12.x so using 11.0.0 is fine.

In order to pick up the fix if you've already build the kiali-molecule image, run these two commands:

```
docker rmi kiali-molecule:latest
make molecule-build
```

That should rebuild the image. Run this to confirm you have 11.0.0 of the library:

```
$ docker run --rm -i -t kiali-molecule:latest pip list | grep kubernetes
kubernetes                    11.0.0
```

### DIFF
--- a/molecule/docker/Dockerfile
+++ b/molecule/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM quay.io/ansible/molecule:3.0.6
 RUN apk add gettext gcc libc-dev
-RUN pip install openshift junit-xml jmespath docker kubernetes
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LOK, forcing the earlier version of python, I guess.

mattmahoneyrh (approved):

PR #222, opened 2020-12-13
optimize removal of monitoring dashboards

jmazzitelli:

backport PR: https://github.com/kiali/kiali-operator/pull/223


jmazzitelli:
I ran the config-values-test molecule test and it passes, indicating this code works.

jmazzitelli:
some of the changes here are because we updated the v1.24 playbook in the v1.24 branch but NOT in the master branch. This fixes that.

jshaughn (approved):
LOK, we need to be careful of the order we do these, with backports coming second.

PR #223, opened 2020-12-13
[v1.24] optimize removal of monitoring dashboards

jmazzitelli:


jshaughn (approved):
LOK

jmazzitelli:
I can confirm master branch and v1.24 branch have identical roles/v1.24 directories - which is what we want. So this PR with PR https://github.com/kiali/kiali-operator/pull/222 puts things back to the way they should be.

PR #227, opened 2020-12-15
No need to check for the istiod endpoint at all - assume it is there.

jmazzitelli:
Since v1.24 does not support 1.5 or earlier, we remove this from there as well.


jshaughn (commented):
Right, makes sense.

jshaughn (approved):

PR #228, opened 2020-12-15
[v1.24] No need to check for the istiod endpoint at all - assume it is there.

jmazzitelli:
Since v1.24 does not support 1.5 or earlier, we remove this from there as well.

this is a backport of https://github.com/kiali/kiali-operator/pull/227


jshaughn (approved):
LOK

PR #231, opened 2021-01-07
bump quay.io/operator-framework/ansible-operator to v0.17.2

morlay:
fix https://github.com/kiali/kiali/issues/3561 

v0.17.2 fix arm64 issue https://github.com/operator-framework/operator-sdk/pull/3291

PR #232, opened 2021-01-07
fix operator on ARM64 by bumping operator sdk

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3561

This is PR https://github.com/kiali/kiali-operator/pull/231 with an additional commit that is needed. Github won't let me push with my commit until someone approves so I am creating this PR and will close that other one so we can move this along.


jshaughn (approved):
LOK

PR #234, opened 2021-01-08
Added header to options for auth.strategy

mlbiam:
to support https://github.com/kiali/kiali/issues/3207


jmazzitelli (changes requested):
Need to add a little comment in the [example kiali_cr.yaml file here](https://github.com/kiali/kiali-operator/blob/v1.28.1/deploy/kiali/kiali_cr.yaml#L120-L134) to document this new strategy value.

mlbiam:
added doc

jmazzitelli (changes requested):
I just realized (after reviewing the helm chart PR and remembering we need to add perms to the operator) that the operator has to be given this permission. So the OLM metadata has to be updated with the new permission.

Also need to update the templates in the operator ansible playbook for the kiali role/role_viewer for k8s and openshift.

Read the DEVELOPING.adoc section on "[Are you altering a kiali server role permission](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission)" for details on what you need to do.

mlbiam:
updates per the checklist

jmazzitelli (changes requested):
Almost there. As per [checklist](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission), you are missing these (the role_viewer templates):

* Update the [Operator Kubernetes role-viewer.yaml](https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/role-viewer.yaml)
* Update the [Operator OpenShift role-viewer.yaml](https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/role-viewer.yaml)

I am under the impression that this "create" permission is always required (whether or not the admin installed Kiali as view-only) because this is required to even be able to log into Kiali using a reverse proxy. If this is true, then we need those role-viewer templates updated, too.

jmazzitelli (approved):
LGTM

jmazzitelli (changes requested):
need to fix the openid-test before merging. I'm checking to see why it is failing

jmazzitelli (approved):
> need to fix the openid-test before merging. I'm checking to see why it is failing

I'm going to re-approve this PR since the crash is a panic caused by server-side code. This PR and the openid-test code should all be OK. So I will approve this PR since it, itself, seems OK. The server side PR is still blocked until we fix the nil-pointer reference panic.

jmazzitelli (changes requested):
Request the molecule test PR get merged into this PR first. See: https://github.com/mlbiam/kiali-operator/pull/1

jmazzitelli (approved):
I am approving. Once all the PRs are "approved" and ready to merge, I am going to run the unit tests and molecule tests one more final time and when I see they all pass I will merge everything.

PR #237, opened 2021-01-19
Kiali metrics: add "hits rate" charts and "Service" label

jotak:
Screenshot:
Capture d’écran de 2021-01-19 09-55-37

This is a prequel to https://github.com/kiali/kiali/issues/3514 (I want some metrics & charts easily available to measure performance gain prior to that task)

@jmazzitelli I've only modified one dashboard, is it correct that it's the main one, and the future versioned ones will be derived from there? There's no need to modify the already released ones I think.

### DIFF
--- a/roles/default/kiali-deploy/templates/dashboards/kiali.yaml
+++ b/roles/default/kiali-deploy/templates/dashboards/kiali.yaml
@@ -10,6 +10,15 @@ spec:
   title: Kiali Internal Metrics
   items:
   - chart:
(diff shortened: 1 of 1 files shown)


jotak:
Just realized this didn't go in, @jmazzitelli perhaps you missed the notification?

jmazzitelli (approved):
when we create future versions, yes, we copy these "default" templates.

approving this PR

jotak:
thanks!

PR #239, opened 2021-01-21
openshift-auth-test molecule test

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3608

This molecule test requires that you set up the htpasswd identity provider in OpenShift so it has a user `kiali` with password `kiali`.

You can set this up by logging into an OpenShift cluster via `oc` and then running `make cluster-add-users`.


jshaughn (approved):
This is excellent to have.  Thanks @jmazzitelli .

PR #242, opened 2021-02-01
molecule test now makes sure logout works

jmazzitelli:
part of https://github.com/kiali/kiali/issues/3595


jshaughn (approved):
Love that we have molecule tests for more and more scenarios.

PR #249, opened 2021-02-11
Jaeger config flag: use_grpc

jotak:
Fixes https://github.com/kiali/kiali/pull/3601

Kiali PR: https://github.com/kiali/kiali/pull/3601

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -580,6 +580,9 @@ spec:
 # url: External URL that will be used to generate links to Jaeger. It must be accessible to clients external to
 #      the cluster (e.g: browser) in order to generate valid links.
 #      If tracing service is deployed in a QUERY_BASE_PATH set this in the url like https://<hostname>/<QUERY_BASE_PATH> . EX: https://tracing-service:8080/jaeger
+# use_grpc: Set "true" to enable GRPC connection between Kiali and Jaeger, in order to speed up the queries. In some setups you might not be able to use
+#      GRPC (e.g. if Jaeger is behind some reverse proxy that doesn't support it).
+#      If not specified, it will be false if deployed within a Maistra/OSSM+OpenShift environment, true otherwise.
 # whitelist_istio_system: Set whitelist services in istio-system to check their traces
 #    ---
 #    tracing:
@@ -596,6 +599,7 @@ spec:
 #      in_cluster_url: ""
 #      namespace_selector: true
 #      url: ""
+#      #use_grpc:
 #      whitelist_istio_system: ["jaeger-query", "istio-ingressgateway"]
 
 ##########
--- a/deploy/kiali/kiali_cr_dev.yaml
+++ b/deploy/kiali/kiali_cr_dev.yaml
@@ -24,3 +24,5 @@ spec:
     image_version: $KIALI_IMAGE_VERSION
     namespace: $NAMESPACE
     service_type: $SERVICE_TYPE
+    logger:
+      log_level: info
--- a/deploy/kiali/kiali_cr_dev_servicemesh.yaml
+++ b/deploy/kiali/kiali_cr_dev_servicemesh.yaml
@@ -17,6 +17,8 @@ spec:
     ingress_enabled: true
     namespace: "${NAMESPACE}"
     service_type: "${SERVICE_TYPE}"
+    logger:
+      log_level: info
   external_services:
     grafana:
       auth:
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -173,6 +173,7 @@ kiali_defaults:
       in_cluster_url: ""
       namespace_selector: true
       url: ""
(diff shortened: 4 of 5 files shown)


jotak:
draft => need to tweak default considering target (ossm/istio)

deploy/kiali/kiali_cr.yaml
```diff
@@ -596,6 +598,7 @@ spec:
 #      in_cluster_url: ""
 #      namespace_selector: true
 #      url: ""
+#      use_grpc: false
```
jmazzitelli, line 601:
```suggestion
#      #use_grpc:
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -580,6 +580,8 @@ spec:
 # url: External URL that will be used to generate links to Jaeger. It must be accessible to clients external to
 #      the cluster (e.g: browser) in order to generate valid links.
 #      If tracing service is deployed in a QUERY_BASE_PATH set this in the url like https://<hostname>/<QUERY_BASE_PATH> . EX: https://tracing-service:8080/jaeger
+# use_grpc: Set "true" to enable GRPC connection between Kiali and Jaeger, in order to speed up the queries. In some setups you might not be able to use
+#      GRPC (e.g. if Jaeger is behind some reverse proxy that doesn't support it).
```
jmazzitelli, line 584:
```suggestion
#      GRPC (e.g. if Jaeger is behind some reverse proxy that doesn't support it).
#      If not specified, it will be true if deployed within a Maistra+OpenShift environment, false otherwise.
```

jmazzitelli (changes requested):
> draft => need to tweak default considering target (ossm/istio)

This should be simple. You want to add a task below here: https://github.com/kiali/kiali-operator/blob/v1.29.1/roles/default/kiali-deploy/tasks/main.yml#L259-L260

Just add something like this (I assume the default value should be true if OSSM and false otherwise):

```yaml
- name: Set default tracing use_grpc setting
  set_fact:
    kiali_vars: "{{ kiali_vars | combine({'external_services': {'tracing': { 'use_grpc': is_maistra }}}, recursive=True) }}"
  when:
  - kiali_vars.external_services.tracing.use_grpc is not defined
```

Note that because we want to set its default based on runtime information (is_maistra) do not define this in the defaults/main.yml (otherwise, we have no way to know if it was not set by the user or by the hardcoded defaults). So just change the defaults/main.yml and the example kiali_cr.yaml as I suggest here (following the example of some other settings where we do the same kind of thing - e.g. `identity.cert_file`).

deploy/kiali/kiali_cr.yaml
FilipB: "If not specified, it will be true if deployed within a Maistra+OpenShift environment, false otherwise."
We need to set use_grpc to false for Service mesh.

jotak: right, it's the opposite. I'll commit mazz' suggestions then update

deploy/kiali/kiali_cr_dev.yaml
```diff
@@ -24,3 +24,5 @@ spec:
     image_version: $KIALI_IMAGE_VERSION
     namespace: $NAMESPACE
     service_type: $SERVICE_TYPE
+    logger:
+      log_level: info
```
jmazzitelli, line 28: can you add this to the service mesh version of this file also to maintain consistency across the two dev files? (that version of the file is used when you use make to install Kiali in an OSSM environment). See https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr_dev_servicemesh.yaml

jmazzitelli (changes requested):

deploy/kiali/kiali_cr_dev.yaml
jotak: sure, done

jmazzitelli (approved):

jotak:
@FilipB
I tested this on my istio upstream: by default, grpc is enabled; set `use_grpc` to false in CR => grpc is disabled.
I couldn't test with maistra (having some troubles to set up my environment). We expect the opposite default behaviour.

To see if the connection is actually http1 or grpc you can increase the logs level (under `deployment` in CR)
```
    logger:
      log_level: trace
```

then grep for "Jaeger", you should see either
```
2021-02-12T10:04:38Z TRC Jaeger GRPC client info: address=tracing.istio-system:80, auth.type=none
```
or
```
2021-02-12T10:06:54Z TRC Using legacy HTTP client for Jaeger: url=http://tracing.istio-system/jaeger, auth.type=none
```

FilipB (approved):
Tested following:
- grpc is disabled by default for service mesh and enabled for istio
- use_grpc flag works as expected
- grpc client (use_grpc=true) is not working for service mesh (which is expected - ERR GetAppTraces, Jaeger GRPC client error: rpc error: code = Unavailable desc = connection closed)
- it works fine for both use_grpc=true/false in istio
- tested with both all-in-one and production-elasticsearch jaeger templates

PR #252, opened 2021-02-15
fix the Ingress apiVersion so it isn't a deprecated version

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3706

### DIFF
--- a/molecule/instance-name-test/converge.yml
+++ b/molecule/instance-name-test/converge.yml
@@ -60,7 +60,7 @@
       apiRoBi: "rbac.authorization.k8s.io/v1"
       apiServ: "v1"
       apiSvcA: "v1"
-      apiIngr: "networking.k8s.io/v1"
+      apiIngr: "networking.k8s.io/{{ 'v1' if (lookup('k8s', kind='Ingress', api_version='networking.k8s.io/v1', errors='ignore') is iterable) else 'v1beta1' }}"
       apiRout: "route.openshift.io/v1"
 
   - name: Kubernetes - Make sure we have the basic resources we expect, with the labels we expect
--- a/molecule/instance-name-test/converge.yml
+++ b/molecule/instance-name-test/converge.yml
@@ -60,7 +60,7 @@
       apiRoBi: "rbac.authorization.k8s.io/v1"
       apiServ: "v1"
       apiSvcA: "v1"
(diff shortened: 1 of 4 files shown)


jmazzitelli:
merge this first: https://github.com/kiali/kiali-operator/pull/251 (this PR is built on top of that one)

jmazzitelli:
Hmm... this v1 version does not look to be available in k8s versions before 1.19. Before merging, need to figure out if this is a problem. I suspect it is - I think k8s versions 1.18 and lower might still be used and not EOL'ed.

jmazzitelli:
Moved to draft - only merge this when we know we only support k8s environments at the 1.19 version and up.

jmazzitelli:
don't forget to change some hack scripts that use `extensions/v1beta1` - change to `networking.k8s.io/v1`

https://github.com/kiali/kiali/search?q=extensions%2Fv1beta1

In addition, anything using https://raw.githubusercontent.com/maistra/istio/maistra-2.0/samples/bookinfo/platform/kube/bookinfo-ingress.yaml needs to change because it, too, uses the deprecated version.

https://github.com/kiali/kiali/search?q=bookinfo-ingress

jmazzitelli:
molecule tests pass

molecule/instance-name-test/converge.yml
```diff
@@ -60,7 +60,7 @@
       apiRoBi: "rbac.authorization.k8s.io/v1"
       apiServ: "v1"
       apiSvcA: "v1"
-      apiIngr: "networking.k8s.io/v1beta1"
+      apiIngr: "networking.k8s.io/{{ 'v1' if (lookup('k8s', kind='Ingress', api_version='networking.k8s.io/v1', errors='ignore') is iterable) else 'v1beta1' }}"
```
nrfox, line 63: This keeps backwards compatibility with 1.18 for the time being as v1beta1 is the fallback right?

jmazzitelli: correct... if that lookup failed to find "v1" that if-stmt will fallback to v1beta1 (which is what we used before - so this is backwards compatible).

nrfox (approved):
LGTM and works on my setup

pbajjuri20 (approved):
I run the regression test, everything works perfectly on OpenShift. LGTM

PR #253, opened 2021-02-16
Enable envoy dashboards

jotak:
Fixes https://github.com/kiali/kiali/issues/2582

### DIFF
--- a/roles/default/kiali-deploy/templates/dashboards/envoy.yaml
+++ b/roles/default/kiali-deploy/templates/dashboards/envoy.yaml
@@ -8,7 +8,7 @@ metadata:
     app.kubernetes.io/part-of: kiali
 spec:
   title: Envoy Metrics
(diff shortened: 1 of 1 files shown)


roles/default/kiali-deploy/templates/dashboards/envoy.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     app.kubernetes.io/part-of: kiali
 spec:
   title: Envoy Metrics
-#  discoverOn: "envoy_server_uptime"
+  discoverOn: "envoy_server_uptime"
```
jmazzitelli, line 11: I believe this needs an update too (the helm chart templates): https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/dashboards/envoy.yaml#L11

jmazzitelli (approved):

PR #255, opened 2021-02-16
Fix the accessible namespace regex check

jmazzitelli:
If you put these two lines at the bottom of only_accessible_namespaces.py and run it with "python":

```
print (only_accessible_namespaces(['2alpha', 'alpha', 'alpha2', 'beta'], ['alpha', 'beta']))
print (only_accessible_namespaces(['2alpha', 'alpha', 'alpha2', 'beta'], ['alpha.*', 'beta']))
```

With the bug, both will print out `alpha` and `alpha2`.
With the bug fixed, the first will only print out `alpha`.

fixes: https://github.com/kiali/kiali/issues/3709

### DIFF
--- a/molecule/accessible-namespaces-test/converge.yml
+++ b/molecule/accessible-namespaces-test/converge.yml
@@ -13,31 +13,53 @@
     assert:
       that:
       - kiali_configmap.api.namespaces.label_selector is not defined
+
   # change to accessible_namespaces to a fixed list of namespaces
   - k8s:
       state: present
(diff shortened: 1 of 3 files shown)


FilipB (approved):
Tested with 3 bookinfo apps. Also tried that the regex bookinfo.* is still working.

jmazzitelli:
I added a molecule test - needs another approval

jshaughn (approved):
LOK

PR #256, opened 2021-02-16
[1.24] Fix accessible namespace regex

jmazzitelli:
This is a backport of https://github.com/kiali/kiali-operator/pull/255


FilipB (approved):
Tested with 3 bookinfo apps. Also tried that the regex bookinfo.* is still working.

PR #257, opened 2021-02-16
We don't need to mention Ingress in the metadata, remove it to save work in the future

jmazzitelli:
At some point in the near future, the Ingress version is going to change. Rather than make more work in the future, just take it out. This will allow us to maintain a simple set of PRs for https://github.com/kiali/kiali/issues/3706 that we can merge in the future without having to worry about future conflicts.

related to https://github.com/kiali/kiali/issues/3706


jshaughn (approved):
LOK

PR #260, opened 2021-03-10
Adding daemonsets permissions to kiali

xeviknal:


manifests/kiali-community/1.31.0/kiali.v1.31.0.clusterserviceversion.yaml
```diff
@@ -493,6 +493,7 @@ spec:
           - post
         - apiGroups: ["extensions", "apps"]
           resources:
+          - daemonsets
```
jmazzitelli, line 496: This will need to be reverted. We already released 1.31.0 and it did not have this permission. The 1.31.0 manifest should be locked down and not changed (I have to create the necessary PRs over in OLM land to publish 1.31.0 to operatorhub.io - that has not been done yet).

So you need to back out this line change (as well as the one in kiali-upstream). We then have to create a new set of manifest files and call it 1.32.0 (we have [a script](https://github.com/kiali/kiali-operator/blob/master/manifests/create-new-version.sh) to help do this) and then make the change there. If you want to wait until I'm back off PTO on Monday, I can do that.

```suggestion
```

manifests/kiali-upstream/1.31.0/kiali.v1.31.0.clusterserviceversion.yaml
```diff
@@ -493,6 +493,7 @@ spec:
           - post
         - apiGroups: ["extensions", "apps"]
           resources:
+          - daemonsets
```
jmazzitelli, line 496: We can't do this - see my previous comment for why. Need to revert this.
```suggestion
```

jmazzitelli (changes requested):
Don't forget to update the helm charts as well. See: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission

xeviknal:
Good catch. I'll check how a new release is entered and try to replicate. I'll create a PR in that case and definitely wait for you approval. thanks!

jmazzitelli:
> I'll check how a new release is entered and try to replicate

Here's how you can do it:

```
cd <kiali-operator-local-repo>/manifests
./create-new-version.sh -ov 1.31.0 -nv 1.32.0 -om kiali-community/
./create-new-version.sh -ov 1.31.0 -nv 1.32.0 -om kiali-upstream/
```

At that point, you should have the new 1.32.0 manifests - you can then update those files with the new permissions.

Note that you should still update the ossm one like you did before (in fact, the rest of this PR was fine - it was just those 2 lines in the 2 community/upstream manifests that were a problem) .. the ossm manifest tracks the permissions needed by head of master

PR #263, opened 2021-03-11
Add clarifications for the istio component status

xeviknal:
Adding clarification for the istio component status. Based in the confusion thread (bink):
https://github.com/kiali/kiali/discussions/3759#discussioncomment-462988

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -53,7 +53,7 @@ spec:
 #    prometheus: prom-ns
 # means Prometheus is installed in the namespace "prom-ns" but all other
 # Istio components are installed in the namespace defined in istio_namespace.
-# NOTE: You can find Istio Components Status settings under `external_services`. This is only used by the operator.
+# NOTE: You can find Istio Components Status settings under `external_services`.
 #  ---
 #  istio_component_namespaces: {}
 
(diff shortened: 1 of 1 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -53,6 +53,7 @@ spec:
 #    prometheus: prom-ns
 # means Prometheus is installed in the namespace "prom-ns" but all other
 # Istio components are installed in the namespace defined in istio_namespace.
+# NOTE: You can find Istio Components Status settings under `external_services`. This is only used by the operator.
```
jmazzitelli, line 56: cc @jshaughn can you confirm this part? I actually thought some code you wrote in the server also uses this `istio_component_namespaces` setting. But I am not familiar. So I will add you as a reviewer to check this. It could just mean we remove the "This is only used by the operator." part - that's the part that I am not sure is true.

xeviknal: @jmazzitelli you are right. It's been used in the kiali server. I misread the comment in the discussion. Removing this.

xeviknal: Leaving only the first part: 

> NOTE: You can find Istio Components Status settings under `external_services`

jmazzitelli (approved):
lgtm

deploy/kiali/kiali_cr.yaml
jshaughn: I'm not exactly sure but I think there is code in the graph-gen that wants to know if the selected namespace is an "istio" namespace because then, presumably, we shouldn't expect them to have sidecars, or something like that.  We should revisit this because it may date back to Mixer.

PR #264, opened 2021-03-11
Annotate Kiali Service with external URL settings

israel-hdez:
Given the following settings in the Kiali CR:

```yaml
spec:
    server:
      web_fqdn: example.domain.com
      web_port: 8443
      web_root: /dashboard
      web_schema: https
```

This is an example of the resulting Service:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    kiali.io/api-spec: https://kiali.io/api
    kiali.io/api-type: rest
    kiali.io/external-url: https://example.domain.com:8443/dashboard # This is the new annotation
    operator-sdk/primary-resource: kiali-operator/kiali
    operator-sdk/primary-resource-type: Kiali.kiali.io
  labels:
    app: kiali
    app.kubernetes.io/name: kiali
    app.kubernetes.io/part-of: kiali
    app.kubernetes.io/version: dev
    version: dev
spec:
  clusterIP: 172.21.32.170
  ports:
  - name: http
    port: 20001
    protocol: TCP
    targetPort: 20001
  - name: http-metrics
    port: 9090
    protocol: TCP
    targetPort: 9090
  selector:
    app: kiali
    version: dev
  sessionAffinity: None
  type: ClusterIP
```

Related front-end PR: https://github.com/kiali/kiali-ui/pull/2102

Related to kiali/kiali#3526

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -796,6 +796,8 @@ spec:
 # of the URL (usually it's a fully-qualified domain name).
 # For example, "kiali.example.org".
 # When empty, Kiali will try to guess this value from HTTP headers.
+# You must populate this value if you want to enable cross-linking between Kiali
+# instances in a multi-cluster setup.
 #    ---
 #    web_fqdn: ""
 #
@@ -823,5 +825,7 @@ spec:
 # Defines the public HTTP schema used to serve Kiali. This can only take
 # two possible values: either "http" or "https".
 # When empty, Kiali will try to guess this value from HTTP headers.
+# You must populate this value if you want to enable cross-linking between Kiali
+# instances in a multi-cluster setup.
 #    ---
 #    web_schema: ""
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -796,8 +796,8 @@ spec:
 # of the URL (usually it's a fully-qualified domain name).
 # For example, "kiali.example.org".
 # When empty, Kiali will try to guess this value from HTTP headers.
-# You must populate this value if you want to enable cross-linking between Kiali
-# instances in a multi-cluster setup.
+# On non-OpenShift clusters, you must populate this value if you want to enable
+# cross-linking between Kiali instances in a multi-cluster setup.
 #    ---
 #    web_fqdn: ""
 #
@@ -825,7 +825,7 @@ spec:
 # Defines the public HTTP schema used to serve Kiali. This can only take
 # two possible values: either "http" or "https".
 # When empty, Kiali will try to guess this value from HTTP headers.
-# You must populate this value if you want to enable cross-linking between Kiali
-# instances in a multi-cluster setup.
+# On non-OpenShift clusters, you must populate this value if you want to enable
+# cross-linking between Kiali instances in a multi-cluster setup.
 #    ---
 #    web_schema: ""
--- a/roles/default/kiali-deploy/templates/kubernetes/service.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/service.yaml
@@ -12,6 +12,9 @@ metadata:
   annotations:
     kiali.io/api-spec: https://kiali.io/api
     kiali.io/api-type: rest
+{% if kiali_vars.server.web_fqdn|length != 0 and kiali_vars.server.web_schema|length != 0 %}
+    kiali.io/external-url: {{ kiali_vars.server.web_schema + '://' + kiali_vars.server.web_fqdn + ((':' + kiali_vars.server.web_port | string) if (kiali_vars.server.web_port | string | length != 0) else '') + (kiali_vars.server.web_root | default('')) }}
+{% endif %}
 {% if kiali_vars.deployment.service_annotations|length > 0 %}
     {{ kiali_vars.deployment.service_annotations | to_nice_yaml(indent=0) | trim | indent(4) }}
 {% endif %}


roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -355,6 +355,30 @@
   - is_openshift == True
   - kiali_vars.external_services.tracing.auth.ca_file is not defined or kiali_vars.external_services.tracing.auth.ca_file == ''
 
+# If the CR specifies Kiali's external URL, annotate the Kiali service with that url
+# This enables cross-linking between Kiali instances in multi-cluster setups.
+- name: Add external-url annotation to service
+  set_fact:
+    kiali_vars: |
+      {{ kiali_vars | combine({
```
jmazzitelli, line 363: I think we need a molecule test for this. Have you tested to see what would happen if the user already set `spec.deployment.service_annotations` with their own custom annotations?

While this seems like a clever solution, this is going to put this value in the ConfigMap that will look different than the value the user specified in the CR. This is not entirely bad, but it will be confusing. We are essentially adding an annotation to their configuration which the user did not add.

I think a better way to do this would be to put this directly in the Service templates
* https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/service.yaml#L15
* https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/service.yaml#L14

Something like:

```
  annotations:
    service.beta.openshift.io/serving-cert-secret-name: kiali-cert-secret
    kiali.io/api-spec: https://kiali.io/api
    kiali.io/api-type: rest
{% if kiali_vars.server.web_fqdn|length > 0 and kiali_vars.server.web_schema|length > 0 %}
    kiali.io/external-url: {{ kiali_vars.server.web_schema + '://' + kiali_vars.server.web_fqdn + ((':' + kiali_vars.server.web_port | string) if (kiali_vars.server.web_port | string | length != 0) else '') + (kiali_vars.server.web_root | default(''))
{% endif %}
{% if kiali_vars.deployment.service_annotations|length > 0 %}
    {{ kiali_vars.deployment.service_annotations | to_nice_yaml(indent=0) | trim | indent(4) }}
{% endif %}
```

jmazzitelli (commented):
See my suggested solution. I'm not crazy about adding an annotation to the config - I can't put my finger on why, but I get a sense something isn't right about it. I know that that annotation will be added to the ConfigMap and will be confusing to the user, but I'm also not sure I like us injecting config settings into the ConfigMap that the user here didn't set.

roles/default/kiali-deploy/tasks/main.yml
israel-hdez: ouch! Yes, your solution is much cleaner.
I'll change this PR.

roles/default/kiali-deploy/templates/kubernetes/service.yaml
```diff
@@ -12,6 +12,9 @@ metadata:
   annotations:
     kiali.io/api-spec: https://kiali.io/api
     kiali.io/api-type: rest
+{% if kiali_vars.server.web_fqdn|length != 0 and kiali_vars.server.web_schema|length != 0 %}
+    kiali.io/external-url: {{ kiali_vars.server.web_schema + '://' + kiali_vars.server.web_fqdn + ((':' + kiali_vars.server.web_port | string) if (kiali_vars.server.web_port | string | length != 0) else '') + (kiali_vars.server.web_root | default('')) }}
+{% endif %}
```
jmazzitelli, line 17: Is this feature only for non-OpenShift environments? If not (i.e. if this works on OpenShift, too), then the same change should go here: https://github.com/kiali/kiali-operator/blob/v1.31.0/roles/default/kiali-deploy/templates/openshift/service.yaml#L15

Don't forget the server helm chart, too --> https://github.com/kiali/helm-charts/blob/v1.31.0/kiali-server/templates/service.yaml#L14

israel-hdez: Here is the equivalent change for Helm: https://github.com/kiali/helm-charts/pull/52

For OpenShift, I think the proper way to do it is to query the created route. It should contain the needed data, so an annotation to the Service would be redundant.

jmazzitelli: OK. So we will only do this annotation thing for non-Openshift clusters, and the Kiali server will auto-detect using the route when on OpenShift.

jmazzitelli (approved):
lgtm

deploy/kiali/kiali_cr.yaml
```diff
@@ -796,6 +796,8 @@ spec:
 # of the URL (usually it's a fully-qualified domain name).
 # For example, "kiali.example.org".
 # When empty, Kiali will try to guess this value from HTTP headers.
+# You must populate this value if you want to enable cross-linking between Kiali
+# instances in a multi-cluster setup.
```
jmazzitelli, line 800: just a comment - is this true if on openshift?

If the server will use the route to autodiscover the external URL, is it true to say "You must populate" this value (and the web_schema, since it has the same comment)? Or should we prefix this comment with something like, "If Kiali is deployed in a non-OpenShift cluster, [you must populate this value....blah blah blah]"

israel-hdez: Oh, yes.... I think we should add the prefix.
Anyway, at the moment I'm testing only with K8s... I haven't spent the time to setup two OSs :-/.

israel-hdez: @jmazzitelli I added the prefix... and, now, I need your re-approval because GitHub drop your previous approval.

jmazzitelli (approved):
lgtm

PR #266, opened 2021-03-16
Make sure we pass through camelCase for node_selector.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3760

Note: cannot easily create a molecule test because it requires the test node to have the test labels on them, which currently the minikube and openshift nodes won't necessarily have. But this is such a simple change and mimics the same kind of change for the other things we do have tests for, that it shouldn't be a problem... I did do a manual test and confirmed this works. Just look at the kiali deployment yaml and see that the `nodeSelector` in the template is as expected (you don't have to worry if the pod doesn't come up - it won't if you used bogus test labels like "anotherCamelCase-Test: foo")

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -179,6 +179,15 @@
   - kiali_vars.deployment.hpa.spec is defined
   - kiali_vars.deployment.hpa.spec | length > 0
 
+- name: Replace snake_case with camelCase in deployment.node_selector
+  set_fact:
+    kiali_vars: |
+      {% set a=kiali_vars['deployment'].pop('node_selector') %}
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LOK.  I suppose no one is depending on us actually applying snake_case?

jmazzitelli:
This should be backward compat for virtually all use-cases. The only use case where this would break someone is if (1) they actually put snake_case labels on their node and (2) they used camelCase notation in the CR with the explicit expectation it be converted to snake_case by kiali operator. That would seem like an odd thing to do - just specify the snake_case in the kiali CR. k8s convention is to use camelCase for labels, so its rare to use snake_case. But in the times when someone does that, they will almost certainly use the exact value of that snake_case and not use camelCase in the kiali CR.

PR #267, opened 2021-03-16
Migrate CI from Travis to GitHub Actions

israel-hdez:
See a run here: https://github.com/israel-hdez/kiali-operator/pull/3/checks?check_run_id=2125323145

Unfortunately, Actions does not have ARM runners. So, we only can check for amd64.
But, at least, we restore CI.


jmazzitelli (approved):
This looks like it is doing what we want it to do.

However, I do notice that we lose the multi-platform testing. Is that right? Do we no longer have the ability to run on different platforms such as arm64 and ppc64?

jmazzitelli:
> Actions does not have ARM runners. So, we only can check for amd64

Never mind - I see you answered it already.

PR #269, opened 2021-03-22
revert back to CRD v1beta1

jmazzitelli:
When we stop supporting OS 4.4, we can revert this revert.
Note this only affects the OSSM metadata - we keep the rest (all upstream stuff) at v1.


jshaughn (approved):
I approve this revert.

PR #271, opened 2021-03-23
molecule tests to be able to skip operator install or to install via OLM or operator helm charts

jmazzitelli:
This is mainly to support CI testing that wants to test using OLM-installed operators. You will need to tell the molecule tests to NOT attempt to install the operator - it will be assumed to already exist.

This requires https://github.com/kiali/kiali/pull/3804

fixes: https://github.com/kiali/kiali/issues/3811


jshaughn (approved):
LOK

PR #117, opened 2020-08-13
Changes necessary to move to Ansible Operator SDK 1.x

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/2983

PR #272, opened 2021-03-24
use the new openshift/origin-ansible-operator base image

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3807

We should get this into this sprint's release because in this sprint [we also did a major upgrade of the operator SDK from 0.x to 1.3.0](https://github.com/kiali/kiali-operator/pull/117) - this just moves us onto the official okd build (which should be fully compatible with potentially additional features added in the future)


jmazzitelli:
testing building, pushing, and running. all looks good to me.

jshaughn (approved):
Wundabar

PR #273, opened 2021-03-25
Copy the console files to a memory-backed volume, which allows Kiali to be run read-only-root-dir

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3810

### DIFF
--- a/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml
@@ -49,6 +49,26 @@ spec:
 {% endif %}
     spec:
       serviceAccount: kiali-service-account
+      initContainer:
+      - name: copy-console
+        securityContext:
+          runAsUser: 1000
+          runAsGroup: 1000
+          runAsNonRoot: true
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          privileged: false
+          capabilities:
+            drop:
+            - ALL
+        image: {{ kiali_vars.deployment.image_name }}:{{ kiali_vars.deployment.image_version }}
+        command:
+        - sh
+        - -c
+        - cp -r /opt/kiali/console/* /tmp/console
+        volumeMounts:
+        - name: kiali-console
+          mountPath: /tmp/console
 {% if kiali_vars.deployment.priority_class_name != "" %}
       priorityClassName: "{{ kiali_vars.deployment.priority_class_name }}"
 {% endif %}
@@ -62,6 +82,16 @@ spec:
       - image: {{ kiali_vars.deployment.image_name }}:{{ kiali_vars.deployment.image_version }}
         imagePullPolicy: {{ kiali_vars.deployment.image_pull_policy }}
         name: kiali
+        securityContext:
+          runAsUser: 1000
+          runAsGroup: 1000
+          runAsNonRoot: true
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          privileged: false
+          capabilities:
+            drop:
+            - ALL
         command:
         - "/opt/kiali/kiali"
         - "-config"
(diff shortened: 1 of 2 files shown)


jmazzitelli:
Some quick testing shows this to be working. I will put on the "do not merge" label because I prefer this to not get into tomorrow's release. I want this change to soak in the dev community for a couple weeks before we put it in the release after that.

israel-hdez:
I'm not sure I like this.
The right solution should be to serve the `env.js` file from memory rather than doing this...

jmazzitelli:
> The right solution should be to serve the `env.js` file from memory rather than doing this...

Actually, the *main* point to this PR (and what the [issue](https://github.com/kiali/kiali/issues/3810) was complaining about) is we should start the pod with restricted permissions (e.g. read-only access to the root fs).

So whether we serve `env.js` and `index.html` from memory or from the file system, we still should do what this PR is doing with respect to setting the pod container's security context:

```
        securityContext:
          runAsUser: 1000
          runAsGroup: 1000
          runAsNonRoot: true
          allowPrivilegeEscalation: false
          readOnlyRootFilesystem: true
          privileged: false
          capabilities:
            drop:
            - ALL
```

The init-container doing the `cp` to the memory volume is needed because right now the kiali server (see kiali.go [here](https://github.com/kiali/kiali/blob/v1.31.0/kiali.go#L208-L228) and [here](https://github.com/kiali/kiali/blob/v1.31.0/kiali.go#L230-L259)) is writing out changes to the `env.js` and `index.html` files. If there is a separate solution to eliminate the need for the kiali server to actually write changes to those files, we can do that as a separate PR (or as a commit in this PR). I wouldn't mind getting rid of that init-container and the memory volume.

israel-hdez:
> Actually, the _main_ point to this PR (and what the [issue](https://github.com/kiali/kiali/issues/3810) was complaining about) is we should start the pod with restricted permissions (e.g. read-only access to the root fs).

Well... first of all: this has already been reported in the past in an issue quite closely related: https://github.com/kiali/kiali/issues/3297. Part of the solution was to create this FAQ entry: https://kiali.io/documentation/latest/faq/#psp-pod. If we are going to apply these changes, that FAQ entry becomes obsolete and should be removed.

Anyway...

The main question to do is: _why sysadmins may want read-only file systems?_ And one of my answers is that it's desirable to prevent changes to the container in case it gets compromised. That way, nobody should be capable of replacing Kiali's UI with something else. A read-only file system prevents serving malicious content and prevents running malicious code. That's why **I think** this is a workaround and we should not make it an official fix. This is opening the attack surface that sysadmins want to close and it's being done without they knowledge.

Serving the `env.js` and `index.html` files from memory is probably doable by adding entries in the [back-end router](https://github.com/kiali/kiali/blob/master/routing/router.go#L83).

FWIW, I have seen (and had to fix) basic web sites where somebody managed to place malicious PHP files. And once up, it is quite easy to run processes in there. It's scary... It's hard to know what had run and for how long. This is part of what an immutable file-system prevents.

jmazzitelli:
> EDIT: although I talked about Kiali-UI, I think that also applies to the binary of Kiali back-end. So, nobody should be capable of replacing the back-end's binary.

Going even further, I think it is more general - nobody should be capable of putting ANY executable binary in here ... we don't want someone getting a compromised pod, creating an ad-hoc script and running it. Having all file systems read-only prevents this. The way the current PRs are, I don't think that is prevented (someone could put an executable in `/tmp/console` - I'm not sure what damage can be done, but I don't like the fact that you can run something, even out of a memory-backed file system).

Certainly seems to me that the correct way to do this is remove the init-container and the memory-backed volume, and have the kiali server use the internal router to serve a modified form of env.js and index.html out of its own memory.

jmazzitelli:
closing this PR - we can always resurrect it if we want to come back to this implementation, but it seems like we want to try something else.

PR #275, opened 2021-03-30
Fix deprecated api link

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3026

Kiali API is not public and that link is wrong, there is no expectation to have it back in the kiali site, so, better to remove this annotation from the operator.

Please, @jmazzitelli let me know if it looks good for you.

@hhovsepy this is a very small fix but I haven't completely checked if there is nothing broken, specifically when browing the Kiali workload/service from Kiali itself.


jmazzitelli:
Interestingly, those annotations are in the kiali-server helm chart's service yaml but not the deployment yaml. So we need to rip this out in another PR: https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/service.yaml#L13-L14

lucasponce:
Thanks for the review @jmazzitelli, yes, I was unsure if we need them in the helm, I'll prepare a separated PR for that

hhovsepy (approved):
Regression tested.

PR #276, opened 2021-04-06
Fix comment in sample Kiali CR

israel-hdez:
Fixes kiali/kiali#3836

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -339,7 +339,7 @@ spec:
 #    resources: {}
 #
 # The name of a secret used by the Kiali.
-# This secret is optinally used when configuring the OpenID authentication strategy.
+# This secret is optionally used when configuring the OpenID authentication strategy.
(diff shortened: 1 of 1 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -338,7 +338,10 @@ spec:
 #    ---
 #    resources: {}
 #
-# The name of a secret used by the Kiali. Reserved for future use - not currently used.
+# The name of a secret used by the Kiali.
+# This secret is optinally used when configuring the OpenID authentication strategy.
```
jmazzitelli, line 342:
```suggestion
# This secret is optionally used when configuring the OpenID authentication strategy.
```

jmazzitelli (changes requested):
typo

deploy/kiali/kiali_cr.yaml
israel-hdez: Fixed.

jmazzitelli (approved):

PR #278, opened 2021-04-07
Abort on ad hoc kiali namespace deployment when it is not allowed

jmazzitelli:
OSSM-438

This fails the reconciliation if user tries to install kiali in a namespace different from where the CR is. This failure only happens when the Operator is configured to disallow such a thing (which is going to be the default)

- https://github.com/kiali/kiali/pull/3850
- https://github.com/kiali/helm-charts/pull/59

### DIFF
--- a/manifests/kiali-community/1.33.0/kiali.crd.yaml
+++ b/manifests/kiali-community/1.33.0/kiali.crd.yaml
@@ -1,0 +1,24 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: kialis.kiali.io
+  labels:
+    app: kiali-operator
+spec:
+  group: kiali.io
+  names:
+    kind: Kiali
+    listKind: KialiList
+    plural: kialis
+    singular: kiali
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    subresources:
+      status: {}
+    schema:
+      openAPIV3Schema:
+        type: object
+        x-kubernetes-preserve-unknown-fields: true
--- a/manifests/kiali-community/1.33.0/kiali.monitoringdashboards.crd.yaml
+++ b/manifests/kiali-community/1.33.0/kiali.monitoringdashboards.crd.yaml
@@ -1,0 +1,22 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: monitoringdashboards.monitoring.kiali.io
+  labels:
+    app: kiali
+spec:
+  group: monitoring.kiali.io
+  names:
+    kind: MonitoringDashboard
+    listKind: MonitoringDashboardList
+    plural: monitoringdashboards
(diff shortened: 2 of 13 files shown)


jmazzitelli:
molecule tests pass.

To test this: in short, if the operator was told not to allow for ad-hoc namespaces, you can see that prohibition enforced by creating a Kiali CR in namespace `foo` but set the value of the Kiali CR `spec.deployment.namespace` to `bar` (anything not `foo`).

To manual test this using our make infrastructure, I did this (note steps 1 and 2 just get you a cluster with Istio - you can skip it if you already have a cluster to test with that has Istio installed):

1. Start a `kind` cluster via `kind create cluster` (you can use minikube or OpenShift if you want, just make sure you set the make environment vars appropriately in the commands below).
2. Install Istio - `hack/istio/install-istio-via-istioctl.sh -c kubectl`
3. Get the three PRs associated with this fix: https://github.com/kiali/kiali/pull/3850, https://github.com/kiali/kiali-operator/pull/278, https://github.com/kiali/helm-charts/pull/59
4. Run: `make -e HELM_CHARTS_REPO_PULL=false -e CLUSTER_TYPE=kind cluster-push operator-create kiali-create`
5. After the make command finishes, confirm you have a Kiali CR in the `kiali-operator` namespace: `kubectl get kiali kiali -n kiali-operator`
6. Wait another 30 or 60 seconds for the operator to do its thing and then confirm Kiali is deployed: `kubectl get pods -n istio-system` should show you a Kiali pod. Here you confirm that the operator still works properly (because it was allowed to install kiali to an ad hoc namespace -  an ad hoc namespace is one that is different from where the Kiali CR was created).
7. Do it again, but this time turn off the allow-ad-hoc-namespace setting: `make -e OPERATOR_ALLOW_AD_HOC_KIALI_NAMESPACE=false -e HELM_CHARTS_REPO_PULL=false -e CLUSTER_TYPE=kind operator-create kiali-create`
8. Confirm the operator aborted the install - look at the Kiali CR status for error messages: `kubectl get kiali kiali -n kiali-operator -o jsonpath='{.status}'`  For example, typically the message in the first conditions item will be what you want to see - look for a message somewhere in the status field like this:
```
$ kubectl get kiali kiali -n kiali-operator -o jsonpath='{.status.conditions[0].message}'
The operator is forbidden from installing Kiali in a namespace [istio-system] that is different from the namespace where the CR was created [kiali-operator]
```

jmazzitelli:
Note that operators installed via the helm chart or operators installed via upstream/community OLM will allow ad-hoc namespace installs by default. The Red Hat operator will not. An admin can change this behavior by setting the env var on the operator deployment appropriately:

```
OPERATOR_NAMESPACE="kiali-operator" # if installed by OLM, could be something like "openshift-operators"

FLAG="true" # set to false if you want to prohibit users from installing Kiali in an ad hoc namespace

kubectl -n ${OPERATOR_NAMESPACE} set env deployment/kiali-operator ALLOW_AD_HOC_KIALI_NAMESPACE=${FLAG} 
```

mattmahoneyrh (approved):

PR #281, opened 2021-04-08
retain the permissions that are still needed for the old supported versions

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3854


jshaughn (approved):
LOK

PR #285, opened 2021-04-12
update kiali-cr docs to indicate auto-generation of certs no longer happens on non-OpenShift envs.

jmazzitelli:
This came about from this discussion: https://github.com/kiali/kiali/discussions/3867


jshaughn (approved):
LG

PR #295, opened 2021-04-21
Adds istio_sidecar_injector_config_map_name setting

bougyman:
This fulfills the docs for https://github.com/kiali/kiali/issues/3923

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -520,9 +520,10 @@ spec:
 # config_map_name: The name of the istio control plane config map. It defaults to `istio`.
 # envoy_admin_local_port: The port which kiali will open to fetch envoy config data information.
 # istio_identity_domain: The annotation used by Istio to identify domains.
+# istio_injector_name: The name of the istio-sidecar-injector config map. This defaults to `istio-sidecar-injector`
 # istio_injection_annotation: The annotation used by Istio to automatically inject a specific workload
 # istio_sidecar_annotation: The pod annotation used by Istio to identify the sidecar.
-# istiod_deployment_name: The name of the istiod deployment.
+# istiod_deployment_name: The name of the istiod deployment. Default: `istiod`
 # url_service_version: The Istio service used to determine the Istio version. If empty, assumes the URL for the well-known Istio version endpoint.
 #    ---
 #    istio:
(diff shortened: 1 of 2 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -520,9 +520,10 @@ spec:
 # config_map_name: The name of the istio control plane config map. It defaults to `istio`.
 # envoy_admin_local_port: The port which kiali will open to fetch envoy config data information.
 # istio_identity_domain: The annotation used by Istio to identify domains.
+# istio_injector_name: The name of the istio-sidecar-injector config map. This defaults to `istio-sidecar-injector`
```
jmazzitelli, line 523: We don't add verbosity to these help strings by specifying defaults because all the values you see here ARE the defaults (we are very careful in this file to set the commented values to be the actual defaults as explained [at the top of this file here](https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L16-L19)). So you will see down below you have `istio_injector_name: "istio-sidecar-injector"` - that value IS the real default. No need to call it out again here, just makes more noise in the docs. Also, shouldn't the name be more descriptive? If this is a literal config map name, shouldn't this be called "istio_injector_config_map_name" ?? I dunno, just weird to me if its a config map name it should be named that.

```suggestion
# istio_injector_name: The name of the istio-sidecar-injector config map.
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -520,9 +520,10 @@ spec:
...
 # istio_injection_annotation: The annotation used by Istio to automatically inject a specific workload
 # istio_sidecar_annotation: The pod annotation used by Istio to identify the sidecar.
-# istiod_deployment_name: The name of the istiod deployment.
+# istiod_deployment_name: The name of the istiod deployment. Default: `istiod`
```
jmazzitelli, line 526: Revert this line change for the same reason described in my earlier comment ("We don't add verbosity to these help strings by specifying defaults because all the values you see here ARE the defaults") - and as you see here, this file does document the correct default when it declares: ` istiod_deployment_name: "istiod"`
```suggestion
# istiod_deployment_name: The name of the istiod deployment.
```

jmazzitelli (changes requested):

jmazzitelli (approved):
this looks correct

PR #297, opened 2021-04-23
Snapshots the default ansible role and names it v1.33.

jmazzitelli:
This is only going in the v1.33 branch because we only need this to temporarily build from the v1.33 branch


jmazzitelli:
If you build and use this within the Maistra/OSSM environment, it requires the Maistra operator to set its generated Kiali CR's `spec.version` field to `v1.33`.

jshaughn (approved):
OK, this is good for the temporary 1.33 support.  And we'll have to later support 1.35 assuming that that remains the target Kiali 2.1 version.

PR #298, opened 2021-04-23
Add host stanza to Ingress

israel-hdez:
In Kubernetes, Ingress is setup as a catch-all ingress. This is
annoying. If the Kiali CR already contains the `web_fqdn` config, the
Ingress resource shouldn't be created as catch-all.


jmazzitelli (approved):
i have not tested this, but it looks correct to me.  But before we merge this, we should also get the same feature in the server helm-chart - so this would need to have the same kind of thing: https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/ingress.yaml

israel-hdez:
Helm PR: https://github.com/kiali/helm-charts/pull/69

PR #293, opened 2021-04-18
make the operator go faster

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3892
fixes: https://github.com/kiali/kiali/issues/3928
fixes: https://github.com/kiali/kiali/issues/3936

Instead of individually (separately) creating the role and rolebinding on each individual namespace, do a bulk create of every role and every role binding in every namespace in one single, big k8s task invocation. This should speed things up considerable.

There are a couple concerns I have. Is there a limit to either the full size of the yaml string or the number of resources that the k8s task or the kubernetes master API itself will accept? Can Ansible build an arbitrarily long (theoretically unlimited) yaml string? And will the ansible k8s task or the kubernetes master API accept an arbitrary (theorteically unlimited) number of resources to create? Hopefully the answer is yes for both, or if not yet, hopefully the limit is so large so as to not concern us. But we should test this out - for example, if I put 1,000 namespaces in the `spec.deployment.accessible_namespaces` setting, will that be OK? How about 2,000? What's the theoretical maximum number of namespaces that we should support?

PR #299, opened 2021-04-27
[v1.33] make the operator go faster (#293)

jmazzitelli:
backports #293 so any builds off this v1.33 branch will get these improvements

(cherry picked from commit 1de94a8720ffa8e815c8cc683eb45f65a98f2b3d)


jshaughn (approved):
This will only be temporary for ossm 2.1, right, as we will plan to update to 1.35.  If that is the plan then this LOK.

jmazzitelli:
> This will only be temporary for ossm 2.1, right, as we will plan to update to 1.35. If that is the plan then this LOK.

correct. These changes are already in master, so we'll get these in any future branch.

PR #300, opened 2021-04-27
checks for namespace existence immediately prior to attempting to add/remove the label

jmazzitelli:
Temporary fix for https://github.com/kiali/kiali/issues/3949 - when that issue is really fixed, we should revert this commit.

This temp fix checks for namespace existence immediately prior to attempting to add/remove the label

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -685,11 +685,13 @@
     # everything to the left of the = is the name of the label we want to remove
     the_namespace_label_name: "{{ current_label_selector | regex_replace('^(.*)=.*$', '\\1') }}"
   # if a namespace happened to have been deleted, we do not want to (nor can we) resurrect it, hence we check for its existence
+  # https://github.com/kiali/kiali/issues/3949 : the if-stmt to query the namespace for existence is temporary - remove when that issue is truly fixed
   k8s:
     state: present
     definition: |
       {% for namespace in no_longer_accessible_namespaces %}
       {% if namespace in all_namespaces %}
+      {% if (query('k8s', kind='Namespace', resource_name=namespace) | length) > 0 %}
       ---
       apiVersion: v1
       kind: Namespace
@@ -698,6 +700,7 @@
         labels:
           {{ the_namespace_label_name }}: null
       ...
+      {% endif %}
       {% endif %}
       {% endfor %}
   when:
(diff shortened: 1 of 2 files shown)


jmazzitelli:
@FilipB now that the operator performance enhancements are in the master branch, but with the "namespace deletion problem", could I ask you to review/test this draft PR? This should "fix" that namespace deleting problem. I am curious, though, how much time it adds to your 200 and 2,000 namespace test. My testing shows it is OK - I think due to the k8s collection caching that it does.

(note: there is still a race condition, but it is very, very brief - there is still a small chance that you delete a namespace between the time this new code checks for the existence of the namespace and when the label is added/removed - and if that happens that error will occur. But that is very brief time slice (because the add/remove happens immediately after the namespace query check). But this is why I consider this a temp fix - if/when we get that k8s collections enhancement request, we'll use that instead. But that enhancement request has no ETA - who knows if/when we'll get it. So this temp fix is a way to workaround it (but I do not want to merge this if it brings us back to 30 minute reconciliation times).

jmazzitelli:
I put the DNM label on this PR... only merge this PR after we get test results from @FilipB

FilipB:
It slows down the removal significantly. For 2000 namespaces it was 246s. With this change it is 738:
```
TASK [Log reconciliation processing time] **************************************
Wednesday 28 April 2021  08:22:09 +0000 (0:00:00.034)       0:12:18.971 ******* 
ok: [localhost] => {
    "msg": "Processing time: [738] seconds"
}

TASK [debug] *******************************************************************
Wednesday 28 April 2021  08:22:09 +0000 (0:00:00.035)       0:12:19.006 ******* 
ok: [localhost] => {
    "msg": "KIALI RECONCILIATION IS DONE."
}
localhost                  : ok=130  changed=18   unreachable=0    failed=0    skipped=70   rescued=0    ignored=0   


PLAY RECAP *********************************************************************
localhost                  : ok=130  changed=18   unreachable=0    failed=0    skipped=70   rescued=0    ignored=0   

Wednesday 28 April 2021  08:22:09 +0000 (0:00:00.028)       0:12:19.035 ******* 
=============================================================================== 
default/kiali-deploy : Remove Kiali label from namespaces that Kiali no longer has access to - 544.77s
default/kiali-deploy : Delete Kiali roles from previously accessible namespaces -- 95.08s
default/kiali-deploy : Delete all ConsoleLinks for namespaces that are no longer accessible -- 48.11s
default/kiali-deploy : Create the Monitoring Dashboards ----------------- 3.23s
```

Also it does not improve the situation for cases when one is using CLI for interaction with the cluster and working with more namespaces.
e.g. this script (deleting 20 namespaces) will hit the issue every time:
```
cat << EOF > kiali.yaml
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  name: kiali
spec:
  deployment:
    accessible_namespaces:
      - default
EOF

oc patch kiali kiali -n kiali-operator --type merge --patch "$(cat kiali.yaml)"

for i in {1..20}
do
  oc delete project test$i
done
```

jmazzitelli:
OK. Then we are just going to have to wait for the Ansible k8s collection team to fix this on their end - we need to wait for https://github.com/ansible-collections/kubernetes.core/issues/84

PR #306, opened 2021-04-30
go back to defaulting to the same kiali image version as that of the operator

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3968


jmazzitelli:
> LOK although there needs to also be a kiali.io PR to make sure the install doco is in good shape.

I'll submit that PR soon. But I'll merge this now.

PR #309, opened 2021-05-04
[v1.24] no longer support 1.0.x kiali servers; revert CRDs back to v1beta1

jmazzitelli:
This changes the OLM metadata for v1.24 builds.


jshaughn (approved):
wunderbar

PR #310, opened 2021-05-06
fix molecule test - make the resource limits more realistic

jmazzitelli:
I saw some errors that occurred until I made this change. The test just passed when I incorporated this change.
This just makes sure the pod can use a realistic amount of memory.

fixes: https://github.com/kiali/kiali/issues/3984


jmazzitelli:
Ran a second test run - passed as well.

jshaughn (approved):
LOK

PR #313, opened 2021-05-13
Be able to customize the names of the kiali resources.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3920

There is now a new setting in the Kiali CR `spec.deployment.instance_name` whose default is `kiali`. If you set it, all the resources (except for the signing key secret) will have that as its name prefix. For example, I set this to `mazzwashere` and the resulting objects are:

```
$ kubectl get deployment,pod,service,sa,cm,ingress,clusterrole,clusterrolebinding,role,rolebinding,secret -n istio-system -l app.kubernetes.io/name=kiali -o name
deployment.apps/mazzwashere
pod/mazzwashere-574b7c7fb8-grzpj
service/mazzwashere
serviceaccount/mazzwashere-service-account
configmap/mazzwashere
ingress.networking.k8s.io/mazzwashere
clusterrole.rbac.authorization.k8s.io/mazzwashere
clusterrolebinding.rbac.authorization.k8s.io/mazzwashere
role.rbac.authorization.k8s.io/mazzwashere-controlplane
rolebinding.rbac.authorization.k8s.io/mazzwashere-controlplane
secret/kiali-signing-key
```

A few caveats. Unlike the helm charts behavior, all the app labels remain fixed with `kiali`. I'm not sure which way it should go, but it seems to me the app labels (e.g. `app` and `app.kubernetes.io/name`) should be fixed with `kiali` since this indicates the resources belong to the Kiali application (regardless of what the resource names are). The helm charts behave differently and, in fact, I would almost consider it a bug that the helm charts do not do this. I updated the helm charts with this PR https://github.com/kiali/helm-charts/pull/75 so the labels are consistent with this new operator behavior (with that PR, the helm charts will also support this new `deployment.instance_name` value, again for more consistency with the operator).

A second thing is the auto-generated signing key secret name is fixed to `kiali-signing-key`. If the user does not provide a signing key (used for encrypting https traffic to the server), the operator has to create one. And when it does so, it puts that signing key in a secret. In order to create a secret, the operator is given CRUD permissions in its cluster role for secrets named `kiali-signing-key`. We cannot have this secret name dynamic (i.e. it cannot be determined at run time with this new instance_name setting) because "resourceNames" in the k8s role definition do not support regex (so we can't give permission to resourceNames of `*-signing-key` in the role - [note that this enhancement request was rejected by the k8s community](https://github.com/kubernetes/kubernetes/issues/56582)). We do not want to grant CRUD permissions to all secrets in the operator role (for security purposes) so we are going to limit the name of this signing key secret to always be `kiali-signing-key`. If the user does not like this, they can very easily create their own signing key and put it in their own secret and tell the operator about that secret via the [`spec.login_token.signing_key` setting](https://github.com/kiali/kiali-operator/blob/v1.34.0/deploy/kiali/kiali_cr.yaml#L776-L784).

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -41,6 +41,9 @@ spec:
 
 ##########
 # Tag used to identify a particular instance/installation of the Kiali server.
+# This is merely a human-readable string that will be used within Kiali to
+# help a user identify the Kiali being used (e.g. in the Kiali UI title bar).
+# See deployment.instance_name for the setting used to customize Kiali resource names that are created.
 #  ---
 #  installation_tag: ""
 
@@ -100,9 +103,11 @@ spec:
 # (e.g. "myLabel=myValue") which is used when fetching the list of available namespaces.
 # This does not affect explicit namespace access.
 # Note that if you do not set this but deployment.accessible_namespaces does not have the
-# special "all namespaces" value of "**" then this label_selector will be set
-# to a default value of "kiali.io/member-of=<deployment.namespace>" where
-# <deployment.namespace> is the namespace where Kiali is to be installed.
+# special "all namespaces" entry of "**" then this label_selector will be set
+# to a default value of "kiali.io/[<deployment.instance_name>.]member-of=<deployment.namespace>"
+# where [<deployment.instance_name>.] is the instance name assigned to the Kiali installation
+# if it is not the default 'kiali' (otherwise, this is omitted) and <deployment.namespace>
+# is the namespace where Kiali is to be installed.
 # If deployment.accessible_namespaces does not have the special value of "**"
 # then the Kiali operator will add a new label to all accessible namespaces - that new
 # label will be this label_selector.
@@ -268,8 +273,8 @@ spec:
 #    ---
 #    ingress_enabled: true
 #
-# The instance name of this Kiali installation. This will be the prefix prepended
-# to the names of all Kiali resources created by the operator and will be used
+# The instance name of this Kiali installation. This instance name will be the prefix
+# prepended to the names of all Kiali resources created by the operator and will be used
 # to label those resources as belonging to this Kiali installation instance.
 # You cannot change this instance name after a Kiali CR is created. If you attempt
 # to change it, the operator will abort with an error. If you want to change it,
@@ -280,6 +285,8 @@ spec:
 # Kiali instances in the same deployment namespace. If you want a different
 # signing key secret, you are free to create your own and tell the operator about it via
 # the login_token.signing_key setting. See the docs on that setting for more details.
+# Note also that if you are setting this value, you may also want to change
+# the installation_tag setting, but this is not required.
 #    ---
 #    instance_name: "kiali"
 #
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -333,6 +333,18 @@ spec:
 #    ---
 #    replicas: 1
 #
+# Defines the prefix prepended to the names of all Kiali resources created by the operator.
+# You cannot change this prefix after a Kiali CR is created. If you attempt to change it,
+# the operator will abort with an error. If you want to change it, you must first
+# delete the original Kiali CR and create a new one.
+# Note that this does not affect the name of the auto-generated signing key secret. If
+# you do not supply a signing key, the operator will create one for you in a secret,
+# but that secret will always be named "kiali-signing-key". If you want a different
+# signing key secret, you are free to create your own and tell the operator about it via
+# the login_token.signing_key setting. See the docs on that setting for more details.
+#    ---
+#    resource_names_prefix: "kiali"
+#
 # Defines compute resources that are to be given to the Kiali pod's container.
 # The value is a dict as defined by Kubernetes. See the Kubernetes documentation
 # https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -268,6 +268,21 @@ spec:
 #    ---
 #    ingress_enabled: true
 #
+# The instance name of this Kiali installation. This will be the prefix prepended
+# to the names of all Kiali resources created by the operator and will be used
+# to label those resources as belonging to this Kiali installation instance.
+# You cannot change this instance name after a Kiali CR is created. If you attempt
+# to change it, the operator will abort with an error. If you want to change it,
+# you must first delete the original Kiali CR and create a new one.
+# Note that this does not affect the name of the auto-generated signing key secret. If
+# you do not supply a signing key, the operator will create one for you in a secret,
+# but that secret will always be named "kiali-signing-key" and shared across all
+# Kiali instances in the same deployment namespace. If you want a different
+# signing key secret, you are free to create your own and tell the operator about it via
+# the login_token.signing_key setting. See the docs on that setting for more details.
+#    ---
+#    instance_name: "kiali"
+#
 # Determines the logger configuration.
 # log_format supports text and json.
 # log_level supports trace, debug, info, warn, error, fatal.
@@ -332,18 +347,6 @@ spec:
 # The replica count for the Kiail deployment.
 #    ---
 #    replicas: 1
-#
-# Defines the prefix prepended to the names of all Kiali resources created by the operator.
-# You cannot change this prefix after a Kiali CR is created. If you attempt to change it,
-# the operator will abort with an error. If you want to change it, you must first
-# delete the original Kiali CR and create a new one.
-# Note that this does not affect the name of the auto-generated signing key secret. If
-# you do not supply a signing key, the operator will create one for you in a secret,
-# but that secret will always be named "kiali-signing-key". If you want a different
-# signing key secret, you are free to create your own and tell the operator about it via
-# the login_token.signing_key setting. See the docs on that setting for more details.
-#    ---
-#    resource_names_prefix: "kiali"
 #
 # Defines compute resources that are to be given to the Kiali pod's container.
 # The value is a dict as defined by Kubernetes. See the Kubernetes documentation
--- a/roles/default/kiali-deploy/templates/kubernetes/service.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/service.yaml
@@ -3,12 +3,7 @@ kind: Service
 metadata:
   name: {{ kiali_vars.deployment.resource_names_prefix }}
   namespace: {{ kiali_vars.deployment.namespace }}
-  labels:
-    app: kiali
-    version: {{ kiali_vars.deployment.version_label }}
-    app.kubernetes.io/name: kiali
-    app.kubernetes.io/version: {{ kiali_vars.deployment.version_label }}
-    app.kubernetes.io/part-of: kiali
+  labels: {{ kiali_resource_metadata_labels }}
   annotations:
 {% if kiali_vars.server.web_fqdn|length != 0 and kiali_vars.server.web_schema|length != 0 %}
     kiali.io/external-url: {{ kiali_vars.server.web_schema + '://' + kiali_vars.server.web_fqdn + ((':' + kiali_vars.server.web_port | string) if (kiali_vars.server.web_port | string | length != 0) else '') + (kiali_vars.server.web_root | default('')) }}
@@ -31,5 +26,5 @@ spec:
 {% endif %}
   selector:
     app.kubernetes.io/name: kiali
-    app.kubernetes.io/version: {{ kiali_vars.deployment.version_label }}
+    app.kubernetes.io/instance: {{ kiali_vars.deployment.resource_names_prefix }}
   {% if kiali_vars.deployment.additional_service_yaml is defined %}{{ kiali_vars.deployment.additional_service_yaml | to_nice_yaml(indent=0) | trim | indent(2) }}{% endif %}
--- a/roles/default/kiali-deploy/templates/kubernetes/service.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/service.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ kiali_vars.deployment.resource_names_prefix }}
+  name: {{ kiali_vars.deployment.instance_name }}
   namespace: {{ kiali_vars.deployment.namespace }}
   labels: {{ kiali_resource_metadata_labels }}
   annotations:
@@ -26,5 +26,5 @@ spec:
 {% endif %}
   selector:
     app.kubernetes.io/name: kiali
-    app.kubernetes.io/instance: {{ kiali_vars.deployment.resource_names_prefix }}
+    app.kubernetes.io/instance: {{ kiali_vars.deployment.instance_name }}
   {% if kiali_vars.deployment.additional_service_yaml is defined %}{{ kiali_vars.deployment.additional_service_yaml | to_nice_yaml(indent=0) | trim | indent(2) }}{% endif %}
--- a/roles/default/kiali-deploy/templates/kubernetes/service.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/service.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: kiali
+  name: {{ kiali_vars.deployment.resource_names_prefix }}
   namespace: {{ kiali_vars.deployment.namespace }}
   labels:
     app: kiali
@@ -30,6 +30,6 @@ spec:
     port: {{ kiali_vars.server.metrics_port }}
 {% endif %}
   selector:
-    app: kiali
-    version: {{ kiali_vars.deployment.version_label }}
+    app.kubernetes.io/name: kiali
+    app.kubernetes.io/version: {{ kiali_vars.deployment.version_label }}
   {% if kiali_vars.deployment.additional_service_yaml is defined %}{{ kiali_vars.deployment.additional_service_yaml | to_nice_yaml(indent=0) | trim | indent(2) }}{% endif %}
--- a/molecule/asserts/roles-test/ro_clusterrole_asserts.yml
+++ b/molecule/asserts/roles-test/ro_clusterrole_asserts.yml
@@ -1,8 +1,10 @@
 - name: Get cluster roles
+  vars:
+    instance_name: "{{ kiali.instance_name | default('kiali') }}"
   k8s_info:
    api_version: rbac.authorization.k8s.io/v1
    kind: ClusterRole
-   name: kiali-viewer
+   name: "{{ instance_name }}-viewer"
   register: clusterroles
 
 - name: Assert that cluster roles exist
@@ -11,14 +13,18 @@
     fail_msg: "The kiali-viewer cluster role does not exist"
 
 - name: Get cluster role binding
+  vars:
+    instance_name: "{{ kiali.instance_name | default('kiali') }}"
   k8s_info:
    api_version: rbac.authorization.k8s.io/v1
    kind: ClusterRoleBinding
-   name: kiali
+   name: "{{ instance_name }}"
   register: clusterrolebindings
 
 - name: Assert the cluster role binding provides the read-only viewer role
+  vars:
+    instance_name: "{{ kiali.instance_name | default('kiali') }}"
   assert:
     that:
-    - "{{ clusterrolebindings.resources[0] | default({}) | json_query('roleRef.name') == 'kiali-viewer' }}"
-    fail_msg: "The kiali cluster role binding did not have the read-only roleref 'kiali-viewer'"
+    - "{{ clusterrolebindings.resources[0] | default({}) | json_query('roleRef.name') == instance_name + '-viewer' }}"
+    fail_msg: "The kiali cluster role binding did not have the read-only roleref {{ instance_name }}-viewer"
--- a/molecule/asserts/roles-test/ro_role_asserts.yml
+++ b/molecule/asserts/roles-test/ro_role_asserts.yml
@@ -1,4 +1,6 @@
 - name: Get roles
+  vars:
+    instance_name: "{{ kiali.instance_name | default('kiali') }}"
   k8s_info:
    api_version: rbac.authorization.k8s.io/v1
    kind: Role
@@ -7,7 +9,7 @@
   register: roles
   with_nested:
   - "{{ kiali.accessible_namespaces }}"
-  - kiali-viewer
+  - "{{ instance_name}}-viewer"
 
 - name: Assert that namespaces have the correct roles
   assert:
@@ -16,6 +18,8 @@
   - "{{ roles.results }}"
 
 - name: Get role binding
+  vars:
+    instance_name: "{{ kiali.instance_name | default('kiali') }}"
   k8s_info:
    api_version: rbac.authorization.k8s.io/v1
    kind: RoleBinding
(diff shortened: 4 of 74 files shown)


jmazzitelli:
How to test:

1. Create a  Kiali CR and set `spec.deployment.resource_names_prefix` to something other than `kiali` (say, `mazzwashere`).
2. Check that all the resources (except for the kiali-signing-key) now have resource names with that prefix and not "kiali" - run this command to see all the Kiali resources created in the istio-system namespace - `kubectl get deployment,pod,service,sa,cm,ingress,clusterrole,clusterrolebinding,role,rolebinding,secret -n istio-system -l app.kubernetes.io/name=kiali -o name`
3. Go to the Kiali UI and make sure everything works. You can port-forward to the Kiali server using app label selector like this: `kubectl -n istio-system port-forward $(kubectl -n istio-system get pod -l app.kubernetes.io/name=kiali -o jsonpath='{.items[0].metadata.name}') 20001:20001`
4. Edit the Kiali CR and try to change that resource_names_prefix setting. This should FAIL! This is not allowed. You cannot change the prefix names after Kiali has been installed. You must delete the Kiali CR and then create a new Kiali CR with the new prefix. The error you should see will be in the "status" field in the original Kiali CR - so look at the Kiali CR yaml, inside the status field and you'll see a message like `The deployment.resource_names_prefix cannot be changed to a different value. It was [mazzwashere] but is now [foo]. In order to install Kiali with a different resource names prefix, please uninstall Kiali first.`
5. Delete the Kiali CR and see the original resources that were created (with your original prefix name e.g. `mazzwashere`) are now deleted.
6. Finally, test the original backward compatible behavior still works. Create a Kiali CR without setting this resource_names_prefix setting and check that Kiali installs as it always has, with the original names starting with `kiali`. Make sure Kiali works. Delete the CR and see everything is deleted.

jmazzitelli:
I'm going to add a `app.kubernetes.io/instance` label to all our resources with the name of this resource_names_prefix setting. This may be useful in the future but immediately allows one to query resources for this specific instance of kiali over all others. See https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

jmazzitelli:
Setting instance_name to `mazzwashere` results in these resources.

```
$ kubectl get all,deployment,pod,service,sa,cm,secret,ingress,clusterrole,clusterrolebinding,role,rolebinding,consolelinks,oauthclient -l app.kubernetes.io/name=kiali --all-namespaces
NAMESPACE      NAME                               READY   STATUS    RESTARTS   AGE
istio-system   pod/mazzwashere-69848888c5-bxs7g   1/1     Running   0          10m

NAMESPACE      NAME                  TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)              AGE
istio-system   service/mazzwashere   ClusterIP   172.30.12.195   <none>        20001/TCP,9090/TCP   10m

NAMESPACE      NAME                          READY   UP-TO-DATE   AVAILABLE   AGE
istio-system   deployment.apps/mazzwashere   1/1     1            1           11m

NAMESPACE      NAME                                     DESIRED   CURRENT   READY   AGE
istio-system   replicaset.apps/mazzwashere-69848888c5   1         1         1       11m

NAMESPACE      NAME                                   HOST/PORT                                  PATH   SERVICES      PORT    TERMINATION          WILDCARD
istio-system   route.route.openshift.io/mazzwashere   mazzwashere-istio-system.apps.ocp4.local          mazzwashere   <all>   reencrypt/Redirect   None

NAMESPACE      NAME                                         SECRETS   AGE
istio-system   serviceaccount/mazzwashere-service-account   2         11m

NAMESPACE      NAME                             DATA   AGE
istio-system   configmap/mazzwashere            1      11m
istio-system   configmap/mazzwashere-cabundle   1      11m

NAMESPACE      NAME                       TYPE     DATA   AGE
istio-system   secret/kiali-signing-key   Opaque   1      11m

NAMESPACE      NAME                                                      CREATED AT
istio-system   role.rbac.authorization.k8s.io/mazzwashere                2021-05-14T16:37:00Z
istio-system   role.rbac.authorization.k8s.io/mazzwashere-controlplane   2021-05-14T16:37:00Z
mazz           role.rbac.authorization.k8s.io/mazzwashere                2021-05-14T16:37:08Z

NAMESPACE      NAME                                                             ROLE                            AGE
istio-system   rolebinding.rbac.authorization.k8s.io/mazzwashere                Role/mazzwashere                11m
istio-system   rolebinding.rbac.authorization.k8s.io/mazzwashere-controlplane   Role/mazzwashere-controlplane   11m
mazz           rolebinding.rbac.authorization.k8s.io/mazzwashere                Role/mazzwashere                10m

NAMESPACE   NAME                                                                  TEXT                  URL                                                                                                 MENU   AGE
            consolelink.console.openshift.io/mazzwashere-namespace-istio-system   Kiali [mazzwashere]   https://mazzwashere-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=istio-system          10m
            consolelink.console.openshift.io/mazzwashere-namespace-mazz           Kiali [mazzwashere]   https://mazzwashere-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=mazz                  10m

NAMESPACE   NAME                                                      SECRET   WWW-CHALLENGE   TOKEN-MAX-AGE   REDIRECT URIS
            oauthclient.oauth.openshift.io/mazzwashere-istio-system            false           default         https://mazzwashere-istio-system.apps.ocp4.local
```

I will see if we can actually install another instance of kiali in this same namespace and have it work - that would be a bonus and give us some capabilities people have been asking for in the past.

jmazzitelli:
currently have this locally (I'll push after I do more testing)... I installed 3 instances of kiali into the same control plane (one with the default instance name `kiali` with the other two named `mazzwashere` and `mazzwashereagain`. All can presumably have different config but can also share namespaces within accessible_namespaces list).
```
$ kubectl get all,deployment,pod,service,sa,cm,secret,ingress,clusterrole,clusterrolebinding,role,rolebinding,consolelinks,oauthclient -l app.kubernetes.io/name=kiali --all-namespaces
NAMESPACE      NAME                                    READY   STATUS    RESTARTS   AGE
istio-system   pod/kiali-65bd6c8d85-mv8q6              1/1     Running   0          4m35s
istio-system   pod/mazzwashere-b8ffd9c67-mslsp         1/1     Running   0          7m17s
istio-system   pod/mazzwashereagain-55cdbf78b7-btk8d   1/1     Running   0          21m

NAMESPACE      NAME                       TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)              AGE
istio-system   service/kiali              ClusterIP   172.30.18.232   <none>        20001/TCP,9090/TCP   4m34s
istio-system   service/mazzwashere        ClusterIP   172.30.199.51   <none>        20001/TCP,9090/TCP   7m16s
istio-system   service/mazzwashereagain   ClusterIP   172.30.40.216   <none>        20001/TCP,9090/TCP   21m

NAMESPACE      NAME                               READY   UP-TO-DATE   AVAILABLE   AGE
istio-system   deployment.apps/kiali              1/1     1            1           4m35s
istio-system   deployment.apps/mazzwashere        1/1     1            1           7m17s
istio-system   deployment.apps/mazzwashereagain   1/1     1            1           21m

NAMESPACE      NAME                                          DESIRED   CURRENT   READY   AGE
istio-system   replicaset.apps/kiali-65bd6c8d85              1         1         1       4m35s
istio-system   replicaset.apps/mazzwashere-b8ffd9c67         1         1         1       7m17s
istio-system   replicaset.apps/mazzwashereagain-55cdbf78b7   1         1         1       21m

NAMESPACE      NAME                                        HOST/PORT                                       PATH   SERVICES           PORT    TERMINATION          WILDCARD
istio-system   route.route.openshift.io/kiali              kiali-istio-system.apps.ocp4.local                     kiali              <all>   reencrypt/Redirect   None
istio-system   route.route.openshift.io/mazzwashere        mazzwashere-istio-system.apps.ocp4.local               mazzwashere        <all>   reencrypt/Redirect   None
istio-system   route.route.openshift.io/mazzwashereagain   mazzwashereagain-istio-system.apps.ocp4.local          mazzwashereagain   <all>   reencrypt/Redirect   None

NAMESPACE      NAME                                              SECRETS   AGE
istio-system   serviceaccount/kiali-service-account              2         4m45s
istio-system   serviceaccount/mazzwashere-service-account        2         7m25s
istio-system   serviceaccount/mazzwashereagain-service-account   2         21m

NAMESPACE      NAME                                  DATA   AGE
istio-system   configmap/kiali                       1      4m43s
istio-system   configmap/kiali-cabundle              1      4m42s
istio-system   configmap/mazzwashere                 1      7m24s
istio-system   configmap/mazzwashere-cabundle        1      7m23s
istio-system   configmap/mazzwashereagain            1      21m
istio-system   configmap/mazzwashereagain-cabundle   1      21m

NAMESPACE      NAME                       TYPE     DATA   AGE
istio-system   secret/kiali-signing-key   Opaque   1      21m

NAMESPACE      NAME                                                           CREATED AT
istio-system   role.rbac.authorization.k8s.io/kiali                           2021-05-14T17:39:14Z
istio-system   role.rbac.authorization.k8s.io/kiali-controlplane              2021-05-14T17:39:15Z
istio-system   role.rbac.authorization.k8s.io/mazzwashere                     2021-05-14T17:36:33Z
istio-system   role.rbac.authorization.k8s.io/mazzwashere-controlplane        2021-05-14T17:36:34Z
istio-system   role.rbac.authorization.k8s.io/mazzwashereagain                2021-05-14T17:22:30Z
istio-system   role.rbac.authorization.k8s.io/mazzwashereagain-controlplane   2021-05-14T17:22:31Z
mazz           role.rbac.authorization.k8s.io/kiali                           2021-05-14T17:39:24Z
mazz           role.rbac.authorization.k8s.io/mazzwashere                     2021-05-14T17:36:42Z
mazz           role.rbac.authorization.k8s.io/mazzwashereagain                2021-05-14T17:22:39Z

NAMESPACE      NAME                                                                  ROLE                                 AGE
istio-system   rolebinding.rbac.authorization.k8s.io/kiali                           Role/kiali                           4m39s
istio-system   rolebinding.rbac.authorization.k8s.io/kiali-controlplane              Role/kiali-controlplane              4m38s
istio-system   rolebinding.rbac.authorization.k8s.io/mazzwashere                     Role/mazzwashere                     7m20s
istio-system   rolebinding.rbac.authorization.k8s.io/mazzwashere-controlplane        Role/mazzwashere-controlplane        7m19s
istio-system   rolebinding.rbac.authorization.k8s.io/mazzwashereagain                Role/mazzwashereagain                21m
istio-system   rolebinding.rbac.authorization.k8s.io/mazzwashereagain-controlplane   Role/mazzwashereagain-controlplane   21m
mazz           rolebinding.rbac.authorization.k8s.io/kiali                           Role/kiali                           4m30s
mazz           rolebinding.rbac.authorization.k8s.io/mazzwashere                     Role/mazzwashere                     7m12s
mazz           rolebinding.rbac.authorization.k8s.io/mazzwashereagain                Role/mazzwashereagain                21m

NAMESPACE   NAME                                                                       TEXT                       URL                                                                                                      MENU   AGE
            consolelink.console.openshift.io/kiali-namespace-istio-system              Kiali                      https://kiali-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=istio-system                     4m27s
            consolelink.console.openshift.io/kiali-namespace-mazz                      Kiali                      https://kiali-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=mazz                             4m27s
            consolelink.console.openshift.io/mazzwashere-namespace-istio-system        Kiali [mazzwashere]        https://mazzwashere-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=istio-system               7m9s
            consolelink.console.openshift.io/mazzwashere-namespace-mazz                Kiali [mazzwashere]        https://mazzwashere-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=mazz                       7m9s
            consolelink.console.openshift.io/mazzwashereagain-namespace-istio-system   Kiali [mazzwashereagain]   https://mazzwashereagain-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=istio-system          21m
            consolelink.console.openshift.io/mazzwashereagain-namespace-mazz           Kiali [mazzwashereagain]   https://mazzwashereagain-istio-system.apps.ocp4.local/console/graph/namespaces?namespaces=mazz                  21m

NAMESPACE   NAME                                                           SECRET   WWW-CHALLENGE   TOKEN-MAX-AGE   REDIRECT URIS
            oauthclient.oauth.openshift.io/kiali-istio-system                       false           default         https://kiali-istio-system.apps.ocp4.local
            oauthclient.oauth.openshift.io/mazzwashere-istio-system                 false           default         https://mazzwashere-istio-system.apps.ocp4.local
            oauthclient.oauth.openshift.io/mazzwashereagain-istio-system            false           default         https://mazzwashereagain-istio-system.apps.ocp4.local
```
And this is important for openshift auth - the client_id_prefix within the configmap is set correctly (which ensures each kiali instance is using its own oauthclient for authentication):

```
$ oc get cm kiali -n istio-system -o yaml  | grep client_id_prefix
        client_id_prefix: kiali

$ oc get cm mazzwashere -n istio-system -o yaml  | grep client_id_prefix
        client_id_prefix: mazzwashere

$ oc get cm mazzwashereagain -n istio-system -o yaml  | grep client_id_prefix
        client_id_prefix: mazzwashereagain
```

The namespace labels are now unique per Kiali instance. If the instance name is the default "kiali" the labels are as before (`kiali.io/member-of`) but specifically named instances now have labels such as `kiali.io/<instance name>.member-of`:

```
$ oc get ns mazz -o jsonpath={.metadata.labels} | jq
{
  "kiali.io/mazzwashere.member-of": "istio-system",
  "kiali.io/mazzwashereagain.member-of": "istio-system",
  "kiali.io/member-of": "istio-system"
}
```

And the OpenShift console-links objects have `kiali.io/home` labels whose values tell you which instance of kiali in which namespace own that link:

```
$ oc get consolelinks.console.openshift.io kiali-namespace-mazz -o jsonpath={.metadata.labels} | jq
{
  "app": "kiali",
  "app.kubernetes.io/instance": "kiali",
  "app.kubernetes.io/name": "kiali",
  "app.kubernetes.io/part-of": "kiali",
  "app.kubernetes.io/version": "v1.34",
  "kiali.io/home": "istio-system",
  "version": "v1.34"
}

$ oc get consolelinks.console.openshift.io mazzwashere-namespace-mazz -o jsonpath={.metadata.labels} | jq
{
  "app": "kiali",
  "app.kubernetes.io/instance": "mazzwashere",
  "app.kubernetes.io/name": "kiali",
  "app.kubernetes.io/part-of": "kiali",
  "app.kubernetes.io/version": "v1.34",
  "kiali.io/home": "mazzwashere.istio-system",
  "version": "v1.34"
}

$ oc get consolelinks.console.openshift.io mazzwashereagain-namespace-mazz -o jsonpath={.metadata.labels} | jq
{
  "app": "kiali",
  "app.kubernetes.io/instance": "mazzwashereagain",
  "app.kubernetes.io/name": "kiali",
  "app.kubernetes.io/part-of": "kiali",
  "app.kubernetes.io/version": "v1.34",
  "kiali.io/home": "mazzwashereagain.istio-system",
  "version": "v1.34"
}
```

Finally, the services are pointing to the proper workloads:
```
$ oc get service kiali -n istio-system -o jsonpath={.spec.selector} | jq
{
  "app.kubernetes.io/instance": "kiali",
  "app.kubernetes.io/name": "kiali"
}

$ oc get service mazzwashere -n istio-system -o jsonpath={.spec.selector} | jq
{
  "app.kubernetes.io/instance": "mazzwashere",
  "app.kubernetes.io/name": "kiali"
}

$ oc get service mazzwashereagain -n istio-system -o jsonpath={.spec.selector} | jq
{
  "app.kubernetes.io/instance": "mazzwashereagain",
  "app.kubernetes.io/name": "kiali"
}
```

jmazzitelli:
One problem with delete I just saw - the signing key is removed when you remove the Kiali Cr even if another kiali wants to still use it. Will need to fix that.

jmazzitelli:
One thing I will not consider addressing is the monitoring dashboards when installing multiple Kialis in the same deployment namespace. Since this is a rare edge case, the following will not be handled gracefully, but can be worked around.

If you elect to install multiple Kialis in the same deployment namespace, the dashboards you enabled in each Kiali CR will overwrite each other. This is fine when installing multiple Kialis, since dashboards have the same spec content whether installed by one or N Kiali instances. The problem is when you remove one Kiali CR to uninstall the one Kiali instance, the dashboards will be removed, thus all other Kiali instances will see them disappear.

The workaround is fairly simple - just tell one Kiali CR to install the dashboards again (to do this, you can simply touch the Kiali CR, thus forcing the operator to run a reconciliation). Another alternative is the user manage the dashboards manually by creating them outside of having the Kiali operator deploy them (this is how you would do it, for example, if you have your own custom dashboards, anyway).

There is talk about changing the way dashboards are published/deployed, so I won't spend any time trying to make this work more gracefully. And, again, this is only a problem if you want to install multiple Kialis in the same namespace and then you delete only one of them and want to keep the others.

jmazzitelli:
A bonus feature of this enhancement means you can share a subset of namespaces across multiple kiali instances. So kiali instance "foo" and kiali instance "bar" (even installed in different control plane namespaces) can have "my-namespace" in accessible_namespaces and not step on each other (in the past this was not possible because they shared the `kiali.io/member-of` label which would get removed if you uninstalled one but wanted to keep the other).

jmazzitelli:
All existing molecule tests pass (ran the full suite on minikube as well as the two openshift-specific tests on OpenShift 4.7). So this means backward compatibility is maintained and this doesn't break anything.

I am going to try to build a molecule test that tests this new instance_name feature - it may require some big changes to the existing molecule test infrastructure so may take some time and probably will touch a bunch of files. I'll do that in a separate commit to be included in this PR.

jmazzitelli:
new molecule test added : `instance-name-test` - passes on minikube. I will soon confirm it runs on openshift.

Note that this changed some of the common infrastructure files in the molecule test framework - so I'll need to make sure nothing breaks due to those changes - preliminary smoke tests show it is working fine, but will need to run full suite of tests on minikube and openshift to confirm all is good.

jmazzitelli:
overnight run of the molecule tests on minikube show everything is working. so nothing is broke there.

jmazzitelli:
molecule tests pass on openshift.

So all is good. This is ready for review and merge.

roles/default/kiali-deploy/templates/kubernetes/service.yaml
```diff
@@ -1,14 +1,9 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: kiali
+  name: {{ kiali_vars.deployment.instance_name }}
```
nrfox, line 4: There are some edge cases with naming kube Services. Specifically they need to conform to the [DNS label standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names). If the value of `instance_name` is being used for the `Service` name, either `instance_name` should also conform to the label standard or be messaged for the `Service` name.

deploy/kiali/kiali_cr.yaml
```diff
@@ -268,6 +273,23 @@ spec:
...
+# Note that this does not affect the name of the auto-generated signing key secret. If
+# you do not supply a signing key, the operator will create one for you in a secret,
+# but that secret will always be named "kiali-signing-key" and shared across all
+# Kiali instances in the same deployment namespace. If you want a different
+# signing key secret, you are free to create your own and tell the operator about it via
+# the login_token.signing_key setting. See the docs on that setting for more details.
+# Note also that if you are setting this value, you may also want to change
+# the installation_tag setting, but this is not required.
+#    ---
+#    instance_name: "kiali"
```
nrfox, line 291: What about using the name of the `Kiali` CR as the prefix instead of adding a new deployment field? It could work the same way where if your `Kiali` CR is named `kiali` then you don't get a prefix, everything is just called `kiali` but if your CR is named something like `admin` then the prefix would be `admin-kiali`. That's more similar to how some kube controllers name their managed resources like the `Deployment` controller and the pods associated with it. Might be less of a backwards compatible change depending on how users have named their `Kiali` CRs though.

nrfox: The Kiali server helm chart would still need a separate "prefix" value to set.

roles/default/kiali-deploy/templates/kubernetes/service.yaml
jmazzitelli: Right - the assumption is the name will follow conventions. We'll document this (I actually meant to, but forgot)... but we won't do any massaging - its simple enough to require people to use names that follow the standards for both DNS and k8s in general.

deploy/kiali/kiali_cr.yaml
jmazzitelli: A couple things:
1) right, we want to maintain consistency between how helm charts configure Kiali and how the operator does things - having a config setting allows for documentation to refer to one way to do this ("set deployment.instance_name") and it works either way. A person moving from helm to operator doesn't need to know about this "special" meaning of the Kiali CR name - all settings are in the spec. field just as they are in the helm values file.
2) Putting things in the CR like this PR does it (inside spec) means that setting automatically goes in the ConfigMap which is accessible to the Server - so the Server could get this setting if it needs to (and I suspect it may need to see this instance_name at some point in the future). And, again, this would allow for the Server to be installer-agnostic (it doesn't need to know it was installed via helm or operator - it just looks in the ConfigMap and gets the setting. Otherwise, it would have to know "Oh, I was installed by the Operator? I need to look up the Kiali CR (and it may not even have access to it or know where it is) and get its name. or... Oh, I was installed by helm? I need to look in the ConfigMap,). I suppose the server could just get its pod name and deduce what the instance_name is, but that's a pretty annoying "special case" if it wants to get a user configuration setting - when all the other ones are in the ConfigMap.

In short, we like to keep all user configuration settings in the Kiali CR spec field so they get transferred to the ConfigMap thus maintaining consistency across all user config and also making it the config available to the Server.

roles/default/kiali-deploy/templates/kubernetes/service.yaml
jmazzitelli: I'm going to update the kiali.io doc PR for this. I'll also see if I can put in a simple "test" in the operator task to check for any problems in the instance_name field... It should be trivial to add a regex check in the ansible task to check that the instance_name doesn't violate any obvious naming rules.

jmazzitelli: added blurb here: https://github.com/kiali/kiali.io/pull/374
I'll add the regex check shortly - it'll be a new commit to this PR

jmazzitelli:
Matt found an interesting issue we need to fix. Just like how the operator doesn't allow you to change instance-name, we should not allow you to change deployment.namespace either. Changing that means you are "uninstalling" from the old namespace - and the operator doesn't do that. To uninstall, you have to delete the Kiali CR - only then will the operator uninstall.

roles/default/kiali-deploy/templates/kubernetes/service.yaml
jmazzitelli: Here's what the regex will look like - and here's how I tested the regex using ansible (this is what the test will look like in the operator) - note that I restrict the name to 50 chars (not 63) because it gives us some leeway as this instance_name is only a prefix and we need more chars to prepend to it for some resource names like our cabundle configmap and other resources.

```
  - debug:
      msg: "Bad dns name: {{ item }}"
    vars:
      str: "{{ item }}"
    when:
    - str is not regex('^(?![0-9]+$)(?!-)[a-z0-9-]{,50}(?<!-)$')
    loop:
    # good names
    - "a"
    - "ab"
    - "abc"
    - "a-c"
    - "1ab"
    - "a1b"
    - "ab1"
    - "a1234567890abcdefghujklmnopqrstuvwxyz01234567890ab"
    # bad names
    - "-"
    - "-xy"
    - "xy-"
    - "x.y"
    - "XY"
    - "a1234567890abcdefghujklmnopqrstuvwxyz01234567890abx"
```

jmazzitelli: see: https://github.com/kiali/kiali-operator/pull/313/commits/e42dd9ada991b480d318adac5679b74db0cc48a6

jmazzitelli:
instance-name-test molecule test still passes.

mattmahoneyrh (approved):

PR #320, opened 2021-05-19
add status.environment so we can see what the operator thinks is its env is; remove superflous check

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4030

This adds an environment dict to the Kiali CR status field so we can tell what the operator thinks its env is.

This removes a check we shouldn't be doing, and causes problems for QE.

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -712,16 +712,6 @@
   - current_configmap.data is defined
   - current_configmap.data['config.yaml'] is defined
 
-# Never allow the instance name to change to avoid leaking resources that we fail to update or clean up.
-# I actually don't know how this will ever fail because we would not have found the configmap if the instance name is different,
-# but do this at least as a sanity check.
-- name: Do not allow user to change deployment.instance_name
-  fail:
-    msg: "The deployment.instance_name cannot be changed to a different value. It was [{{ current_instance_name }}] but is now [{{ kiali_vars.deployment.instance_name }}]. In order to install Kiali with a different deployment.instance_name, please uninstall Kiali first."
-  when:
-  - current_instance_name is defined
-  - current_instance_name != kiali_vars.deployment.instance_name
-
 # Because we need to remove the labels that were created before, we must not allow the user to change
 # the label_selector. So if the current accessible_namespaces is not ** but the label_select is being changed,
 # we need to abort since we won't know what the old labels were. If current accessible_namespaces is ** then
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Note - if you try to change `spec.version` from `v1.24` to `default` this used to fail due to the "instance name" changing (or so the error message said - that check was bad. This PR removes that bad check).

However, with that fixed, if you do the same change to spec.version, the operator will still fail but with this error (that makes more sense):

```
fatal: [localhost]: FAILED! => {"attempts": 6, "changed": false, "error": 422, "msg": "Failed to patch object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"Deployment.apps \\\\\"kiali\\\\\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\\\\\"app.kubernetes.io/instance\\\\\":\\\\\"kiali\\\\\", \\\\\"app.kubernetes.io/name\\\\\":\\\\\"kiali\\\\\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable\",\"reason\":\"Invalid\",\"details\":{\"name\":\"kiali\",\"group\":\"apps\",\"kind\":\"Deployment\",\"causes\":[{\"reason\":\"FieldValueInvalid\",\"message\":\"Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\\\\\"app.kubernetes.io/instance\\\\\":\\\\\"kiali\\\\\", \\\\\"app.kubernetes.io/name\\\\\":\\\\\"kiali\\\\\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable\",\"field\":\"spec.selector\"}]},\"code\":422}\\n'", "reason": "Unprocessable Entity", "status": 422}
```

This is because you still have Kiali v1.24 installed and you are trying to overlay the latest (aka "default") version.. and in the latest version the Kiali Deployment has a different set of label selectors - but those are immutable. If a deployment already exists, you cannot patch its label selectors.

So you cannot change spec.version from one version to another. What you should do is delete the Kiali CR and create a new one in order to change spec.version.

If you are running Maistra, you should tell Maistra to not enable Kiali so you can manage it yourself:

1. Install Maistra, let it create a Kiali CR
2. Make a copy of the Kiali CR so you can use it as a base for your own (`oc get kiali kiali -n istio-system > my-cr.yaml`)
3. Tell Maistra to disable Kiali - you do this by editing the SMCP and setting `spec.addons.kiali.enabled: false` - you can patch like this: `oc patch $($(oc get smcp -n istio-system -o name)  -n istio-system -p '{"spec":{"addons":{"kiali":{"enabled": false}}}}' --type=merge` (this should delete Kiali and the Kiali CR - at least, I think that's what it should do)
4. In `my-cr.yaml` edit `spec.version` and set it to `default` - you also have to remove some fields out of my-cr.yaml so you can "oc create -f my-cr.yaml" later (e.g. remove the status field, remove the ID fields out of metadata, etc).
5. Create your own Kiali CR: `oc create -f my-cr.yaml`

Kiali should install using "default" spec.version.

FilipB (approved):
I can see new env values in the status field as expected:
```
  environment:
    isMaistra: true
    isOpenshift: true
    kubernetesVersion: 1.20.0+7d0a2b2
    openshiftVersion: 4.7.9
    operatorVersion: v1.35.0-SNAPSHOT
```

PR #322, opened 2021-05-23
[operator][molecule] increase wait timeout when deleting resource

jmazzitelli:
molecule test fails sometimes because the default wait timeout is too short - give the test enough time to delete this kiali instance


jshaughn (approved):
fabulous

PR #325, opened 2021-06-01
move to CRD v1

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3912

Note that no helm changes are necessary - we already moved those to v1. This just moves the previous versions and OSSM metadata to v1.


jshaughn (approved):
I have review deja vu.  This again looks OK to me, as long as we are finally good for the update.

jmazzitelli:
see also: https://issues.redhat.com/browse/OSSM-580

PR #327, opened 2021-06-04
convert community operator metadata to new bundle format

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4069

see the first commit for the script I used to convert.


jmazzitelli:
once these are merged, I'll submit the PRs to the OperatorHub.io community.

jshaughn (approved):
Is convert-to-bundle.sh something we want to commit?  If so then we're good.

jmazzitelli:
> Is convert-to-bundle.sh something we want to commit? If so then we're good.

I'm committing it here just to have a record of what was run to generate the new bundles.

PR #328, opened 2021-06-09
community metadata fix

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4089

Read that issue for the problem.

This PR puts the golden copy for 1.36.0 back to the way it existed when published and it also prepares the next 1.37.0 metadata which removes the monitoringdashboard CRD and related permissions (those are no longer needed or used).

Do NOT publish the 1.37.0 metadata until the 1.37 sprint ends and we release 1.37.

### DIFF
--- a/manifests/kiali-community/1.37.0/bundle.Dockerfile
+++ b/manifests/kiali-community/1.37.0/bundle.Dockerfile
@@ -1,0 +1,11 @@
+FROM scratch
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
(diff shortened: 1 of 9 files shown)


jmazzitelli:
I think we can merge this now. We just cannot change the v1.36.0 metadata for any reason because it has already been published. If (between now and the next sprint release next week) we need to change things like operator permissions, we can release v1.36.1 with those changes.

jshaughn (approved):
This seems OK to me.  We just need to be careful with operator PRs.  But you monitor all of those so we should be OK.

PR #330, opened 2021-06-11
Bump up to 4.8 operator SDK.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4094
backport PR: https://github.com/kiali/kiali-operator/pull/375

Note that while we can move our images to 4.9, we cannot introduce any code that uses anything not also backward compatible with 4.7 because any downstream builds are still going to be on 4.7 (that's the latest downstream version we have available as of today).

I'm going to only bump to 4.8. since that is most likely going to be the next downstream build that will be available. Seems better to remain as close as possible. No sense jumping 2 minor versions here.

I also question if we even should do this now. Since downstream is still only at 4.7, we might want to hold off on merging this until downstream has 4.8. I will leave this as a "draft" PR because of this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ OPERATOR_QUAY_TAG ?= ${OPERATOR_QUAY_NAME}:${OPERATOR_CONTAINER_VERSION}
 DORP ?= docker
 
 # The version of the SDK this Makefile will download if needed, and the corresponding base image
-OPERATOR_SDK_VERSION ?= 1.3.0
-OPERATOR_BASE_IMAGE_VERSION ?= 4.7.0
+OPERATOR_SDK_VERSION ?= 1.8.0
+OPERATOR_BASE_IMAGE_VERSION ?= 4.8
 
 .PHONY: help
 help: Makefile
--- a/DEVELOPING.adoc
+++ b/DEVELOPING.adoc
@@ -52,14 +52,15 @@ Sometimes you made simple changes to the operator playbook and roles that can be
 
 In order to use this `run-operator-playbook` target, you must have Python3 in your PATH (having an alias is not enough).
 
-You also must make sure your local environment matches as closely as possible to the environment of the Kiali operator. To find out the different versions of software within the Kiali operator image, run the following:
+You also must make sure your local Python and Ansible environment matches as closely as possible to the environment of the Kiali operator. To find out the different versions of software within the Kiali operator image, run the following:
 
 * To get the Python version: `docker run --rm -i -t --entrypoint "" quay.io/kiali/kiali-operator:latest python3 --version`
 * To get the Python libraries (`openshift` and `kubernetes` are important ones to look at): `docker run --rm -i -t --entrypoint "" quay.io/kiali/kiali-operator:latest pip3 list`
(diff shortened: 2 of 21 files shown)


jmazzitelli:
molecule tests on minikube all pass:

```
              accessible-namespaces-test... success [4m 1s]
     affinity-tolerations-resources-test... success [3m 56s]
                                api-test... success [3m 31s]
                      config-values-test... success [3m 26s]
                  default-namespace-test... success [2m 23s]
                        header-auth-test... success [2m 1s]
                      instance-name-test... success [2m 58s]
                            metrics-test... success [4m 48s]
                     null-cr-values-test... success [2m 11s]
                only-view-only-mode-test... success [2m 56s]
                             openid-test... success [2m 12s]
                     openshift-auth-test... skipped
                              roles-test... success [5m 7s]
                    rolling-restart-test... success [2m 59s]
                              token-test... success [2m 16s]

```

jmazzitelli:
tests on openshift 4.7 are ok.

jmazzitelli:
I'm going to change from 4.9 to 4.8. The weird thing is, nothing changes between those two base images - the ansible operator sdk is still 1.8.0 and the kubernetes collections is still 1.2. And we have to use "4.8' not "4.8.0" because the .0 versions are all old. I have a feeling something isn't right with the ansible operator builds that are released. But for now, we will use 4.8.

jmazzitelli:
molecule tests all pass with 4.8, too.

jmazzitelli:
Note that this PR also enables metadata validation checks for the latest bundles - this should be independent of the version bump. I may rip that out and put it into a separate PR because I'm not convinced we should be moving to 4.8 now (considering the downstream version doesn't have 4.8 yet).

jmazzitelli:
This is ready for review. Moved to v4.8. This also moves all usages of the k8s plugin to `community.kubernetes.k8s` (with the intention that in the future, when the base image moves, we move to `kubernetes.core` with it).

jmazzitelli:
I am going to run the molecule tests. Once confirmed, I'll backport this PR to v1.36.

jmazzitelli:
molecule testing looks ok. backport PR created. Ready for merge

Makefile
```diff
@@ -23,8 +23,8 @@ OPERATOR_QUAY_TAG ?= ${OPERATOR_QUAY_NAME}:${OPERATOR_CONTAINER_VERSION}
 DORP ?= docker
 
 # The version of the SDK this Makefile will download if needed, and the corresponding base image
-OPERATOR_SDK_VERSION ?= 1.3.0
-OPERATOR_BASE_IMAGE_VERSION ?= 4.7.0
+OPERATOR_SDK_VERSION ?= 1.8.0
+OPERATOR_BASE_IMAGE_VERSION ?= 4.8
```
jshaughn, line 27: confirming that you meant to drop the patch version.  Probably because I see other occurrences w/o the `.0`.

jshaughn (approved):
LOK as long as you check my comment to make sure it's OK.

Makefile
jmazzitelli: yes. meant to drop. So we pick up CVEs/patch builds on rebuilds

PR #331, opened 2021-06-12
Adds a validate make target to check the metadata for correctness.

jmazzitelli:
The github CI makes use of this new target so it validates the metadata for each PR merge.


jshaughn (approved):
LOK

PR #333, opened 2021-06-18
Mount the "kiali-cabundle" configmap, if present (Kubernetes)

israel-hdez:
Similarly to the OpenShift version of the deployment, this mounts under /kiali-cabundle the configmap with name "kiali-cabundle". The difference against the OpenShift template, is that this volume is optional.

This is to support custom CA certificates for OpenID.

Related helm-charts PR: https://github.com/kiali/helm-charts/pull/82
Related back-end PR: https://github.com/kiali/kiali/pull/4112

Related kiali/kiali#4050


jmazzitelli (approved):
I didn't run this, but it is the same as what we have in the OS template with the additional "optional: true" flag. This all looks good.

PR #336, opened 2021-06-22
No need for pods/proxy sub-resource so far

xeviknal:
needs kiali/kiali#4108

There is no need to communicate with Istiod pods using the pod/proxy. Kubernetes assigns a PodIP accessible from all nodes of a cluster.

### DIFF
--- a/manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
@@ -419,7 +419,6 @@ spec:
           - configmaps
           - endpoints
           - pods/log
-          - pods/proxy
           verbs:
           - get
           - list
--- a/manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
@@ -419,6 +419,7 @@ spec:
           - configmaps
           - endpoints
           - pods/log
+          - pods/proxy
           verbs:
           - get
           - list
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -471,6 +471,7 @@ spec:
           - configmaps
           - endpoints
           - pods/log
+          - pods/proxy
           verbs:
           - get
           - list
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -471,7 +471,6 @@ spec:
           - configmaps
           - endpoints
           - pods/log
-          - pods/proxy
           verbs:
           - get
           - list
--- a/manifests/kiali-upstream/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
+++ b/manifests/kiali-upstream/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
@@ -419,7 +419,6 @@ spec:
           - configmaps
           - endpoints
           - pods/log
-          - pods/proxy
           verbs:
           - get
           - list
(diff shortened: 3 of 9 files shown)


manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
```diff
@@ -471,7 +471,6 @@ spec:
           - configmaps
           - endpoints
           - pods/log
-          - pods/proxy
```
jmazzitelli, line 474: The Kiali Operator supports several previous versions of Kiali Server. For OSSM 1.1 we support Kiali Server v1.12 and for OSSM 2.0 we support  Kiali v1.24. Can you confirm that those two versions do not need this permission?

I looked in the [v1.12](https://github.com/kiali/kiali-operator/blob/master/roles/v1.12/kiali-deploy/templates/openshift/role.yaml) and [v1.24](https://github.com/kiali/kiali-operator/blob/master/roles/v1.24/kiali-deploy/templates/openshift/role.yaml) openshift templates and I do not see it, so I think we do not need this (since no supported version of the productized Kiali Server supports this) but I want to make sure.

jmazzitelli: (note that if we do not need this, we need to backport this to the v1.36 branch - assuming this entire feature is going in v1.36)

lucasponce:
@jmazzitelli could you take a final look to this one ? thx

manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
```diff
@@ -419,7 +419,6 @@ spec:
           - configmaps
           - endpoints
           - pods/log
-          - pods/proxy
```
jmazzitelli, line 422: we already released 1.37.0 and this didn't make it. So remove this change from this 1.37 manifest. I will create a new 1.38 manifest and you can put this change in there. I will do this soon and post back when that is ready

manifests/kiali-upstream/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
```diff
@@ -419,7 +419,6 @@ spec:
           - configmaps
           - endpoints
           - pods/log
-          - pods/proxy
```
jmazzitelli, line 422: we already released 1.37.0 and this didn't make it. So remove this change from this 1.37 manifest. I will create a new 1.38 manifest and you can put this change in there. I will do this soon and post back when that is ready

jmazzitelli (changes requested):
we need to put the manifest changes to 1.38.

Also, if this feature is going into the next OSSM release (changes to kiali-ossm metadata in this PR indicate this) then do not forget to backport this PR

manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
jmazzitelli: Someone approve and merge this: https://github.com/kiali/kiali-operator/pull/358
Then you can revert this line change and make the identical change to the v1.38.0 version of the manifest.

manifests/kiali-upstream/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
jmazzitelli: Someone approve and merge this: https://github.com/kiali/kiali-operator/pull/358
Then you can revert this line change and make the identical change to the v1.38.0 version of the manifest.

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
jmazzitelli: I do not see a "backport/1.36" label on this PR - that tells me this feature is NOT going into v1.36. If this is true, then you need to remove this change because kiali-ossm is going to still need this permission.

prachiyadav (approved):

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
xeviknal: yep, you are right. This is not going to be backported to the 1.36. So the 1.36 needs this permission to function. Reverting this deletion. Thanks

xeviknal:
@jmazzitelli I've approached, or I think I did, your comments. The deletion of the permission is only for the 1.38 version (reverted for 1.37). And since the ossm will use 1.36, I've also reverted the change in its manifest.

jmazzitelli (approved):
lgtm

PR #341, opened 2021-06-25
add instructions to get things to work - things are getting weird wit…

jmazzitelli:
things are getting weird with newer python libraries being broken but the ansible base image isn't using them anyway.


jshaughn (approved):
LOK

PR #342, opened 2021-06-26
Add ui default support for graph traffic rates

jshaughn:
**Server PR:** https://github.com/kiali/kiali/pull/4113
**UI PR:** https://github.com/kiali/kiali-ui/pull/2181

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -728,7 +728,7 @@ spec:
 #    ui_defaults:
 #
 # Default settings for UI Graph
-#   ---
+#      ---
 #      graph:
 #
 # Each Find/Hide option consists of a human readable description and a valid Find/Hide expression
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -746,10 +746,13 @@ spec:
 #        - description: "Hide: unknown nodes"
 #          expression:  "name = unknown"
 #
-# Traffic settings determine which rates are used to determine graph traffic. Valid settings:
-# grpc: none | sent (messages) | received (messages) | requests | total (messages)
+# Traffic settings determine which rates are used to determine graph traffic. gRPC traffic is
+# measured in requests or sent/received/total messages.  HTTP traffic is measure in requests.
+# TCP traffic is measured in sent/received/total bytes.  Only request traffic supplies response
+# codes.  Valid values:
+# grpc: none | requests | sent | received | total
 # http: none | requests
-# tcp:  none | sent (bytes) | received (bytes) | total (bytes)
+# tcp:  none | sent | received | total
 #      ---
 #        traffic:
 #          grpc: "requests"
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -753,7 +753,7 @@ spec:
 # grpc: none | requests | sent | received | total
 # http: none | requests
 # tcp:  none | sent | received | total
-#      ---
+#        ---
 #        traffic:
 #          grpc: "requests"
 #          http: "requests"
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -727,9 +727,12 @@ spec:
 #    ---
 #    ui_defaults:
 #
-# Default presets for Graph Find/Hide
-#      ---
+# Default settings for UI Graph
+#   ---
 #      graph:
+#
+# Each Find/Hide option consists of a human readable description and a valid Find/Hide expression
+#      ---
 #        find_options:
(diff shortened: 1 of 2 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -731,6 +734,16 @@ spec:
 #        - description: "Hide: unknown nodes"
 #          expression:  "name = unknown"
 #
+# Traffic settings determine which rates are used to determine graph traffic. Valid settings:
+# grpc: none | sent (messages) | received (messages) | requests | total (messages)
```
jmazzitelli, line 738: These gRPC config settings in the CR are:
* sent
* received
* requests
* total

but the UI shows:

* response messages
* requests
* request messages
* total messages

"Requests" is the same - "total" in the CR correlates to "total messages" (as this comment indicates) - so that's fine. But the other two don't jive. The CR "sent" and "received" must correlate to the UI's "response messages" and "request messages" but its confusing. The names aren't the same. And the comments aren't helpful. `sent (messages)` ? I assume this means it is the same as the UI "request messages".

I recommend the CR settings and comments should be the same as the UI labels.

image

jmazzitelli (changes requested):

deploy/kiali/kiali_cr.yaml
jshaughn: Yeah, I get what you are saying.  I was trying to limit the number of values and just have "sent" be used to talk about traffic originating from the source and "received" for the destination.  It's just that for gRPC it's a message count and for TCP it's a byte count.   I think maybe I could just be more clear in the help, because the values to me don't have to map exactly to the display settings.  It's similar in the URL query params, just using consolidated sent/received terminology.  Tbh, I don't really like "request messages" and "response messages".  What if I changed the display options for gRPC to match what we have had for TCP forever and just be "Sent Messages" and "Received Messages"?

jmazzitelli: I don't have an opinion on WHAT the different options should be or how they are worded, but my concern is making them consistent (whatever they end up being).

Consider how a user will set these. He'll look at the UI and say, "Oh, I want the GRPC "Response Messages" to be the default. Then he'll go to the Kiali CR and ... be confused because he won't see any option "response (messages)" in the value or comments. (this is exactly what I did and what triggered my original comment).

So.. make the UI and the Kiali CR settings consistent. I don't think its important that the different traffic types (grpc, http, tcp) have the same values. They are different traffic types, so I would expect the list of options to be different. If they are all the same, bonus. But I wouldn't even expect them to be the same.

jshaughn: OK, I updated the operator and ui prs.  I did not change any of the setting values but instead consistently used the "Sent/Received" terminology.  I also updated the gRPC info tooltips mentioning a correlation between Sent/Request and Received(Response) for any user that may need that little tip.

jmazzitelli (approved):
built, tested, and ran the PRs. Modified the settings between none and non-none settings. All looks good.

jshaughn:
@jmazzitelli Can you re-approve, no changes, just rebased.

deploy/kiali/kiali_cr.yaml
```diff
@@ -727,9 +727,12 @@ spec:
 #    ---
 #    ui_defaults:
 #
-# Default presets for Graph Find/Hide
-#      ---
+# Default settings for UI Graph
+#   ---
```
jmazzitelli, line 731:
```suggestion
#      ---
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -727,9 +727,12 @@ spec:
...
 #      graph:
+#
+# Each Find/Hide option consists of a human readable description and a valid Find/Hide expression
+#      ---
```
jmazzitelli, line 735:
```suggestion
#        ---
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -743,6 +746,19 @@ spec:
...
 #          expression:  "name = unknown"
 #
+# Traffic settings determine which rates are used to determine graph traffic. gRPC traffic is
+# measured in requests or sent/received/total messages.  HTTP traffic is measure in requests.
+# TCP traffic is measured in sent/received/total bytes.  Only request traffic supplies response
+# codes.  Valid values:
+# grpc: none | requests | sent | received | total
+# http: none | requests
+# tcp:  none | sent | received | total
+#      ---
```
jmazzitelli, line 756:
```suggestion
#        ---
```

jmazzitelli (approved):
OCD alert... Those "---" markers are supposed to be indented at the same level as the settings below them. It's supposed to demarcate the indentation of the settings below it.

Feel free to correct or not, I'm approving anyway.

jmazzitelli (approved):

pbajjuri20 (approved):

PR #344, opened 2021-06-28
[v1.36][molecule] wait for test namespace to be actually deleted

jmazzitelli:
cherry pick of https://github.com/kiali/kiali-operator/pull/343

Want to cherry pick this so we can be assured the molecule tests can still run when using the v1.36 branch.


jshaughn (approved):
LOK, 2 minute timeout (as mentioned in that master PR)

PR #348, opened 2021-07-02
remove app and version fields from service selector

jmazzitelli:
When upgrading from pre-v1.35 to 1.35-and-later, we have to make sure we remove app and version fields from the service selector, otherwise, upgrades are broken.

When we no longer support operator v1.34 and below, we can remove these two null fields.

fixes: https://github.com/kiali/kiali/issues/4143
backport PR: https://github.com/kiali/kiali-operator/pull/349

Note: this doesn't work due to what appears to be a bug in the `k8s` ansible task. Unless this is fixed or we can workaround it, I don't see how we can fix this issue -- https://github.com/ansible-collections/kubernetes.core/issues/161 (UPDATE: I added an `{% if...endif %}` to workaround that ansible issue - so this should work now).

### DIFF
--- a/roles/default/kiali-deploy/templates/kubernetes/service.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/service.yaml
@@ -25,6 +25,10 @@ spec:
     port: {{ kiali_vars.server.metrics_port }}
 {% endif %}
   selector:
+{% if query('k8s', kind='Service', resource_name=kiali_vars.deployment.instance_name, namespace=kiali_vars.deployment.namespace) | length > 0 %}
+    app: null
+    version: null
+{% endif %}
     app.kubernetes.io/name: kiali
     app.kubernetes.io/instance: {{ kiali_vars.deployment.instance_name }}
   {% if kiali_vars.deployment.additional_service_yaml is defined %}{{ kiali_vars.deployment.additional_service_yaml | to_nice_yaml(indent=0) | trim | indent(2) }}{% endif %}
--- a/roles/default/kiali-deploy/templates/openshift/service.yaml
+++ b/roles/default/kiali-deploy/templates/openshift/service.yaml
@@ -23,6 +23,10 @@ spec:
     port: {{ kiali_vars.server.metrics_port }}
 {% endif %}
   selector:
+{% if query('k8s', kind='Service', resource_name=kiali_vars.deployment.instance_name, namespace=kiali_vars.deployment.namespace) | length > 0 %}
+    app: null
+    version: null
+{% endif %}
     app.kubernetes.io/name: kiali
     app.kubernetes.io/instance: {{ kiali_vars.deployment.instance_name }}
   {% if kiali_vars.deployment.additional_service_yaml is defined %}{{ kiali_vars.deployment.additional_service_yaml | to_nice_yaml(indent=0) | trim | indent(2) }}{% endif %}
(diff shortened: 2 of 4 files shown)


jmazzitelli:
To explain further what is going on:

Prior to v1.35 of the operator, the operator created a Kiali service with a selector that contains the "version" field. See: https://github.com/kiali/kiali-operator/blob/v1.34.0/roles/default/kiali-deploy/templates/kubernetes/service.yaml#L33-L34

In v1.35, that selector field "version" was removed. See: https://github.com/kiali/kiali-operator/blob/v1.35.0/roles/default/kiali-deploy/templates/kubernetes/service.yaml#L27-L29

In order to support upgrading from pre-v1.35 to v1.35-or-higher, this PR asks the `k8s` task to remove any existing "version" field from the selector. This works if you really are upgrading (that is, there is already a service resource that exists).

However, if you are installing v1.35 or higher (not doing an upgrade, just a clean install) this doesn't work as intended - the `k8s` task will put `version: ""` in the selector, which is obviously not what we want.

This unexpected behavior has been logged at https://github.com/ansible-collections/kubernetes.core/issues/161.

jmazzitelli:
Any solution to this should be testable using this mechanism (this assumes you can build and push dev images to minikube - so you need minikube installed and you need a dev environment).

1. First make sure you have Istio installed in a minikube cluster.
2. Create this CR yaml file (this is used to install and delete the Kiali CR inside the istio-system namespace later)
```bash
cat <<EOM > /tmp/kiali-cr-test.yaml
> apiVersion: kiali.io/v1alpha1
> kind: Kiali
> metadata:
>   name: kiali
>   namespace: istio-system
>   annotations:
>     ansible.sdk.operatorframework.io/verbosity: "1"
> spec:
>   istio_namespace: istio-system
>   auth:
>     strategy: anonymous
>   deployment:
>     accessible_namespaces: ['**']
>     ingress_enabled: false
>     image_version: operator_version
> EOM
```
3. Install Kiali Operator v1.34 using the published helm chart: `helm install --create-namespace --namespace kiali-operator --repo https://kiali.org/helm-charts --version 1.34.0 --set cr.create=false --set allowAdHocKialiImage=true kiali-operator kiali-operator`
4. Install the Kiali CR so Kiali Server v1.34 is installed: `kubectl apply -f /tmp/kiali-cr-test.yaml`
5. Wait for the kiali server's service to be created and confirm it has the old selector fields of "app" and "version": `kubectl get svc kiali -n istio-system -o jsonpath={.spec.selector}`
6. Checkout this PR in your local operator repo, and from the kiali/kiali repo, run the make target to build/push the operator image to minikube: `make -e CLUSTER_TYPE=minikube cluster-push-operator`
7. In your local kiali helm-chart repo, use make to build the helm charts so you are up to date with master (this PR currently requires no changes to the helm chart so using master branch is fine): `make clean build-helm-charts`
8. Now upgrade the operator to the new operator (this PR version of the operator). This should upgrade the existing Kiali server because the new operator should immediately reconcile the existing Kiali CR: `helm upgrade --namespace kiali-operator --set cr.create=false --set allowAdHocKialiImage=true --set image.tag=dev --set image.repo=localhost:5000/kiali/kiali-operator kiali-operator <the full path to the root of your kiali helm-charts local repo>/_output/charts/kiali-operator-*-SNAPSHOT.tgz`
9. Wait for the kiali server's service to be updated and confirm the old selector fields of "app" and "version" are gone and the 2 new ones ("app.kubernetes.io/instance" and "app.kubernetes.io/name") are created: `kubectl get svc kiali -n istio-system -o jsonpath={.spec.selector}`

jmazzitelli:
Not crazy about it, but I have a relative simple-to-implement solution. It performs an additional k8s query when creating the service to determine if the "null" yaml should be include in the template. I tested it (with the above test procedures that I documented in the earlier comment) and it works.

leandroberetta (approved):
Verified on my local enviroment, instructions to test worked ok.

PR #350, opened 2021-07-06
ensure the label value is quoted in case the user wants to set this as a boolean

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4157

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -789,7 +789,7 @@
       metadata:
         name: "{{ namespace }}"
         labels:
-          {{ the_namespace_label_name }}: {{ the_namespace_label_value }}
+          {{ the_namespace_label_name }}: "{{ the_namespace_label_value }}"
       ...
       {% endfor %}
   when:


jmazzitelli:
This fix protects against the use-case where the user wants to set the label_selector whose value is a boolean. The label_selector (when used with non-`**` accessible namespaces) will also be used to label the namespaces. The value must not be a boolean because k8s will not allow you to create labels whose values are boolean (they must be strings). So kiali operator will always quote the value to ensure `true` or `false` is wrapped in quotes so they become strings.

jmazzitelli:
To test this PR... build and push the operator image to your cluster. Then run:

```
helm install \
  --create-namespace \
  --namespace kiali-operator \
  --repo https://kiali.org/helm-charts \
  --set cr.create=true \
  --set allowAdHocKialiImage=true \
  --set image.tag=dev \
  --set image.repo=localhost:5000/kiali/kiali-operator \
  --set cr.spec.deployment.accessible_namespaces[0]=default \
  --set cr.spec.api.namespaces.label_selector=kiali-enabled=true \
  kiali-operator \
  kiali-operator
```

You can confirm this is fixed if the "default" namespace has a label "kiali-enabled" whose value is `"true"` (specifically, it will be a string that is quoted ):

```
$ kubectl get ns default -o yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    kiali-enabled: "true"
  name: default
...
```

jshaughn (approved):
LOK, from what I can see we don't actually have to worry about the value being quoted initially.

jmazzitelli:
accessible-namespace-test molecule test passes. we can merge.

PR #351, opened 2021-07-08
only hide system namespaces that start with "kube-"

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4162

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -102,7 +102,7 @@ spec:
 #    namespaces:
 #      exclude:
 #      - "istio-operator"
-#      - "kube.*"
+#      - "kube-.*"
 #      - "openshift.*"
 #      - "ibm.*"
 #      - "kiali-operator"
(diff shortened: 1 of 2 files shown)


jmazzitelli:
Run with this, and create a namespace "kubeless" and you will now see it:

image

Here's what the Kiali CR and resulting ConfigMap looks like:

```
$ kubectl get cm kiali -n istio-system -o yaml
...
    api:
      namespaces:
        exclude:
        - istio-operator
        - kube-.*
        - openshift.*
        - ibm.*
        - kiali-operator
...
```

If you do not set accessible namespaces, you should now see the namespaces in the list by default (the default value of accessible namespaces will now not include `kube-.*`)

jmazzitelli:
To test this, create a Kiali CR that does NOT set accessible_namespaces to any value (pick up the default). Create a namespace "kubefoo". See that "kubefoo" appears in the Kiali UI.

jshaughn (approved):
LOK

PR #352, opened 2021-07-08
molecule test for the new dashboard annotations

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4160

Goes with issue https://github.com/kiali/kiali/issues/4145 and PR https://github.com/kiali/kiali/pull/4146 - only when that PR gets merged should we merge this PR since this new molecule test will only pass with that new code.

### DIFF
--- a/molecule/api-test/cleanup.yml
+++ b/molecule/api-test/cleanup.yml
@@ -8,5 +8,16 @@
   # Cleanup should only ever execute when a molecule test fails
   # api-test has its own cleanup because we need to ensure the simple-mesh namespace is deleted
 
+  - name: Get information about the cluster
+    set_fact:
+      api_groups: "{{ lookup('kubernetes.core.k8s', cluster_info='api_groups') }}"
+  - name: Determine the cluster type
+    set_fact:
+      is_openshift: "{{ True if 'route.openshift.io' in api_groups else False }}"
+      is_k8s: "{{ False if 'route.openshift.io' in api_groups else True }}"
+  - name: Determine the Istio implementation
+    set_fact:
+      is_maistra: "{{ True if 'maistra.io' in api_groups else False }}"
+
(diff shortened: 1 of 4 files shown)


jmazzitelli:
I'm running this with PR https://github.com/kiali/kiali/pull/4146 and I'm seeing errors. I think there might be a problem there.

jmazzitelli:
Here's a way to test what this molecule test is doing - I documented this over in the github issue: https://github.com/kiali/kiali/pull/4146#issuecomment-876752008

jmazzitelli:
`make -e CLUSTER_TYPE=minikube -e MOLECULE_SCENARIO="api-test" -e MOLECULE_USE_DEV_IMAGES=true molecule-test`

That will run the api-test which includes this new dashboard set of tests.

jmazzitelli:
These tests (correctly) fail right now given the current state of the dashboard PRs - once those are fixed, these tests should pass.

jmazzitelli:
This molecule test now passes with the new dashboard annotations PR. We can merge this when that gets merged also.

jmazzitelli:
@lucasponce when you merge your [dashboard annotations PR](https://github.com/kiali/kiali/issues/4145), please approve and merge this one (this contains molecule tests that show your PR working).

jmazzitelli:
We [need to](https://github.com/kiali/kiali/pull/4146#issuecomment-879093471) modify this test so the Namespace is created as a Project when run on OpenShift.

jmazzitelli:
This passes on minikube but on OpenShift it fails now (given the current state of the server-side PR) - looks like it is due to a problem in server-side PR and its handling of Project vs.Namespace. But this molecule test should be good - and actually is doing its job detecting problems in the server. Once that problem is fixed, this test should all pass on OpenShift.

lucasponce (approved):

PR #353, opened 2021-07-09
operator installmodes - only support AllNamespaces

jmazzitelli:
OLM is going to deprecate all install modes except "AllNamespaces" (at least, currently, they are telling operator authors to move away from the others now).

### DIFF
--- a/manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.37.0/manifests/kiali.v1.37.0.clusterserviceversion.yaml
@@ -122,9 +122,9 @@ spec:
     url: https://github.com/kiali/kiali-operator
   installModes:
   - type: OwnNamespace
-    supported: true
+    supported: false
   - type: SingleNamespace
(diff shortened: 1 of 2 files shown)


jmazzitelli:
The reason for this change:

```
From: Daniel Messer
Date: Fri, Jul 9, 2021 at 10:35 AM
Subject: Re: specifying a dependency of an operator on another operator
Cc: Operator Framework <operator-framework@googlegroups.com>

As for the general question about installmodes - this is something the we are moving
away from as part of the descoping exercise, in which all Operator just become global
singleton's and we have more precise RBAC controls that define where they are
available to be discovered by tenants, which tenants actually have usage rights and
inversely, in which namespaces the operator can become active.

Our recommendation today is to only support AllNamespace mode in your Operator.
That way dependencies will not run into the conflict you describe here.
```

israel-hdez (approved):
I guess, in general, we are not using those modes when installing with OLM.
But when installing through Helm, I guess we still want to support installing the operator with less privileges.

jshaughn (approved):
LOK

PR #357, opened 2021-07-14
Adding Jaeger (internal) integration basic test

leandroberetta:
This PR adds a new test to validate that Kiali can communicate with the in cluster Jaeger instance. It does it with the Istio components statuses, if a Jaeger object in the list is present, it means an error.

Fixes kiali/kiali#3028

### DIFF
--- a/molecule/jaeger-test/converge.yml
+++ b/molecule/jaeger-test/converge.yml
@@ -1,0 +1,65 @@
+- name: Tests
+  hosts: localhost
+  connection: local
+  collections:
+  - kubernetes.core
+  vars:
+    custom_resource: "{{ lookup('template', cr_file_path) | from_yaml }}"
+  tasks:
+  - import_tasks: ../common/tasks.yml
+  - import_tasks: ../asserts/pod_asserts.yml
+  - import_tasks: ../common/wait_for_kiali_running.yml
+  
+  - name: Get statuses from Istio components
+    uri:
+      url: "{{ kiali_base_url }}/api/istio/status"
+      validate_certs: no
+      return_content: yes      
+    register: status_response
+  
+  - name: "Results: /api/istio/status"
+    debug:
+      msg: "{{ status_response.json }}"  
+
+  - name: Check if Jaeger status if present (if it is present it means error)
+    set_fact:
+      jaeger_status: "{{ item.status }}"     
+    loop: "{{ status_response.json }}"
+    when: "item.name == 'jaeger'"
+  
+  - name: Assert that Jaeger status if not in the response
+    assert:
+      that:
+      - jaeger_status is not defined
+
+  # Update Jaeger URL to a bad URL just to test a bad integration
+  - import_tasks: update-jaeger-url.yml
+  - import_tasks: ../common/wait_for_kiali_cr_changes.yml
+  - import_tasks: ../common/wait_for_kiali_running.yml
+  - import_tasks: ../common/tasks.yml
+  - import_tasks: ../asserts/pod_asserts.yml
+
+  - name: Get statuses from Istio components
+    uri:
+      url: "{{ kiali_base_url }}/api/istio/status"
+      validate_certs: no
+      return_content: yes      
+    register: status_response
+  
+  - name: "Results: /api/istio/status"
+    debug:
+      msg: "{{ status_response.json }}"  
+
+  - set_fact:
+      current_kiali_cr: "{{ lookup('kubernetes.core.k8s', api_version='kiali.io/v1alpha1', kind='Kiali', namespace=cr_namespace, resource_name=custom_resource.metadata.name) }}"
+
+  - name: Check if Jaeger status if present (presence means an error)
+    set_fact:
+      jaeger_status: "{{ item.status }}"     
+    loop: "{{ status_response.json }}"
+    when: "item.name == 'jaeger'"
+  
+  - name: Assert that Jaeger status is in the response
+    assert:
+      that:
+      - jaeger_status is defined
--- a/molecule/jaeger-test/converge.yml
+++ b/molecule/jaeger-test/converge.yml
@@ -21,16 +21,18 @@
     debug:
       msg: "{{ status_response.json }}"  
 
-  - name: Check if Jaeger status if present (if it is present it means error)
+  # IstioStatus API returns a statuses list of external components (Jaeger is one of them)
+  # If Jaeger is in the list, it means that there is an error (Unreacheable)
+  - name: Check if Jaeger status is present
     set_fact:
-      jaeger_status: "{{ item.status }}"     
+      jaeger_error: "{{ item.status }}"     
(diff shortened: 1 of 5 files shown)


molecule/jaeger-test/converge.yml
```diff
@@ -0,0 +1,65 @@
...
+      url: "{{ kiali_base_url }}/api/istio/status"
+      validate_certs: no
+      return_content: yes      
+    register: status_response
+  
+  - name: "Results: /api/istio/status"
+    debug:
+      msg: "{{ status_response.json }}"  
+
+  - name: Check if Jaeger status if present (if it is present it means error)
```
jmazzitelli, line 24: typo.
```suggestion
  - name: Check if Jaeger status is present (if it is present it means error)
```

molecule/jaeger-test/converge.yml
```diff
@@ -0,0 +1,65 @@
...
+    register: status_response
+  
+  - name: "Results: /api/istio/status"
+    debug:
+      msg: "{{ status_response.json }}"  
+
+  - set_fact:
+      current_kiali_cr: "{{ lookup('kubernetes.core.k8s', api_version='kiali.io/v1alpha1', kind='Kiali', namespace=cr_namespace, resource_name=custom_resource.metadata.name) }}"
+
+  - name: Check if Jaeger status if present (presence means an error)
```
jmazzitelli, line 56: typo 
```suggestion
  - name: Check if Jaeger status is present (presence means an error)
```

jmazzitelli (changes requested):
just some typos to fix
I have a question though. When it  says, "if it is present it means error" - that is confusing. Does this mean that `api/istio/status` endpoint returns error messages and if "jaeger" is present it means there is an error message regarding jaeger? I think that is what this means, but I just want to be sure.

leandroberetta:
> I have a question though. When it says, "if it is present it means error" - that is confusing. Does this mean that `api/istio/status` endpoint returns error messages and if "jaeger" is present it means there is an error message regarding jaeger? I think that is what this means, but I just want to be sure.

That's exactly how it works, the IstioStatus API returns a list of "Unreacheable" external components like Jaeger. Thanks for your review, I'll add better comments in the test to clarify the situation.

jmazzitelli:
@leandroberetta I will be looking at this - one thing our molecule tests support (or did, hopefully still do) is running not only with upstream Istio but also with Service Mesh. I am going to run these tests to make sure this also passes with Service Mesh and its Jaeger installation (they have authentication proxy in front of all the addons, so this will confirm kiali gets the proper credentials from Service Mesh). It may take me some time to get there because I have to make sure I get Service Mesh installed properly. But this PR is definitely on my radar to review.

leandroberetta:
@lucasponce @jmazzitelli this test is basic but I think that covers what is mentioned in the issue. What do you think?.

leandroberetta:
> @leandroberetta I will be looking at this - one thing our molecule tests support (or did, hopefully still do) is running not only with upstream Istio but also with Service Mesh. I am going to run these tests to make sure this also passes with Service Mesh and its Jaeger installation (they have authentication proxy in front of all the addons, so this will confirm kiali gets the proper credentials from Service Mesh). It may take me some time to get there because I have to make sure I get Service Mesh installed properly. But this PR is definitely on my radar to review.

Sorry I missed this comment, yes, good point and thanks

jmazzitelli:
These tests pass when testing with minikube and upstream Istio.

However, there is a failure when testing with Service Mesh. This is because the test Kiali CR does not set up the Jaeger configuration that works with Service Mesh. Our molecule framework does set this up for Prometheus - we just need to do the same for Jaeger. If you look at the geneated Kiali CR that ends up getting created by our tests, you will see this when running with Service Mesh:

```
                    "prometheus": {
                        "auth": {
                            "ca_file": "/kiali-cabundle/service-ca.crt",
                            "insecure_skip_verify": false,
                            "password": "AbC...some long password...XyZ",
                            "token": "",
                            "type": "basic",
                            "use_kiali_token": false,
                            "username": "internal"
                        },
```

But the tracing does not have this - we need to set `username`, `password`, and `type` here as well just like how we do it for Prometheus:

```
                    "tracing": {
                        "auth": {
                            "ca_file": "/kiali-cabundle/service-ca.crt",
                            "insecure_skip_verify": false,
                            "password": "",  <<------- NEEDS TO BE SET
                            "token": "",
                            "type": "none",  <<------- NEEDS TO BE SET
                            "use_kiali_token": false,
                            "username": ""   <<------- NEEDS TO BE SET
                        },
```

The magic happens in [default/prepare.yml](https://github.com/kiali/kiali-operator/blob/v1.37.0/molecule/default/prepare.yml#L29-L41) and [here](https://github.com/kiali/kiali-operator/blob/v1.37.0/molecule/default/prepare.yml#L123-L128) further down in that file. We just need to set the tracing config as well here. So something like this:

Below the task "Define url and auth for Prometheus for Service Mesh":

```
  - name: Define url and auth for Tracing for Service Mesh
    set_fact:
      tracing_config:
        url: "https://tracing.{{ istio.control_plane_namespace }}.svc:16686"
        auth:
          username: internal
          type: basic
          use_kiali_token: false
          password: "{{ ossm_secret_htpasswd_raw.resources[0].data.rawPassword | b64decode }}"
    when:
    - ossm_secret_htpasswd_raw is defined
    - ossm_secret_htpasswd_raw.resources is defined
    - ossm_secret_htpasswd_raw.resources | length == 1
```

and modify the task "Create Kiali CR with specific prometheus config added" by renaming it and adding the tracing config to it:

```
  - name: Create Kiali CR with specific prometheus and tracing config added
    k8s:
      namespace: "{{ cr_namespace }}"
      definition: "{{ lookup('template', cr_file_path) | from_yaml | combine({'spec':{'external_services':{'prometheus': prometheus_config, 'tracing': tracing_config }}}, recursive=True) | to_yaml }}"
    when:
    - prometheus_config is defined
```

I'm not sure if that is all correct. In fact, let me just try it and see if this works. If it is, I'll post the diff of the changes.

jmazzitelli:
See my PR to this PR: https://github.com/kiali/kiali-operator/pull/357

Still trying to get it to pass.

jmazzitelli:
I got this to pass with Service Mesh when I added this code - https://github.com/leandroberetta/kiali-operator/pull/1
So merge that into here and this PR will be good.

jmazzitelli (approved):
this is good now - works on minikube/upstream Istio and OpenShift/ServiceMesh.

jmazzitelli:
@leandroberetta let's backport this to the v1.36 branch. That way we can run it during our testing of the v1.36 builds. I added the backport label.

leandroberetta:
> @leandroberetta let's backport this to the v1.36 branch. That way we can run it during our testing of the v1.36 builds. I added the backport label.

Great, good idea!

PR #359, opened 2021-07-16
don't look at tracing settings - molecule framework overrides it if on service mesh

jmazzitelli:
this is a side effect of https://github.com/kiali/kiali-operator/pull/357


leandroberetta (approved):
LGTM.

PR #362, opened 2021-07-16
[v1.36] fix molecule tests to pass when running with OSSM

jmazzitelli:
* determine the name of the SMCP at runtime to avoid hardcoding it - some environments might not name the SMCP the same

* looks like OSSM 2.1 changed the metric labels - metrics test now passes but only for OSSM 2.1 now


jshaughn (approved):
LOK

PR #364, opened 2021-07-19
CSV metadata annotation value should be all lowercase

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4163


jshaughn (approved):
good

PR #365, opened 2021-07-19
Move to new community repos 4190

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4190

Please merge this using "merge" not "squash" - I would like to keep these commits separate so we can track them individually.


jshaughn (approved):
LOK

PR #370, opened 2021-07-26
Change default grpc port for tracing when grpc is enabled

xeviknal:
fix https://github.com/kiali/kiali/issues/4207

When grpc is enabled for Jaeger, use the exclusive port for gRPC traffic that the service has.

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -285,11 +285,22 @@
   when:
     kiali_vars.external_services.tracing.in_cluster_url == "" and is_maistra
 
-- name: Set default Tracing in_cluster_url
+- name: Set default Tracing in_cluster_url for grpc consumption
+  set_fact:
+    kiali_vars: "{{ kiali_vars | combine({'external_services': {'tracing': {'in_cluster_url': 'http://tracing.' + kiali_vars.istio_namespace + ':16685/jaeger'}}}, recursive=True) }}"
+  when:
+  - is_maistra == False
+  - kiali_vars.external_services.tracing.in_cluster_url == ""
+  - kiali_vars.external_services.tracing.use_grpc is not defined or kiali_vars.external_services.tracing.use_grpc == True
+
+- name: Set default Tracing in_cluster_url for http consumption
   set_fact:
     kiali_vars: "{{ kiali_vars | combine({'external_services': {'tracing': {'in_cluster_url': 'http://tracing.' + kiali_vars.istio_namespace + '/jaeger'}}}, recursive=True) }}"
   when:
-    kiali_vars.external_services.tracing.in_cluster_url == "" and is_maistra == False
+  - is_maistra == False
+  - kiali_vars.external_services.tracing.in_cluster_url == ""
+  - kiali_vars.external_services.tracing.use_grpc is defined
+  - kiali_vars.external_services.tracing.use_grpc == False
 
 - name: Set default Istio service that provides version info (istiod service that was introduced in Istio 1.6)
   set_fact:
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -630,7 +630,8 @@ spec:
 # health_check_url: Used in the Components health feature. This is the url which kiali will ping to determine whether the component is reachable or not. It defaults to `in_cluster_url` when not provided.
 # in_cluster_url: Set URL for in-cluster access, which enables further integration between Kiali and Jaeger.
 #      When not provided, Kiali will only show external links using the "url" config.
-#      Example: "http://tracing.istio-system".
+#      Note: Jaeger v1.20+ has separated ports for GRPC(16685) and HTTP(16686) requests. Make sure you use the appropriate port according to the "use_grpc" value.
+#      Example: "http://tracing.istio-system:16685".
 # is_core: Used in the Components health feature. When true, the unhealthy scenarios will be raised as errors. Otherwise, they will be raised as a warning.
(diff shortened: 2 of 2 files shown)


roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -285,11 +285,22 @@
   when:
     kiali_vars.external_services.tracing.in_cluster_url == "" and is_maistra
 
-- name: Set default Tracing in_cluster_url
+- name: Set default Tracing in_cluster_url for grpc consumption
+  set_fact:
+    kiali_vars: "{{ kiali_vars | combine({'external_services': {'tracing': {'in_cluster_url': 'http://tracing.' + kiali_vars.istio_namespace + ':16685/jaeger'}}}, recursive=True) }}"
```
jmazzitelli, line 290: Just pointing out this is 16685 (when deployed with Maistra it is 16686 -- see above line 284) - just want to make sure this isn't a typo and you meant it to end with "5".

jmazzitelli (approved):
This looks correct - approved. But see my question below.

jmazzitelli:
@xeviknal is this going to be backported to v1.36? If so, you will also need to update the [v1.36 ansible role in this PR](https://github.com/kiali/kiali-operator/tree/master/roles/v1.36) before you cherry-pick this to the v1.36 branch.

roles/default/kiali-deploy/tasks/main.yml
xeviknal: Thanks for the double check. The new port for gRPC is the `16685` while in older versions the opened port was 16686 for both HTTP and gRPC. In new versions, they moved all the traffic to port 80 (which is the version we currently support in master). And now it is splitted: 80:16686 for HTTP and 16685:16685 for gRPC.

```
$ k get svc -n istio-system tracing -o yaml | more
...
spec:
  clusterIP: 10.103.72.232
  clusterIPs:
  - 10.103.72.232
  ports:
  - name: http-query
    port: 80
    protocol: TCP
    targetPort: 16686
  - name: grpc-query
    port: 16685
    protocol: TCP
    targetPort: 16685

``

Comment (jmazzitelli) [reply roles/default/kiali-deploy/tasks/main.yml:290]: Never mind - I see in the jaeger PR that the port is 16685: https://github.com/jaegertracing/jaeger/pull/2387/files#diff-872d5d3fba389504d376a8dc71533d58db75d6e4bcbceaace2cf383d8352a62aR42

**Comment by xeviknal** [comment]
@jmazzitelli I think it is not necessary to backport it because maistra doesn't need these changes.
And if maistra used a 1.20+ version and used the grpc, we'd need a PR to approach it anyway because won't help.
Plus, I think that this bugfix is not worth for upstream. They can just bump to next release or even just change the kiali-cr with the new port.
wdyt?

File: roles/default/kiali-deploy/tasks/main.yml (hunk 1 above)
Comment (xeviknal) [reply roles/default/kiali-deploy/tasks/main.yml:290]: Note that this is the pod port, not the service one. Therefore, we should rely more on this one: https://github.com/istio/istio/blob/master/samples/addons/jaeger.yaml#L72
However, as usual, we really rely on user's deployment and this might change.

**Comment by jmazzitelli** [comment]
OK. We can not worry about back port - just merge to master

**Comment by xeviknal** [comment]
@jmazzitelli I've added the comment as you and @jshaughn mentioned in the issue. Feel free to suggest a more appropriate one. I haven't been in touch with the muses of inspiration lately. Thanks

**Review by jmazzitelli — APPROVED** [review]
lgtm

File: roles/default/kiali-deploy/tasks/main.yml [hunk 2]
```diff
@@ -285,11 +285,22 @@
...
+  - is_maistra == False
+  - kiali_vars.external_services.tracing.in_cluster_url == ""
+  - kiali_vars.external_services.tracing.use_grpc is not defined or kiali_vars.external_services.tracing.use_grpc == True
+
+- name: Set default Tracing in_cluster_url for http consumption
   set_fact:
     kiali_vars: "{{ kiali_vars | combine({'external_services': {'tracing': {'in_cluster_url': 'http://tracing.' + kiali_vars.istio_namespace + '/jaeger'}}}, recursive=True) }}"
   when:
-    kiali_vars.external_services.tracing.in_cluster_url == "" and is_maistra == False
+  - is_maistra == False
```
jmazzitelli, line 300: I just noticed this - why do we have is_maistra here? If we are running in Maistra but we are not created by the Maistra operator, don't we want this default?

If created by the maistra operator, this is already set, and the `kiali_vars.external_services.tracing.in_cluster_url == ""` will be false and this still won't be executed.

I have a feeling we need to take out these is_maistra==False conditionals because it will most likely screw up devs when they deploy their own Kiali CR in a maistra environment

xeviknal: To be completely honest, I don't know. Just expanded the existing expression :sweat_smile: 
It sounded that we didn't want to do the port differentiation in maistra because it uses different ports.
If the tasks are executed in order of declaration, this maistra condition wouldn't be necessary however it makes it fragile.
The role `Set default Tracing in_cluster_url for Maistra` sets the url for maistra. The other Tracing roles do it for upstream (and specifically have the condition of !is_maistra)

jmazzitelli: > The role `Set default Tracing in_cluster_url for Maistra` sets the url for maistra.

Oh, yeah, I didn't notice that - the existing task above will set this "if Maistra" so the following tasks won't be invoked if on Maistra anyway since that variable will have been set already and will not be null.

OK. We can leave this.

PR #374, opened 2021-07-26
sort, and show defaults properly

jmazzitelli:
this file is different between master and v1.36 - ensure settings are sorted so diff's are easy to see what's changed.


jshaughn (approved):
Better

PR #378, opened 2021-07-29
Grant permissions to Kiali sa to read some secrets according to feature flag configuration

leandroberetta:
Needed by [#4227](https://github.com/kiali/kiali/pull/4227)

### DIFF
--- a/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
@@ -278,14 +278,6 @@ spec:
           - watch
         - apiGroups: [""]
           resourceNames:
-          - cacerts
-          - istio-ca-secret
-          resources:
-          - secrets
-          verbs:
-          - get
-        - apiGroups: [""]
-          resourceNames:
           - kiali-signing-key
           resources:
           - secrets
--- a/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
@@ -274,6 +274,7 @@ spec:
           - secrets
           verbs:
           - create
+          - get
           - list
           - watch
         - apiGroups: [""]
--- a/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
@@ -274,9 +274,16 @@ spec:
           - secrets
           verbs:
           - create
-          - get
           - list
           - watch
+        - apiGroups: [""]
+          resourceNames:
+          - cacerts
+          - istio-ca-secret
+          resources:
+          - secrets
+          verbs:
+          - get
         - apiGroups: [""]
           resourceNames:
           - kiali-signing-key
--- a/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
+++ b/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
@@ -278,14 +278,6 @@ spec:
           - watch
         - apiGroups: [""]
           resourceNames:
-          - cacerts
-          - istio-ca-secret
-          resources:
-          - secrets
-          verbs:
-          - get
-        - apiGroups: [""]
-          resourceNames:
           - kiali-signing-key
           resources:
           - secrets
--- a/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
+++ b/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
@@ -274,6 +274,7 @@ spec:
           - secrets
           verbs:
           - create
+          - get
           - list
           - watch
         - apiGroups: [""]
--- a/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
+++ b/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
@@ -274,9 +274,16 @@ spec:
           - secrets
           verbs:
           - create
-          - get
           - list
           - watch
+        - apiGroups: [""]
+          resourceNames:
+          - cacerts
+          - istio-ca-secret
+          resources:
+          - secrets
+          verbs:
+          - get
         - apiGroups: [""]
           resourceNames:
           - kiali-signing-key
--- a/molecule/read-certs-secrets-test/converge.yml
+++ b/molecule/read-certs-secrets-test/converge.yml
@@ -1,0 +1,46 @@
+- name: Tests
+  hosts: localhost
+  connection: local
+  collections:
+  - kubernetes.core
+  vars:
+    custom_resource: "{{ lookup('template', cr_file_path) | from_yaml }}"
+  tasks:
+  - import_tasks: ../common/tasks.yml
+  - import_tasks: ../asserts/pod_asserts.yml
+  - import_tasks: ../common/wait_for_kiali_running.yml
+  
+  - name: Get control plane role
+    vars:
+      instance_name: "{{ kiali.instance_name | default('kiali') }}"
+    k8s_info:
+      api_version: rbac.authorization.k8s.io/v1
+      kind: Role
+      name: "{{ instance_name }}-controlplane"
+      namespace: "{{ istio.control_plane_namespace }}"
+    register: controlplane_role
+
+  - name: Gather the rules that contains resource names
+    set_fact:
+      rules: "{{ controlplane_role.resources[0].rules | selectattr('resourceNames','defined') }}"
+
+  - name: Get the needed role for reading certificates
+    vars:
+      istio_secret: "istio-ca-secret"
+      user_secret: "cacerts"
+    set_fact:
+      cert_role: "{{ item }}"
+    with_items: "{{ rules }}"
+    when: "{{ istio_secret in item.resourceNames and user_secret in item.resourceNames }}"
+
+  - name: Check that the role has the right verbs
+    vars:
+      get_verb: "get"
+      list_verb: "list"
+      watch_verb: "watch"
+    assert:
+      that: 
+      - cert_role is defined
+      - get_verb in cert_role.verbs
+      - list_verb in cert_role.verbs
+      - watch_verb in cert_role.verbs
--- a/molecule/read-certs-secrets-test/converge.yml
+++ b/molecule/read-certs-secrets-test/converge.yml
@@ -41,6 +41,7 @@
     assert:
       that: 
       - cert_role is defined
+      - cert_role.verbs | length == 3
       - get_verb in cert_role.verbs
       - list_verb in cert_role.verbs
       - watch_verb in cert_role.verbs
--- a/molecule/read-certs-secrets-test/molecule.yml
+++ b/molecule/read-certs-secrets-test/molecule.yml
@@ -1,0 +1,46 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: $DORP
+platforms:
+- name: default
+  groups:
+  - k8s
+provisioner:
+  name: ansible
+  config_options:
+    defaults:
+      callback_whitelist: junit
+  playbooks:
+    destroy: ../default/destroy.yml
+    prepare: ../default/prepare.yml
+    cleanup: ../default/cleanup.yml
+  inventory:
+    group_vars:
+      all:
+        cr_file_path: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') }}/molecule/jaeger-test/kiali-cr.yaml"
+        cr_namespace: "{{ 'kiali-operator' if (lookup('env', 'MOLECULE_OPERATOR_INSTALLER') | default('helm', True) == 'helm') else 'istio-system' }}" # if external operator, assume CR must go in control plane namespace
+        wait_retries: "{{ lookup('env', 'MOLECULE_WAIT_RETRIES') | default('360', True) }}"
+        istio:
+          control_plane_namespace: istio-system
+        kiali:
+          spec_version: "{{ lookup('env', 'MOLECULE_KIALI_CR_SPEC_VERSION') | default('default', True) }}"
+          install_namespace: istio-system
+          accessible_namespaces: ["**"]
+          auth_strategy: anonymous
+          operator_namespace: "{{ 'kiali-operator' if (lookup('env', 'MOLECULE_OPERATOR_INSTALLER') | default('helm', True) == 'helm') else ('openshift-operators' if (query('kubernetes.core.k8s', kind='Namespace', resource_name='openshift-operators') | length > 0) else 'operators') }}" # if external operator, assume operator is in OLM location
+          operator_image_name: "{{ 'image-registry.openshift-image-registry.svc:5000/kiali/kiali-operator' if lookup('env', 'MOLECULE_KIALI_OPERATOR_IMAGE_NAME') == 'dev' else (lookup('env', 'MOLECULE_KIALI_OPERATOR_IMAGE_NAME')|default('quay.io/kiali/kiali-operator', True)) }}"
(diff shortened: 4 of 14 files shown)


jmazzitelli (changes requested):
This is a problem. We are going to need a project-wide discussion before we merge. Do not merge until this is vetted by the team and community.

We specifically have been avoiding adding this "read any secret" permission. People do not like it (we recently had a community member complain about our current ability to read secrets in the control plane namepace). This is a very, very bad (potential) security hold that kiali would be opening up.

We need to avoid this permission. If you need to read a specific secret, we can do so by name as we do here: https://github.com/kiali/helm-charts/blob/master/kiali-operator/templates/clusterrole.yaml#L41-L52

leandroberetta:
> We need to avoid this permission. If you need to read a specific secret, we can do so by name as we do here: https://github.com/kiali/helm-charts/blob/master/kiali-operator/templates/clusterrole.yaml#L41-L52

I understand, reading specific secrets by name would work for some cases of this feature set, but there is one case that we need to read secrets configured in istio configmap and we couldn't know the names in advance. 

+1 for an open discussion.

Thanks @jmazzitelli.

roles/default/kiali-deploy/templates/kubernetes/role.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
   labels: {{ kiali_resource_metadata_labels }}
 rules:
 - apiGroups: [""]
-  resources:
+  resources:
```
jmazzitelli, line 11: revert this - just looks like trailing spaces were inserted here

roles/default/kiali-deploy/templates/openshift/role.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
   labels: {{ kiali_resource_metadata_labels }}
 rules:
 - apiGroups: [""]
-  resources:
+  resources:
```
jmazzitelli, line 11: revert this - just looks like trailing spaces were inserted here

jmazzitelli (changes requested):

leandroberetta:
@jmazzitelli I updated the code according to [this](https://github.com/kiali/kiali-operator/pull/395#issuecomment-896818860) conversation.

```yaml
spec:
  kiali_feature_flags:
    certificates_information_indicators:
      enabled: true
      secrets:
      - cacerts
      - istio-ca-secret
```

molecule/read-certs-secrets-test/molecule.yml
```diff
@@ -0,0 +1,46 @@
...
+    defaults:
+      callback_whitelist: junit
+  playbooks:
+    destroy: ../default/destroy.yml
+    prepare: ../default/prepare.yml
+    cleanup: ../default/cleanup.yml
+  inventory:
+    group_vars:
+      all:
+        cr_file_path: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') }}/molecule/jaeger-test/kiali-cr.yaml"
```
jmazzitelli, line 22:
```suggestion
        cr_file_path: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') }}/molecule/read-certs-secrets-test/kiali-cr.yaml"
```

roles/default/kiali-deploy/templates/openshift/role.yaml
```diff
@@ -97,4 +97,14 @@ rules:
...
+- apiGroups: [""]
+  resourceNames:
+  - cacerts
+  - istio-ca-secret
+  resources:
+  - secrets
+  verbs:
+  - create
+  - list
+  - watch
```
jmazzitelli, line 109: Not sure why this is here. Is the Kiali Server actually creating these secrets?

jmazzitelli (changes requested):

roles/default/kiali-deploy/templates/openshift/role.yaml
jmazzitelli: Note - there isn't any analogous change to the server helm-chart template - I assume this needs to be reverted. I don't think this should be here.

molecule/read-certs-secrets-test/converge.yml
```diff
@@ -0,0 +1,46 @@
...
+    vars:
+      get_verb: "get"
+      list_verb: "list"
+      watch_verb: "watch"
+    assert:
+      that: 
+      - cert_role is defined
+      - get_verb in cert_role.verbs
+      - list_verb in cert_role.verbs
+      - watch_verb in cert_role.verbs
```
jmazzitelli, line 46: I would also confirm that those are the ONLY 3 verbs (ensures no one accidentally gives "patch" or "delete", for example):
```suggestion
      - watch_verb in cert_role.verbs
      - cert_role.verbs | length == 3
```

jmazzitelli (changes requested):
as per our [developer checklist here](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission-that-applies-to-resources-only-in-the-control-plane-namespace), add these permissions to the OLM metadata.

Add `get` verb above the existing `list` verb in the three OLM CSVs:

* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L277
* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L277
* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L318

I still have some reservations here. I'm thinking about setting `resourceNames` here in the CSVs with the 2 secrets we expect to need access to (`cacerts`, `istio-ca-secret`). But the above addition of `get` verb needs to happen regardless.

jmazzitelli:
Here's the last thing I'm thinking about. With the CSVs (which control the permissions given to the operator when installed via OLM):
* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L277
* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L277
*https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L318

Adding "get" would give the operator (and anyone who gets a hold of the operator SA) the ability to read any secret anywhere in the cluster. I think we should consider restricting this to only supporting "cacerts" and "istio-ca-secret" for now. I assume those are going to be the out-of-box secret names that no one will really want to change. So it would be:

```
        #### THIS ALREADY EXISTS
        - apiGroups: [""]
          resources:
          - secrets
          verbs:
          - create
          - list
          - watch
        #### WE ADD THIS
        - apiGroups: [""]
          resourceNames:
          - cacerts
          - istio-ca-secret
          resources:
          - secrets
          verbs:
          - get
```

This will only really work if those two secret names will never change in any control plane in the cluster. If these names are configurable by the user (i.e. if they can add their own custom names in the  Kiali CR) then this really won't work.

But I'm paranoid of the fact we are creating a service account that can read all secrets in the cluster. So if possible we should restrict this further.

leandroberetta:
> Here's the last thing I'm thinking about. With the CSVs (which control the permissions given to the operator when installed via OLM):
> 
> * https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L277
> * https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L277
> * https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L318
> 
> Adding "get" would give the operator (and anyone who gets a hold of the operator SA) the ability to read any secret anywhere in the cluster. I think we should consider restricting this to only supporting "cacerts" and "istio-ca-secret" for now. I assume those are going to be the out-of-box secret names that no one will really want to change. So it would be:
> 
> ```
>         #### THIS ALREADY EXISTS
>         - apiGroups: [""]
>           resources:
>           - secrets
>           verbs:
>           - create
>           - list
>           - watch
>         #### WE ADD THIS
>         - apiGroups: [""]
>           resourceNames:
>           - cacerts
>           - istio-ca-secret
>           resources:
>           - secrets
>           verbs:
>           - get
> ```
> 
> This will only really work if those two secret names will never change in any control plane in the cluster. If these names are configurable by the user (i.e. if they can add their own custom names in the Kiali CR) then this really won't work.
> 
> But I'm paranoid of the fact we are creating a service account that can read all secrets in the cluster. So if possible we should restrict this further.

I'm ok with this situation, because at least we are covering the most common cases. 

Remember that we have 3 situations:

- Istio generates certificates with its CA, secret always will be called "istio-ca-secret"
- Users inject their CA, secret needs to be called "cacerts"
- Chiron is configured, in this case, the Kubernetes CA is used, for any DNS record and in this scenario, we could'nt know the secrets names in advances, and can be more than one

For the last scenario, we wouldn't be able to do anything, but if the users want, they can grant this permissions to the Kiali SA extending the allowed list of secrets it can read. We have the information warning in the UI about this situation, and a FAQ entry will be handy.

With this last comment, the work done respecting the field secrets in certificates_information_indicators, makes sense? because the operator won't be able to grant permission for other secrets the users may configure.

jmazzitelli:
Here's how I see it. First, assume we add "cacerts" and "istio-ca-secret" to the OPERATOR role (so the operator can only grant permissions for the server to see at most those 2).

Yes, it is a little weird to also have `certificates_information_indicators.secrets` with the same list of secret names. Because a user could then think "I can add my own secrets here in the Kiali CR and things will work". But they won't work because the operator will not (by default) be allowed to grant the server permissions to read those extra secrets.

But here is why I think we should do this:

1) That feature of wanting the server to read additional custom secrets is not the default use-case (as far as I understand it). In fact, the user will probably never have to set `certificates_information_indicators.secrets` in their Kiali CR (the default set they get out-of-box ("cacerts" and "istio-ca-certs") is probably all they ever need) They at most will set `certificates_information_indicators.enabled` to false to turn that off. In other words, it isn't like we are making things harder on the user - they probably won't even know this setting exists let alone be forced to set it.
2) But at least we give the ability to add more secrets via `certificates_information_indicators.secrets`. If for some reason a Kiali install needs additional secrets in the list (the chiron use-case you mention), then they are free to update the operator role to then allow users to update their Kiali CRs to add additional secrets. So at least we provide a mechanism for the user to do this.
3) But (and here's why I think we want this) we now will have a way to limit the operator service account to only be able to see a fixed set of secrets. We do not really want to blindly give the operator access to every secret in the cluster -- this is major security concern. Forcing the user to update the operator role to add new secrets is "a good thing" because it will force the user to explicitly do something to open up access. Otherwise, if we do not do anything to restrict secret access, then I predict the vast majority of our users will install the operator and never know they are creating a service account in their cluster that gives read access to all secrets, and when some do find out they will most likely immediately want a way to alter that (especially those users who have very strict security requirements).

We will need to add an FAQ to tell users how they can update the operator that was installed via OLM how they can update their operator role. If the user used a helm chart to install the operator, it will be easy - they just update the "secretReader" list setting. For OLM, it is more difficult. There is nothing today that allows you to configure the operator role when installing via OLM. So users will need to manually update the role themselves (most like via a "oc patch" command - we will need to document that command for them in an FAQ).

jmazzitelli:
I think the final thing to do is merge https://github.com/leandroberetta/kiali-operator/pull/3

That will lock things down a bit more in the operator when installed via OLM. We will need to add an FAQ to tell users what they need to do if the operator needs more access to other secrets (like the chiron secrets). See the bottom of my [previous comment](https://github.com/kiali/kiali-operator/pull/378#issuecomment-901200544) on that.

jmazzitelli (changes requested):
merge https://github.com/leandroberetta/kiali-operator/pull/3

jmazzitelli:
I tested with all defaults, then with `--set "secretReader={cacerts}"` and then with `--set "secretReader="`. 

All three tests produced results that looked correct via `curl http://localhost:20001/kiali/api/istio/certs` ("accessible" field was "true" and "false" at appropriate times). I'm not sure what certs should be returned but I was seeing results.

jmazzitelli:
```
$ kubectl get kiali kiali -n kiali-operator -o yaml
...
  kiali_feature_flags:
    certificates_information_indicators:
      enabled: false
...
```

Results in:

```
$ curl -s http://localhost:20001/kiali/api/istio/certs | jq
[
  {
    "secretName": "cacerts",
    "secretNamespace": "istio-system",
    "dnsNames": null,
    "issuer": "",
    "notBefore": "0001-01-01T00:00:00Z",
    "notAfter": "0001-01-01T00:00:00Z",
    "error": "",
    "accessible": false
  }
]
```

Not sure if that is correct or not. Just documenting it here.

jmazzitelli:
with all defaults (operator and kiali cr) then all permissions are granted and the feature is enabled. This results:

```
$ curl -s http://localhost:20001/kiali/api/istio/certs | jq
[
  {
    "secretName": "istio-ca-secret",
    "secretNamespace": "istio-system",
    "dnsNames": null,
    "issuer": "O=cluster.local",
    "notBefore": "2021-08-18T08:04:32Z",
    "notAfter": "2031-08-16T08:04:32Z",
    "error": "",
    "accessible": true
  }
]
```

Not sure why this cert name is different.

jmazzitelli:
using `--set "secretReader={cacerts}"` and with this in kiali cr:
```
  kiali_feature_flags:
    certificates_information_indicators:
      enabled: true
      secrets:
      - cacerts
```

this results:

```
$ curl -s http://localhost:20001/kiali/api/istio/certs | jq
[
  {
    "secretName": "istio-ca-secret",
    "secretNamespace": "istio-system",
    "dnsNames": null,
    "issuer": "",
    "notBefore": "0001-01-01T00:00:00Z",
    "notAfter": "0001-01-01T00:00:00Z",
    "error": "",
    "accessible": false
  }
]
```

jmazzitelli:
I also played with the clustering.enabled flag (setting it to false) while also toggling this secret enable flag between true and false. I looked at the kiali-controlplane role and it all looks correct for the different scenarios and the server can be installed and runs. So I don't see anything wrong there.

jmazzitelli (approved):
after a bunch of testing, the operator is working properly.

I have a concern about the server-side functionality (see this [comment](https://github.com/kiali/kiali-operator/pull/378#issuecomment-902039512) and below) but that doesn't involve the operator code - this is all doing what it should be doing.

jmazzitelli (changes requested):
use 1.40 metadata files now.

manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
```diff
@@ -276,6 +276,14 @@ spec:
...
           - list
           - watch
+        - apiGroups: [""]
+          resourceNames:
+          - cacerts
+          - istio-ca-secret
+          resources:
+          - secrets
+          verbs:
+          - get
```
jmazzitelli, line 286: revert this change from this 1.39 file but make the identical change to the 1.40 file.

manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
```diff
@@ -276,6 +276,14 @@ spec:
...
           - list
           - watch
+        - apiGroups: [""]
+          resourceNames:
+          - cacerts
+          - istio-ca-secret
+          resources:
+          - secrets
+          verbs:
+          - get
```
jmazzitelli, line 286: revert this change from this 1.39 file but make the identical change to the 1.40 file.

jmazzitelli (changes requested):
more specifics as to what needs to change (make the changes to the 1.40 metadata files and revert the changes made here to the 1.39 files)

manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
leandroberetta: Done!.

manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml
leandroberetta: Done!.

jmazzitelli (approved):
lgtm

jmazzitelli (approved):
re-approve

PR #380, opened 2021-08-03
Grafana integration test

leandroberetta:
Fixes [#4195](https://github.com/kiali/kiali/issues/4195)

### DIFF
--- a/molecule/grafana-test/converge.yml
+++ b/molecule/grafana-test/converge.yml
@@ -1,0 +1,117 @@
+- name: Tests
+  hosts: localhost
+  connection: local
+  collections:
+  - kubernetes.core
+  vars:
+    custom_resource: "{{ lookup('template', cr_file_path) | from_yaml }}"
+  tasks:
+  - import_tasks: ../common/tasks.yml
+  - import_tasks: ../asserts/pod_asserts.yml
+  - import_tasks: ../common/wait_for_kiali_running.yml
+  
+  - name: Get statuses from Istio components
+    uri:
+      url: "{{ kiali_base_url }}/api/istio/status"
+      validate_certs: no
+      return_content: yes      
+    register: status_response
+  
+  - name: "Results: /api/istio/status"
+    debug:
+      msg: "{{ status_response.json }}"  
+
+  # IstioStatus API returns a statuses list of external components (Grafana is one of them)
+  # If Grafana is in the list, it means that there is an error (Unreacheable)
+  - name: Check if Grafana status is present
+    set_fact:
+      grafana_error: "{{ item.status }}"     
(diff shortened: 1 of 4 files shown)


molecule/grafana-test/converge.yml
```diff
@@ -0,0 +1,117 @@
...
+      - grafana_error is defined
+        
+  - import_tasks: ../common/wait_for_kiali_cr_changes.yml
+
+  # Update Grafana Health URL to used an alternative URL for health checking 
+  - name: Update Grafana Health URL
+    vars:
+      current_kiali_cr: "{{ kiali_cr_list.resources[0] }}"
+    set_fact:
+      new_kiali_cr: "{{ current_kiali_cr | combine({'spec': {'external_services': {'grafana': {'health_check_url': 'https://grafana.{{ istio.control_plane_namespace }}.svc'}}}}, recursive=True) }}"
```
jmazzitelli, line 86: You are already inside an evaluation block (`{{ }}') so you don't need to embed a variable in another set of double-braces. You are also inside a single-quoted string so you have to build the string with `+` operator. Try this:

```suggestion
      new_kiali_cr: "{{ current_kiali_cr | combine({'spec': {'external_services': {'grafana': {'health_check_url': 'https://grafana.' + istio.control_plane_namespace + '.svc'}}}}, recursive=True) }}"
```

molecule/grafana-test/converge.yml
```diff
@@ -0,0 +1,127 @@
...
+    set_fact:
+      new_kiali_cr: "{{ current_kiali_cr | combine({'spec': {'external_services': {'grafana': {'health_check_url': 'http://grafana.' +  istio.control_plane_namespace + '.svc:3000'}}}}, recursive=True) }}"
+    when:
+    - is_openshift == False
+
+  - name: Update Grafana Health URL (OpenShift)
+    vars:
+      current_kiali_cr: "{{ kiali_cr_list.resources[0] }}"
+    set_fact:
+      new_kiali_cr: "{{ current_kiali_cr | combine({'spec': {'external_services': {'grafana': {'health_check_url': 'https://grafana.' +  istio.control_plane_namespace + '.svc:3000'}}}}, recursive=True) }}"
```
jmazzitelli, line 94: I don't see a difference between these two tasks that update the health url (one for kube , one for OS, but they both look the same). What am I missing?

jmazzitelli: Ah... yes... as you see, I do NOT have eagle eyes :)

jmazzitelli (approved):
tested on minikube/upstream istio 1.11 and openshift 4.8/OSSM 2.1 - all pass.

PR #383, opened 2021-08-04
Added option to skip DestinationRule subset validations in Kiali.

hhovsepy:
POC for the RFE https://github.com/kiali/kiali/issues/4218

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -724,6 +724,10 @@ spec:
 #    ---
 #    istio_upgrade_action: false
 #
+# When there is no VirtualService referencing to a subset in a DestinationRule, skip that subset's validation.
+#    ---
+#    skip_subset_validation: false
+#
 # Default settings for the UI. These defaults apply to all users.
 #    ---
 #    ui_defaults:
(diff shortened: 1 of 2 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -724,6 +724,10 @@ spec:
 #    ---
 #    istio_upgrade_action: false
 #
+# When there is no VirtualService referencing to a subset in a DestinationRule, skip that subset's validation.
+#    ---
+#    skip_subset_validation: false
+#
```
jmazzitelli, line 730: We have a section here "validations" that is perfect for this kind of thing. Any settings that have to deal with validations (like this one) should go under there. So move this under `spec.kiali_feature_flags.validations`

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -210,6 +210,7 @@ kiali_defaults:
   kiali_feature_flags:
     istio_injection_action: true
     istio_upgrade_action: false
+    skip_subset_validation: false
```
jmazzitelli, line 213: move this down to the `validations` section - under line 235.

I also think you can rename it since it will be under "validations" so it will be assumed it is related to validations. So you can call this "skip_subsets" or something.

jmazzitelli (changes requested):

hhovsepy:
Skipping subset validation in case of not referenced VS will be always on, so closing this PR. Thanks for the reviewing.

PR #384, opened 2021-08-04
use_grpc by default. health_check_url removal from tracing

xeviknal:
needs https://github.com/kiali/kiali/pull/4249

1. `use_grpc` is by default true. It was before, now it is explicitly set to true.
2. removing the `health_check_url` from tracing config. We use actual jaeger API to check availability. So no need to add an extra url.

I assume that we don't need to change any yaml from the helm, am I right @jmazzitelli ?

### DIFF
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -186,12 +186,11 @@ kiali_defaults:
         use_kiali_token: false
         username: ""
       enabled: true
-      health_check_url: ""
       in_cluster_url: ""
       is_core: false
       namespace_selector: true
       url: ""
-      #use_grpc:
+      use_grpc: true
       whitelist_istio_system: ["jaeger-query", "istio-ingressgateway"]
 
   health_config:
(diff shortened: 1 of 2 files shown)


roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -186,12 +186,11 @@ kiali_defaults:
         use_kiali_token: false
         username: ""
       enabled: true
-      health_check_url: ""
       in_cluster_url: ""
       is_core: false
       namespace_selector: true
       url: ""
-      #use_grpc:
+      use_grpc: true
```
jmazzitelli, line 193: In the comment above on line 641 it says,"#If not specified, it will be false if deployed within a Maistra/OSSM+OpenShift environment, true otherwise."

I think this is still true but I would like to confirm

jmazzitelli: I answered my own question - the comment is correct. See: https://github.com/kiali/kiali-operator/blob/v1.38.0/roles/default/kiali-deploy/tasks/main.yml#L324-L326

jmazzitelli: Because this is now defaulting to true, we should update the kiali_cr.yaml example file to reflect this - see line 658 of kiali_cr.yaml

jmazzitelli (changes requested):
https://github.com/kiali/kiali-operator/pull/384/files#diff-b05613685dff7d8a5eb70946010acb19192f6f9edc83d292af21db46aa9be89dR658

line 658 - the default is now true, so the comment should show that.

jmazzitelli:
> I assume that we don't need to change any yaml from the helm, am I right @jmazzitelli ?

correct - no changes needed for the helm charts

jmazzitelli:
once merged, we will backport to `v1.38` branch

jmazzitelli (approved):
i'm going to approve so we can merge

PR #395, opened 2021-08-10
Add feature flag for clustering/federation features

israel-hdez:
For now, this flag is only allowing or preventing creation of a role and a rolebinding that grants privileges for reading some secrets in the Istio namespace. There is no effect on Kiali server side, because it is failing gracefully. In case it's needed later, this flag is being passed to Kiali in its ConfigMap.

Default value is `true` for backwards compatibility.

Related kiali/kiali#4147.

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -716,6 +716,10 @@ spec:
 #  ---
 #  kiali_feature_flags:
 #
+# Flag to indicate Kiali to enable/disable clustering and federation related features.
+#    ---
+#    clustering: true
+#
 # Flag to indicate Kiali to enable/disable an Action to label a namespace for automatic Istio Sidecar injection.
 #    ---
 #    istio_injection_action: true
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -207,6 +207,7 @@ kiali_defaults:
     version_label_name: "version"
 
   kiali_feature_flags:
+    clustering: true
     istio_injection_action: true
     istio_upgrade_action: false
     ui_defaults:
--- a/roles/default/kiali-deploy/tasks/kubernetes/k8s-main.yml
+++ b/roles/default/kiali-deploy/tasks/kubernetes/k8s-main.yml
@@ -4,6 +4,29 @@
   when:
   - is_k8s == True
 
+- name: Process kiali-controlplane Role on Kubernetes
+  k8s:
+    state: "{{ 'present' if kiali_vars.kiali_feature_flags.clustering|bool == True else 'absent' }}"
+    definition: "{{ lookup('template', 'templates/kubernetes/role-controlplane.yaml') }}"
+  register: process_resource_result
+  until:
+  - process_resource_result.error is not defined
+  - process_resource_result.result is defined
+  - process_resource_result.result.metadata is defined
+  retries: 6
+
+- name: Process kiali-controlplane RoleBinding on Kubernetes
+  k8s:
+    state: "{{ 'present' if kiali_vars.kiali_feature_flags.clustering|bool == True else 'absent' }}"
+    definition: "{{ lookup('template', 'templates/kubernetes/rolebinding-controlplane.yaml') }}"
+  register: process_resource_result
+  until:
+  - process_resource_result.error is not defined
+  - process_resource_result.result is defined
+  - process_resource_result.result.metadata is defined
+  retries: 6
+  delay: 10
+
 - name: Create Kiali objects on Kubernetes
   include_tasks: process-resource.yml
   vars:
@@ -13,9 +36,7 @@
   - serviceaccount
   - configmap
   - "{{ 'role-viewer' if kiali_vars.deployment.view_only_mode|bool == True else 'role' }}"
-  - role-controlplane
   - rolebinding
-  - rolebinding-controlplane
   - deployment
   - service
   - "{{ 'hpa' if kiali_vars.deployment.hpa.spec | length > 0 else '' }}"
--- a/roles/default/kiali-deploy/tasks/openshift/os-main.yml
+++ b/roles/default/kiali-deploy/tasks/openshift/os-main.yml
@@ -4,6 +4,29 @@
   when:
   - is_openshift == True
 
+- name: Process kiali-controlplane Role on OpenShift
+  k8s:
+    state: "{{ 'present' if kiali_vars.kiali_feature_flags.clustering|bool == True else 'absent' }}"
+    definition: "{{ lookup('template', 'templates/openshift/role-controlplane.yaml') }}"
+  register: process_resource_result
+  until:
+  - process_resource_result.error is not defined
+  - process_resource_result.result is defined
+  - process_resource_result.result.metadata is defined
+  retries: 6
+
+- name: Process kiali-controlplane RoleBinding on OpenShift
+  k8s:
+    state: "{{ 'present' if kiali_vars.kiali_feature_flags.clustering|bool == True else 'absent' }}"
+    definition: "{{ lookup('template', 'templates/openshift/rolebinding-controlplane.yaml') }}"
+  register: process_resource_result
+  until:
+  - process_resource_result.error is not defined
+  - process_resource_result.result is defined
+  - process_resource_result.result.metadata is defined
+  retries: 6
+  delay: 10
+
 - name: Create Kiali objects on OpenShift
   include_tasks: process-resource.yml
   vars:
@@ -14,9 +37,7 @@
   - configmap
   - cabundle
   - "{{ 'role-viewer' if kiali_vars.deployment.view_only_mode|bool == True else 'role' }}"
-  - role-controlplane
   - rolebinding
-  - rolebinding-controlplane
   - deployment
   - service
   - "{{ 'hpa' if kiali_vars.deployment.hpa.spec | length > 0 else '' }}"


deploy/kiali/kiali_cr.yaml
```diff
@@ -716,6 +716,10 @@ spec:
 #  ---
 #  kiali_feature_flags:
 #
+# Flag to indicate Kiali to enable/disable clustering and federation related features.
+#    ---
+#    clustering: true
```
jmazzitelli, line 721: I have two comments here.

1. Is this a good name for this? The name "clustering" seems generic. It may be the proper name for this feature, but the word "clustering" doesn't bring to my mind a specific feature here. In our [Features](https://kiali.io/documentation/latest/features/#_advanced_mesh_deployment_and_multi_cluster_support) doc page I see words like "multi-cluster" and "multi-network" and "mesh". Should this be called "multi_clustering"? I just want to make sure we think about this name before committing to support it - if "clustering" is the correct name for this feature, leave it.

2. This multi-cluster support ("clustering") is becoming a major feature in Kiali. I think there are going to be many more pieces of this large feature set that we may want to be configurable. Let's prepare for this future by having this "clustering" be a parent to a potential group of settings (in the same way we have ui_defaults here):
```
kiali_feature_flags:
  clustering:
    enabled: true
```

This way in the future we can support additional settings that go under the clustering heading.

Otherwise, I can easily see this section get out of hand with a bunch of independent and separate clustering settings - we do not want to have something like this happen:
```
kiali_feature_flags:
  clustering: true
  clustering_cache_size: 5
  clustering_secrets_edit: foo
  clustering_allow_rbac: true
```

It would be better for everything to go under the clustering heading:
```
kiali_feature_flags:
  clustering:
    enabled: true
    cache_size: 5
    secrets_edit: foo
    allow_rbac: true
```

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -207,6 +207,7 @@ kiali_defaults:
     version_label_name: "version"
 
   kiali_feature_flags:
+    clustering: true
```
jmazzitelli, line 210: I would prefer this:
```suggestion
    clustering:
      enabled: true
```

jmazzitelli (changes requested):

jmazzitelli:
@leandroberetta @israel-hdez This has the potential to affect what @leandroberetta is doing wrt displaying certificate information in the UI. @leandroberetta is adding permissions to this controlplane role (but he is adding permissions related to secrets too).

In other words, this control plane role is not necessary all about clustering - so I am hesitant to think we can just blindly not install this control plane role entirely when clustering.enabled is false. You may have to just not install parts of the role, but leave some of the role intact.

BTW: side note @leandroberetta  - read the [issue](https://github.com/kiali/kiali/issues/4147) related to this PR - you'll see what I was talking about last week when I said some people are paranoid about having our stuff be given any kind of permissions for secrets :) 

Because the issue this PR is addressing talks about not wanting to grant privileges to read secrets, I am sure the permissions @leandroberetta is introducing also needs to be removed as well. So now I am questioning if this new section should be called "clustering" at all - it sounds like we need a new setting in `spec.deployment` (since those settings involve the deployment resources) - something like `spec.deployment.minimal_secret_permissions: true` or something like that.

In other words, we need to think more deeply about how we want to implement this - I'm thinking a "clustering" section in kiali_feature_flags isn't the proper way to do this since it involves more than just clustering feature.

leandroberetta:
> In other words, this control plane role is not necessary all about clustering - so I am hesitant to think we can just blindly not install this control plane role entirely when clustering.enabled is false. You may have to just not install parts of the role, but leave some of the role intact.
> 

+1

> BTW: side note @leandroberetta - read the [issue](https://github.com/kiali/kiali/issues/4147) related to this PR - you'll see what I was talking about last week when I said some people are paranoid about having our stuff be given any kind of permissions for secrets :)
> 

Yes, thanks for linking this, very useful feedback!.

> Because the issue this PR is addressing talks about not wanting to grant privileges to read secrets, I am sure the permissions @leandroberetta is introducing also needs to be removed as well. So now I am questioning if this new section should be called "clustering" at all - it sounds like we need a new setting in `spec.deployment` (since those settings involve the deployment resources) - something like `spec.deployment.minimal_secret_permissions: true` or something like that.
> 

+1 on this strategy, and maybe split this flag into two:

- minimal_secret_permissions for the secrets that the certificates issue needs to read (istio-ca-secret and cacerts)
- clustering_secrets for the actual permissions the role has (list all secrets)

jmazzitelli:
> In other words, we need to think more deeply about how we want to implement this - I'm thinking a "clustering" section in kiali_feature_flags isn't the proper way to do this since it involves more than just clustering feature.

As I think about this, I think perhaps we need a `spec.deployment.allowed_secrets` section. This could be a list of named control plane secrets that the Kiali Server is granted permission to get and/or list. This can be used to dynamically build the cluster role. Something like this maybe? (I'm just thinking out loud)

```yaml
spec:
  deployment:
    allowed_secrets:
    # here is a secret that kiali will be allowed to read and list
    - name: cacert
      can_read: true
    # here is a secret that kiali is only allowed to list but not read its content
    - name: cluster-secret
      can_read: false
```

leandroberetta:
> > In other words, we need to think more deeply about how we want to implement this - I'm thinking a "clustering" section in kiali_feature_flags isn't the proper way to do this since it involves more than just clustering feature.
> 
> As I think about this, I think perhaps we need a `spec.deployment.allowed_secrets` section. This could be a list of named control plane secrets that the Kiali Server is granted permission to get and/or list. This can be used to dynamically build the cluster role. Something like this maybe? (I'm just thinking out loud)
> 
> ```yaml
> spec:
>   deployment:
>     allowed_secrets:
>     # here is a secret that kiali will be allowed to read and list
>     - name: cacert
>       can_read: true
>     # here is a secret that kiali is only allowed to list but not read its content
>     - name: cluster-secret
>       can_read: false
> ```

I like this idea even more to avoid hardcoding the secrets I need but we also need a flag to list all secrets for the clustering feature.

jmazzitelli:
> * minimal_secret_permissions for the secrets that the certificates issue needs to read (istio-ca-secret and cacerts)
> * clustering_secrets for the actual permissions the role has (list all secrets)

Building off this and my previous suggestion ... maybe:

```
spec:
  deployment:
    control_plane_secret_access:
      list_only:
      - clustering_secret_foo
      - another_one
      read_only:
      - cacerts
      - istio-ca-secret
```

Or, if we just want to allow to list all secrets in control plane without naming them:

```
spec:
  deployment:
    control_plane_secret_access:
      list_all_allowed: true
      read_only:
      - cacerts
      - istio-ca-secret
```

israel-hdez:
For "clustering", I though the `feature_flags` was the most appropriate section to place this stuff, as it may control future features related to clustering.

I mean, that flag would control all clustering related features. For example, if some day in the future we choose that we need "agents" to properly support clustering capabilities, this flag would control to either trying to connect to those agents, or not try at all. Then, at the same time, roles and role bindings should also be adjusted for features to work properly. If we go the `allowed_secrets` way, granularity is too fine IMO.

In any case, conceptually, I think it's better to turn on/off "features" and letting the operator do the needed adjustments to Kubernetes resources (not neccesarily only secrets). This is because users will realize more clearly what is not going to work in Kiali, rather than giving low level options about what Kubernetes objects should be created :-/. 

I mean, a feature flag named something like `certificate_management_indicators: on/off` tells me more about what is not going to work in Kiali, rather than a low level `allowed_secrets` stanza. Also, in Kiali we can more precisely hide features with such feature flags, rather than passively trying to list secrets to find that there are no privileges and then posting an error to the message center :-/ (which makes me think that the user could be flooded by unneeded error messages).

jmazzitelli:
> For "clustering", I though the `feature_flags` was the most appropriate section to place this stuff, as it may control future features related to clustering.
> 
> I mean, that flag would control all clustering related features. For example, if some day in the future we choose that we need "agents" to properly support clustering capabilities, this flag would control to either trying to connect to those agents, or not try at all.

Yes, this is why I recommended having a group setting where all the clustering settings go under:

```
spec:
  kiali_feature_flags:
    clustering:
      enabled: true
      connect_to_agents: true
```

jmazzitelli:
> I mean, a feature flag named something like `certificate_management_indicators: on/off` tells me more about what is not going to work in Kiali, rather than a low level `allowed_secrets` stanza. Also, in Kiali we can more precisely hide features with such feature flags, rather than passively trying to list secrets to find that there are no privileges and then posting an error to the message center :-/ (which makes me think that the user could be flooded by unneeded error messages).

So do you suggest something like:

```
spec:
  kiali_feature_flags:
    clustering:
      enabled: true
    certificate_management_indicators:
      enabled: true
```

If we do that, we could still provide an optional set of secret names (with a well-known set of names as a default)

```
spec:
  kiali_feature_flags:
    clustering:
      enabled: true
    certificate_management_indicators:
      enabled: true
      secrets:
      - cacerts
      - istio-ca-cert
```

israel-hdez:
> So do you suggest something like:
> 
> ```
> spec:
>   kiali_feature_flags:
>     clustering:
>       enabled: true
>     certificate_management_indicators:
>       enabled: true
> ```
> 
> If we do that, we could still provide an optional set of secret names (with a well-known set of names as a default)
> 
> ```
> spec:
>   kiali_feature_flags:
>     clustering:
>       enabled: true
>     certificate_management_indicators:
>       enabled: true
>       secrets:
>       - cacerts
>       - istio-ca-cert
> ```

I would be more comfortable with this.

Listing secrets under that `certificate_management_indicators` is what I'm not very sure. Those secrets are Istio things, and we also have `external_services.istio` (see https://github.com/kiali/kiali-operator/blob/0796748bb6498dc6d3e1bb6351e4186014789e0e/deploy/kiali/kiali_cr.yaml#L555-L578) for those purposes. However, I find confusing detaching settings :-/

jmazzitelli:
@israel-hdez OK, so for this PR, just implement that new clustering section:

```
spec:
  kiali_features_flags:
    clustering:
      enabled
```

And @leandroberetta can implement that other section in his PR.

But because both of those are going to use the controlplane role, this PR cannot eliminate the role entirely. Just have this PR update the role with the secret permission needed. Then @leandroberetta will do the same with his PR (he'll need to add his secret perms).

Of course, the operator will need all of these perms set. We'll have to figure out later if/how we turn on or off those perms in the operator itself. For now, just give the operator all these permissions

deploy/kiali/kiali_cr.yaml
israel-hdez: Well, "clustering" is the term that has been used in some calls.

TBH, I don't know why that term is being used. "Clustering" is usually used in machine learning. By googling, I see this definition: "Clustering is the task of dividing the population or data points into a number of groups such that data points in the same groups are more similar to other data points in the same group than those in other groups." Although the definition may give a close idea to multi-cluster, I'm not sure if it's really OK to use that term.

About (2), I'm working on it.

roles/default/kiali-deploy/tasks/kubernetes/k8s-main.yml
```diff
@@ -13,9 +36,7 @@
   - serviceaccount
   - configmap
   - "{{ 'role-viewer' if kiali_vars.deployment.view_only_mode|bool == True else 'role' }}"
-  - role-controlplane
   - rolebinding
-  - rolebinding-controlplane
```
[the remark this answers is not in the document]
israel-hdez: Well, the idea was to either create or delete the role and rolebinding, depending on the clustering flag. By ommiting from this list, deletion of the role does not happen if the CR is edited and `clustering` is switched off.

But well.. Now that the role is only going to include or omit privileges over the secrets, this becomes unnecesary.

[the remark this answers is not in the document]
jmazzitelli: > the idea was to either create or delete the role and rolebinding, depending on the clustering flag. By ommiting from this list, deletion of the role does not happen if the CR is edited and `clustering` is switched off.

Ahh.. good point!

> Now that the role is only going to include or omit privileges over the secrets, this becomes unnecesary.

Another good point :)

israel-hdez:
I'm starting to find that `kiali_feature_flags` a little confusing.

When we should use it?
If `clustering` has the potential to grow in settings, perhaps should it be a first-level citizen (like `auth`)?

jmazzitelli:
> When we should use it?
> If `clustering` has the potential to grow in settings, perhaps should it be a first-level citizen (like `auth`)?

Hmmm.... possible. Is this a large enough feature to warrant that? That may be. I don't know the answer right now.

israel-hdez:
> Hmmm.... possible. Is this a large enough feature to warrant that? That may be. I don't know the answer right now.

Yes, I don't know either. It's too early to know.
Also, I have the thought that everything inside that `kiali_feature_flags` should be pulled one level up (specially that `ui_defaults`), because I see the KialiCR as configuration for both installation and runtime.

Meh, anyway, let's let it be.

jmazzitelli (approved):
very nice.

PR #398, opened 2021-08-19
[molecule] when setting the health check url, just set it to the known good url

jmazzitelli:
This sets the health check url to the one that was originally in the in-cluster-url setting.

The way this was originally, the in-cluster-url remained set to that "wrong" url and that must have caused this to fail on OpenShift for some reason.

This passes on OpenShift and minikube now.

fixes: https://github.com/kiali/kiali/issues/4289


leandroberetta (approved):
LGTM

PR #420, opened 2021-09-10
ansible setting is deprecated, use the new one

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4338

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -9,4 +9,4 @@ COPY requirements.yml ${HOME}/requirements.yml
 RUN ansible-galaxy collection install -r ${HOME}/requirements.yml \
  && chmod -R ug+rwx ${HOME}/.ansible
 
-RUN cp /etc/ansible/ansible.cfg ${HOME}/ansible-profiler.cfg && echo "callback_whitelist = profile_tasks" >> ${HOME}/ansible-profiler.cfg
+RUN cp /etc/ansible/ansible.cfg ${HOME}/ansible-profiler.cfg && echo "callback_enabled = profile_tasks" >> ${HOME}/ansible-profiler.cfg && echo "callback_whitelist = profile_tasks" >> ${HOME}/ansible-profiler.cfg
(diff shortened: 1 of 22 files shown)


build/Dockerfile
```diff
@@ -9,4 +9,4 @@ COPY requirements.yml ${HOME}/requirements.yml
 RUN ansible-galaxy collection install -r ${HOME}/requirements.yml \
  && chmod -R ug+rwx ${HOME}/.ansible
 
-RUN cp /etc/ansible/ansible.cfg ${HOME}/ansible-profiler.cfg && echo "callback_whitelist = profile_tasks" >> ${HOME}/ansible-profiler.cfg
+RUN cp /etc/ansible/ansible.cfg ${HOME}/ansible-profiler.cfg && echo "callback_enabled = profile_tasks" >> ${HOME}/ansible-profiler.cfg && echo "callback_whitelist = profile_tasks" >> ${HOME}/ansible-profiler.cfg
```
jmazzitelli, line 12: I just want to comment on this. I left the callback_whitelist in here because the current base image of the ansible operator in use by Kiali today does not support the new callback_enabled setting yet. I put the callback_enabled in here now so it will be there when we do upgrade (having both won't hurt anything). It makes sense to do this as part of this PR. We can remove the callback_whitelist entry later when we no longer need it (i.e. when we are using an ansible base image that supports it).

jmazzitelli:
To test this - run a molecule test with `make` and the env var `MOLECULE_OPERATOR_PROFILER_ENABLED=true`:

```
make -e MOLECULE_OPERATOR_PROFILER_ENABLED=true -e MOLECULE_SCENARIO="default" molecule-test)
```

Watch the operator logs and see that the profiler report spits out after the operator reconcilation is done (`kubectl logs -n kiali-operator -l app=kiali-operator`).

The same with the operator deployed normally and using env var `OPERATOR_PROFILER_ENABLED=true`:

```
make -e OPERATOR_PROFILER_ENABLED=true operator-create kiali-create
```

and look at the operator logs when the reconciliation is done - the profiler report should spit out at the end.

leandroberetta (approved):
LGTM, no signs of deprecation warnings after the end of the reconciliation.

jmazzitelli:
FWIW: this PR is wrong. It used `callback_enabled` when it should be plural - `callbacks_enabled`. I'm only 2 years late to this party. I'll fix this in a new PR.

PR #421, opened 2021-09-13
add new custom headers setting to prometheus config.

jmazzitelli:
side note: this also makes sure the prom config is the same for custom_dashboards.prometheus since that follows the identical config schema as the external_services.prometheus config.

part of: https://github.com/kiali/kiali/issues/4323


jshaughn (approved):
LOK

pbajjuri20 (approved):

PR #426, opened 2021-09-22
new olm metadata for v1.41.0

jmazzitelli:


jshaughn (approved):
LOK.  We don't need that openshift version annotation that we added in the other PR, right?

jmazzitelli:
> LOK. We don't need that openshift version annotation that we added in the other PR, right?

correct - that was only for our older versions that used CRD v1beta1 - we are no longer on that version of CRD so no need to restrict our operator to only install on 4.8 and earlier. The new stuff can run on OpenShift 4.9+

PR #427, opened 2021-09-23
Kiali Istio dashboards incompatible with thanos-query

aljesusg:
Fix https://github.com/kiali/kiali/issues/4303
Require: https://github.com/kiali/kiali/pull/4386

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -623,6 +623,10 @@ spec:
 #      custom_headers: {}
 #      health_check_url: ""
 #      is_core: true
+#      thanos_proxy:
+#        enabled: false
+#        retention_period: "7d"
+#        scrape_interval: "30s"
 #      url: ""
 #
 # **Tracing-specific settings:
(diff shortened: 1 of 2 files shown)


deploy/kiali/kiali_cr.yaml
```diff
@@ -623,6 +623,10 @@ spec:
 #      custom_headers: {}
 #      health_check_url: ""
 #      is_core: true
+#      thanos_proxy:
+#        enabled: false
+#        retention_period: "7d"
+#        scrape_interval: "30s"
```
jmazzitelli, line 629: you should explain these settings above -- around line 607. Just add a comment up there to briefly talk about these so people know what they are for.

jmazzitelli (changes requested):

deploy/kiali/kiali_cr.yaml
aljesusg: Yes you're right, let me do it

jshaughn:
@aljesusg , why was this merged with no QE?

jshaughn:
@aljesusg Your logic makes sense but I think QE should make the decision as to whether the community testing is sufficient.  They may want to perform some sort of regression tests, even if there is no need to test the feature.  I'll assign @mattmahoneyrh to take a look and make the decision.

aljesusg:
You're right. My apologies for my fault. It's not going to happen again. Thanks

PR #441, opened 2021-11-01
Add root namespace property into Kiali config map

leandroberetta:
This PR adds a new property for Istio, the root namespace (https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/).

If it is not configured it defaults to the namespace where Kiali is installed (usually istio-system), same logic from istio_namespace property.


xeviknal (approved):
LGTM!

jmazzitelli (approved):
lgtm

PR #449, opened 2021-11-15
implement some best practices

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4485

### DIFF
--- a/deploy/kiali/kiali_cr.yaml
+++ b/deploy/kiali/kiali_cr.yaml
@@ -308,6 +308,11 @@ spec:
 #    ---
 #    ingress:
 #
+# Additional labels to add to the Ingress (or Route if OpenShift). These are added to the
+# labels that are created by default; these do not override the default labels.
+#      ---
+#      additional_labels: {}
+#
 # If class_name is a non-empty string, it will be used as the spec.ingressClassName
 # in the created Kubernetes Ingress resource. This setting is ignored if on OpenShift.
 # This is also ignored if override_yaml.spec is defined (i.e. you must define the
@@ -331,6 +336,8 @@ spec:
 # Note that only 'metadata.annotations' and 'spec' is valid and only they will
 # be used to override those same sections in the created resource. You can define
 # either one or both.
+# Note that `metadata.labels` is not allowed - you cannot override the labels; to add
+# labels to the default set of labels, use the "additional_labels" setting.
 # Example:
 #      override_yaml:
 #        metadata:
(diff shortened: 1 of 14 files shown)


jmazzitelli:
the first commit provides the ability to add custom labels to the Ingress/Route.

jmazzitelli:
second commit sets the resource requests. We might also consider setting resource limits, but so far that is not implemented.

jmazzitelli:
third commit sets the resource requests on the server. We might also consider setting resource limits, but so far that is not implemented.

jmazzitelli:
fourth commit adds memory limits to the server pod. We are going to opt not to create memory limits for the operator pod right now

jmazzitelli:
to test - in the Kiali CR set `spec.deployment.ingress.additional_labels.foo=bar` and (if on non-OpenShift/minikube/kind) set `spec.deployment.ingress.enabled=true`. When you install, check that the server pod has the label "foo: bar" and check its container to ensure it has `resources` defaults:
```yaml
resources:
  requests:
    cpu: 10m
    memory: 64Mi
  limits;
    memory: 1Gi
```

deploy/kiali/kiali_cr.yaml
```diff
@@ -331,6 +336,8 @@ spec:
 # Note that only 'metadata.annotations' and 'spec' is valid and only they will
 # be used to override those same sections in the created resource. You can define
 # either one or both.
+# Note that `metadata.labels` is not allowed - you cannot override the labels; to add
+# labels to the default set of labels, use the "additional_labels" setting.
```
nrfox, line 340: Why can't you set `metadata.labels` if you want to override the defaults completely?

jmazzitelli: We do not want users to override the defaults completely (which is why we have it separate like this). We still want our own labels still on the resource (things like "app: kiali") - we want these labels so they identify the resources Kiali creates. All our created resources have the same labels on them.

nrfox:
> second commit sets the resource requests on the operator. We might also consider setting resource limits, but so far that is not implemented.

This might just be my specific setup and how I am testing but I don't see the resource requirements on my operator pod:
```
metadata:                                                                                                                                                                                                          
  ...
  name: kiali-operator-7c56857485-mp8cr                                                                                                                                                                            
  namespace: kiali-operator                                                                                                                                                                                        
spec:                                                                                                                                                                                                              
  affinity: {}                                                                                                                                                                                                     
  containers:                                                                                                                                                                                                      
  - args:                                                                                                                                                                                                          
    - --zap-log-level=info                                                                                                                                                                                         
    - --leader-election-id=kiali-operator                                                                                                                                                                                                                                                                                                                                               
    resources: {}                                                                                                                                                                                                  
    ...
```

jmazzitelli:
> This might just be my specific setup and how I am testing but I don't see the resource requirements on my operator pod:

How did you install? You need the helm chart PR, too (because our makefile infrastructure uses the helm chart to install the operator): https://github.com/kiali/helm-charts/pull/118

nrfox:
> 

Bingo! That was it. Seeing the resource requirements on the operator pod now too.

nrfox (approved):
Tested on Kube 1.22

PR #451, opened 2021-11-16
fix 1.36 playbook to only hide kube- namespaces (not kube*)

jmazzitelli:
part of https://github.com/kiali/kiali/issues/4162


jshaughn (approved):
Approving change to master to be inline with the backport to v1.36 branch.

PR #452, opened 2021-11-17
Add rank to default find queries

nrfox:


roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -242,6 +242,8 @@ kiali_defaults:
           expression:  "! healthy"
         - description: "Find: unknown nodes"
           expression:  "name = unknown"
+        - description: "Find: nodes with the 2 highest ranks"
```
jmazzitelli, line 245: I still say this wording is going to create confusion among users. So I reserve my right to say "I told you so" in a few months when people start to ask "should I use < or > when testing rank?" :-p

nrfox: Fair enough, it is a little awkward to do `rank <= 2` to find the nodes with the top 2 ranks but what would be the alternative?

If ranking worked the opposite direction where higher ranks were a larger number you'd need to know the upper bound in order to find the "top ranking" nodes e.g. `rank > 99`.

jmazzitelli: I don't know a solution, I just know "rank <= 2" looks confusing to me :) And I don't know why - because it is true when you see top 10 lists people say things like.. "And the number 1 song this week is <drum roll>... !!!" So there is experience here where the highest rank == 1. This is why I am not requiring any changes here.

I don't know a good substitute - "priority"... "preference".... "precedence" ... "degree" ...."distinction"... "primacy"... "stratum" I'm just looking at a thesaurus :)

jshaughn: There is no good substitute.  A Rank of 1 is in general considered the best, or highest rank.  Feel free to say, "I told you so" :-) 

Having said that, one last shot at it:  "Find: nodes with the top 2 rankings".

jmazzitelli (approved):

nrfox:
@mattmahoneyrh brought up a good point while testing: https://github.com/kiali/kiali/pull/4504... @jmazzitelli why does the operator need these defaults at all when the server is defaulting these to the same values: https://github.com/kiali/kiali-operator/pull/452/files#diff-2ffd20932f9d933d8679373e95a0c458ade4fadb7a3e363a997b9ae9007b0dcdR238-R253? Can we just remove those from the `defaults/main.yml` and let the server apply its defaults so we don't need to maintain them in two different places?

jmazzitelli:
> why does the operator need these defaults at all

The convention has been to explicitly set the default values for everything in the operator defaults (except for those very few settings that truly need to default as "nil" - there are maybe 3 or 4 of these). This has always been the case for almost as long as the operator has been around. The main driver for that is if the operator needs to access those values (for whatever reason) the values are always set (thus avoiding the operator developer from having to code "is defined" checks everywhere - Ansible is not very tolerant of undefined values, and why our convention exists.)

Now, for these specific values in this PR, it is true the operator probably never needs to access them. So it probably doesn't need them in default/main.yml. But, if we stop doing this (always putting defaults in the operator) then NOW we truly will have to look at two different sources to find out what the defaults are (operator default/main.yaml and now the server Go code). Today, you just have to go to the operator default/main.yml for that source of truth. The defaults in the Go code were really only there (at least this was true in the beginning) for unit tests (when we run the unit tests, those defaults in the Go code were the ones that take effect).

Personally, I would prefer to keep it the way things are - otherwise, it will now get confusing when trying to track down where a value comes from ("what is the default for this value? Lemme look in the operator - oh, it isn't there. Well, let me see if the server Go code has the default set."). And, again, if for whatever reason the operator WILL need to look at these values, it should have them defined anyway. This convention just always puts them in there so we know they are always there if we ever need them.

nrfox:
> The main driver for that is if the operator needs to access those values (for whatever reason) the values are always set (thus avoiding the operator developer from having to code "is defined" checks everywhere - Ansible is not very tolerant of undefined values, and why our convention exists.)

That makes sense and seems fine to set the defaults in the operator if it makes life easier for the operator specifically to know that there's a value set there.

> The defaults in the Go code were really only there (at least this was true in the beginning) for unit tests (when we run the unit tests, those defaults in the Go code were the ones that take effect).

Another benefit is having the same defaults set for non-operator deployments like through helm right?

> Personally, I would prefer to keep it the way things are 

That's fine with me. Thanks for the context.

jmazzitelli:
> Another benefit is having the same defaults set for non-operator deployments like through helm right?

That's true -- not all of the settings are defined by helm values (though a lot are, see: https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml#L17)

PR #453, opened 2021-11-18
Adding root namespace to accessible namespaces

leandroberetta:


jmazzitelli:
see my review of the code and why i think this will work, as commented in these two comments:
* https://github.com/kiali/kiali/pull/4470#issuecomment-971671805
* https://github.com/kiali/kiali/pull/4470#issuecomment-971680519

leandroberetta:
@jmazzitelli yes, I tried this solution and works fine. The root_namespace defaults to istio_namespace in case of not being configured by the user.

PR #454, opened 2021-11-19
ingress_enabled is deprecated, but still use it if the new enabled setting isn't used

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4510

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -78,6 +78,18 @@
 - debug:
     msg: "OPERATOR VERSION: [{{ operator_version }}]"
 
+# To remain backward compatible with some settings that have changed in later releases,
+# let's take some deprecated settings and set the current settings appropriately.
+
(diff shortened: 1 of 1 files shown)


jmazzitelli:
For this test procedure, when I say "dot" setting I mean `spec.deployment.ingress.enabled` and when I say "underscore" setting I mean `spec.deployment.ingress_enabled`.

Test on minikube. Create simple Kiali CR without dot setting or underscore setting. On minikube, there is no ingress created by default - confirm that (`kubectl get ingress --all-namespaces`).

Now edit the Kiali CR and add the underscore setting and set it to true. When operator is done reconciling, see that the ingress was created (`kubectl get ingress --all-namespaces`).

Now edit the Kiali CR and add the dot setting, too, but set the dot setting to false. So the Kiali CR should have:

```yaml
spec:
  deployment:
    ingress_enabled: true
    ingress:
      enabled: false
```

When both underscore and dot settings both exist, the dot setting takes precedence  (remember, the underscore setting is being deprecated). So when the operator reconciles that, you should NOT see an ingress anymore.

aljesusg (approved):
LGFM

PR #460, opened 2021-12-03
[operator] upgrade base image

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4540
fixes: https://github.com/kiali/kiali/issues/4121

backport PR: https://github.com/kiali/kiali-operator/pull/462 (we will not merge this - this backport PR has been closed)

PR #462, opened 2021-12-06
[v1.36] [operator] upgrade base image

jmazzitelli:
* update operator to new base image and new ansible k8s collection fixes: https://github.com/kiali/kiali/issues/4540
* switch default and v1.36 playbooks to use kubernetes/core v2
* kubernetes.core query behaves differently now - use lookup instead in the remove task
* for the upstream releases, we only will support "default" version (though technically v1.36 will work, too). If we want to support "v1.12" and "v1.24" then we will need to add 1.2.1 of community.kubernetes to the requirements.yaml

backport of:
* https://github.com/kiali/kiali/issues/4540
* https://github.com/kiali/kiali-operator/pull/460


jmazzitelli:
I installed OSSM 2.1 and ran the molecule tests with "spec.version" set to "v1.36" and all pass.
I had this PR branch's operator pushed to my cluster, the helm chart repo's v1.36 branch checked out locally (so the molecule test uses it when installing the operator via helm) and then ran this:
`hack/run-molecule-tests.sh --cluster-type openshift --client-exe $(which oc) -dorp podman -udi true -sv v1.36`

jmazzitelli:
we are not going to backport this change.

PR #467, opened 2021-12-14
Ignore workload validation by default

leandroberetta:
Ignoring by default KIA1201 that relates to workload not covered by any auth policy. 

This is for [#4409](https://github.com/kiali/kiali/pull/4409).

The motivation for this is that is pretty common for some workloads to not have an Authorization Policy associated with them. So to avoid filling the UI with errors, we decided to make this validation optional.

missing_ap_list

### DIFF
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -269,7 +269,7 @@ kiali_defaults:
       namespaces: []
       refresh_interval: "15s"
     validations:
-      ignore: []
+      ignore: ["KIA1201"]
 
   kubernetes_config:
     burst: 200
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -373,8 +373,7 @@ spec:
       namespaces: ["istio-system"]
       refresh_interval: "15s"
     validations:
-      # default: ignore is an empty list
-      ignore: ["KIA0101"]
+      ignore: ["KIA1201"]
 
   kubernetes_config:
     burst: 200


roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -261,7 +261,7 @@ kiali_defaults:
       namespaces: []
       refresh_interval: "15s"
     validations:
-      ignore: []
+      ignore: ["KIA1201"]
```
jmazzitelli, line 264: because we are changing the defaults, update the exampe kiali_cr.yaml which is the way we document for users what the defaults are. So you want to make a similar change here: https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L911

jmazzitelli (changes requested):

roles/default/kiali-deploy/defaults/main.yml
leandroberetta: Right, thanks for the heads up.

jmazzitelli (approved):

deploy/kiali/kiali_cr.yaml
```diff
@@ -908,7 +908,7 @@ spec:
 #
 #      If you wish to ignore one or more validation errors, put their validation codes (e.g. KIA0101) in this list.
 #      ---
-#      ignore: []
+#      ignore: ["KIA1201"]
```
jmazzitelli, line 911: Now that we have the automated cr docs being generated, this file is deprecated (though we can still manage it for now if we want, so you can leave this line 911 alone).. but you now have to modify this file with the new default: https://github.com/kiali/kiali-operator/blob/master/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L375-L377

I would change those lines 375-377 so they look like this:
```
    validations:
      ignore: ["KIA1201"]
```

jmazzitelli (changes requested):

deploy/kiali/kiali_cr.yaml
leandroberetta: Thanks, resolved.

jmazzitelli (approved):
looks good!

PR #469, opened 2021-12-21
kiali cr validation tool

jmazzitelli:
_(this PR is currently being superceded by https://github.com/kiali/kiali-operator/pull/470/files - the same motivation exists, so this description is still valid. But that new PR will also contain a make target that will auto-generate CR documentation for the kiali.io website as well as includes an example CR in addition to this validation script)._

tl;dr run this tool to validate that your Kiali CR is correct

Right now, the Kiali CRD has a "preserve all unknown fields" schema to allow the Kiali CR to be basically open-ended with the settings it allows. We do this for a couple reasons: (1) it allows a single CRD to support multiple versions of the Kiali CR in order to easily support multiple versions of the server and (2) it means there is minimal maintenance needed in the operator releases. It is non-trivial to implement and ship operators with multiple CRDs and have OLM deploy and manage them. And Kiali very frequently adds/deletes/modifies the settings that are supported in the Kiali CR, so it is conceivable for every new operator release a new CRD version with a new schema will be needed to be deployed.

With that said, users are very often running into problems because their Kiali CR is not formatted correctly or has incorrect settings.

What are the options? The options are to officially deploy CRDs with correct schemas (for reasons explained above, we would like to avoid this right now) or to create a validation script that will validate existing Kiali CRs for correctness.

This PR does the latter. There is a validation script that contains a CRD with a schema that supports the current master schema. You can validate a Kiali CR yaml file or an already-existing Kiali CR found in any namespace in your cluster. You can run it like this:

* If you have a Kiali CR yaml file but have not created the Kiali CR resource yet:
```
./validate-kiali-cr.sh --kiali-cr-file "./my-kiali-cr.yaml"
```
* If you have already created a Kiali CR named `kiali` in the namespace `istio-system`:
```
./validate-kiali-cr.sh --kiali-cr-name "kiali" --namespace "istio-system"
```

Any errors found in the Kiali CR will be output by the script.

The thinking is you can run this script via curl so anyone can use it to validate a CR without having to download additional files or git clone any repos. As Kiali releases new versions, the validation script for the tagged branch can be used to validate a CR for that version of Kiali. e.g.

```
bash <(curl -L https://raw.githubusercontent.com/kiali/kiali-operator/v1.45.0/deploy/validate-kiali-cr.sh) --kiali-cr-name "kiali" --namespace "istio-system"
```


jmazzitelli:
You can try this now by running the script from this PR branch:

```
bash <(curl -L https://raw.githubusercontent.com/jmazzitelli/kiali-operator/openapi-crd/deploy/validate-kiali-cr.sh) --kiali-cr-name "kiali" --namespace "istio-system"
```

jmazzitelli:
I added an example kiali_cr_full.yaml file with all the settings set. You can use this to validate the validator. Try it like this (after you pull this PR):

```
./validate-kiali-cr.sh -kcf kiali/kiali_cr_full.yaml
```

Should output this at the end:

```
Kiali CR [kiali/kiali_cr_full.yaml] is valid.
```

If you edit that yaml file and make something invalid (like set `server.web_schema` to an integer `1`), you will see it fail like this:

```
The TestKiali "kiali" is invalid: spec.server.web_schema: Invalid value: "integer": spec.server.web_schema in body must be of type string: "integer"
ERROR! Validation failed for Kiali CR [kiali/kiali_cr_full.yaml]
```

jmazzitelli:
I'm going to close this in lieu of https://github.com/kiali/kiali-operator/pull/470

PR #471, opened 2021-12-26
fix: update '/healthz' APIs

Xunzhuo:
fix: https://github.com/kiali/kiali/pull/4632


jmazzitelli:
This is wrong - the older versions you are changing (e.g. 1.24) will not be on /api so this will break those. I will assume you did not test this PR with the previous versions because if you did, you probably will have seen the readiness probes fail and kubernetes kill the pods.

Regardless, I'm closing this with -1 - please see https://github.com/kiali/kiali/pull/4632#issuecomment-1001191346 for the reason.

PR #354, opened 2021-07-12
session timeout test

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4164

PR #470, opened 2021-12-22
openapi schema in CRD, validation tool, and auto-generation of doc for examples

jmazzitelli:
same as https://github.com/kiali/kiali-operator/pull/469 so please read that PR description for the motivation.

This PR includes the same validation script but it also includes descriptions in the CRD schema and a make target that can be used to [generate a .md file](https://github.com/giantswarm/crd-docs-generator) containing documentation for use in a hugo-generated website (such as the kiali.io website). The documentation can be used in place of the existing example kiali_cr.yaml, as an example kiali CR will be included in the documentation.

PR #473, opened 2021-12-27
Add note about which strategies use of "expiration_seconds" config

israel-hdez:
Per discussion in https://github.com/kiali/kiali-operator/pull/354


jmazzitelli (approved):
I approve. Just a note: eventually, this kiali_cr will not have to be updated in the future - we'll instead have a CRD schema to update which is then used by a tool to auto-generate kiali.io docs. See my draft PR for details on my proposal: https://github.com/kiali/kiali-operator/pull/470

PR #474, opened 2022-01-12
Enforce specific lengths for signing key

israel-hdez:
This is to support sessions with AES-GCM encryption.

Related to kiali/kiali#4542

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -680,6 +680,19 @@
   # this regex is not 100% accurate, but we want to at least catch obvious errors
   - kiali_vars.api.namespaces.label_selector is not regex('^[a-zA-Z0-9/_.-]+=[a-zA-Z0-9_.-]+$')
 
+# If the signing key is not empty string, and is not of the special value secret:name:key,
+# do some validation on it's length
+- name: Validate signing key, if it is set in the CR
+  fail:
+    msg: "Signing key must be 16, 24 or 32 byte length"
+  when:
+    - kiali_vars.auth.strategy != 'anonymous'
+    - kiali_vars.login_token.signing_key != ""
+    - not(kiali_vars.login_token.signing_key | regex_search('secret:.+:.+'))
+    - kiali_vars.login_token.signing_key | length != 16
+    - kiali_vars.login_token.signing_key | length != 24
+    - kiali_vars.login_token.signing_key | length != 32
+
 # If the signing key is empty string, we need to ensure a signing key secret exists. If one does not exist, we need to generate one.
 # Note that to avoid granting to the operator the very powerful permission to CRUD all secrets in all namespaces, we always generate
 # a signing key secret with the name "kiali-signing-key" regardless of the value of kiali_vars.deployment.instance_name.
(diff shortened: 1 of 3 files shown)


roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -680,6 +680,19 @@
...
+- name: Validate signing key, if it is set in the CR
+  fail:
+    msg: "Signing key must be 16, 24 or 32 byte length"
+  when:
+    - kiali_vars.auth.strategy != 'anonymous'
+    - kiali_vars.login_token.signing_key != ""
+    - not(kiali_vars.login_token.signing_key | regex_search('secret:.+:.+'))
+    - kiali_vars.login_token.signing_key | length != 16
+    - kiali_vars.login_token.signing_key | length != 24
+    - kiali_vars.login_token.signing_key | length != 32
```
jmazzitelli, line 694: would be good to have a simple molecule test for this... somewhere in here? https://github.com/kiali/kiali-operator/blob/master/molecule/config-values-test/converge.yml#L27

also, make sure the other molecule tests pass - like https://github.com/kiali/kiali-operator/blob/master/molecule/null-cr-values-test/converge.yml#L41

israel-hdez: > also, make sure the other molecule tests pass - like https://github.com/kiali/kiali-operator/blob/master/molecule/null-cr-values-test/converge.yml#L41

That one passes... However, all three auth molecules don't :-)
I'll check what I can do...

> don't need to test all 16, 24 and 32 - just to test to get into there and see this pass through ok

I think I don't get this...

jmazzitelli: > I think I don't get this...

I mean to say: you don't have to test with 3 separate signing keys - if you just test with one (say, of size 16) and see that it works, that's good enough

jmazzitelli:
When we merge this, we must submit a PR against the Istio addon because we use `CHANGEME` as the signing key over there - and that isn't one of the valid sizes. We need to change it to `CHANGEMECHANGEME` or something similar. See: https://github.com/istio/istio/blob/1.12.2/manifests/addons/values-kiali.yaml#L12

roles/default/kiali-deploy/tasks/main.yml
israel-hdez: @jmazzitelli I added a simple test. I hope that's enough.

israel-hdez:
@jmazzitelli Sorry for the rush. But I wonder if we can merge this?
The changes to the token auth kind-of require this.

israel-hdez:
...and also, the back-end side is already merged: https://github.com/kiali/kiali/pull/4655

jmazzitelli (approved):
i haven't run this, but the changes and test look good.

israel-hdez:
> When we merge this, we should submit a PR against the Istio addon because we use `CHANGEME` as the signing key over there - and that isn't one of the valid sizes. We need to change it to `CHANGEMECHANGEME` or something similar. See: https://github.com/istio/istio/blob/1.12.2/manifests/addons/values-kiali.yaml#L12

I think I'm going to remove that string. It's a little misleading it's there when it's unused.

PR #475, opened 2022-01-13
Remove Iter8 extension

lucasponce:
Related to https://github.com/kiali/kiali/issues/4643

https://github.com/kiali/kiali-ui/pull/2284
https://github.com/kiali/kiali/pull/4653

### DIFF
--- a/manifests/kiali-community/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml
@@ -527,6 +527,16 @@ spec:
           - routes
           verbs:
           - get
+        - apiGroups: ["iter8.tools"]
+          resources:
+          - experiments
+          verbs:
+          - get
+          - list
+          - watch
+          - create
+          - delete
+          - patch
         - apiGroups: ["authentication.k8s.io"]
           resources:
           - tokenreviews
--- a/manifests/kiali-community/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml
@@ -527,16 +527,6 @@ spec:
           - routes
           verbs:
           - get
-        - apiGroups: ["iter8.tools"]
-          resources:
-          - experiments
-          verbs:
-          - get
-          - list
-          - watch
-          - create
-          - delete
-          - patch
         - apiGroups: ["authentication.k8s.io"]
           resources:
           - tokenreviews
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -586,6 +586,16 @@ spec:
           verbs:
           - get
           - list
+        - apiGroups: ["iter8.tools"]
+          resources:
+          - experiments
+          verbs:
+          - get
+          - list
+          - watch
+          - create
+          - delete
+          - patch
         - apiGroups: ["authentication.k8s.io"]
           resources:
           - tokenreviews
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -586,16 +586,6 @@ spec:
           verbs:
           - get
           - list
-        - apiGroups: ["iter8.tools"]
-          resources:
-          - experiments
-          verbs:
-          - get
-          - list
-          - watch
-          - create
-          - delete
-          - patch
         - apiGroups: ["authentication.k8s.io"]
           resources:
           - tokenreviews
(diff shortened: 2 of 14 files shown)


jshaughn (commented):
The removals look correct.  I think we can remove this without backward compatibility errors, as any definitions in existing CRs should, I think, just be ignored.  The changes are relevant only to the iter8 extension.  I would like to hear from @jmazzitelli, just to confirm.  We would lose any support for iter8 in OSSM 2.0 and 2.1, when 2.2 is released.  I don't think anyone is actually using iter8 but I don't have any proof.  If the iter8 extension was experimental and never truly supported, I guess it wouldn't matter.

jshaughn:
Built and sanity checked with operator, server and ui prs.

lucasponce:
Thanks. I have also some concerns about backward compatibility, too. Well, not fully worried as this extension is experimental but I haven't removed the data from the previous versions to "make it compatible". But some look to check if that's the right way would be good.

I think this PR can wait until we have a proper test, if not shipped in 1.45, it can get in for the next one, as the last thing I'd like to do is to introduce a regression in the operator.

pbajjuri20:
I have tested with Operator PR, it looks good to me. Nothing broken

pbajjuri20 (approved):

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
```diff
@@ -586,16 +586,6 @@ spec:
...
-        - apiGroups: ["iter8.tools"]
-          resources:
-          - experiments
-          verbs:
-          - get
-          - list
-          - watch
-          - create
-          - delete
-          - patch
```
jmazzitelli, line 598: We probably need to revert this. This `manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml` is the CSV for the productized Kiali operator. This must support v1.12 and v1.24 as well as v1.36 Kiali servers. Because I assume those older versions of Kiali still support the iter8 extension, we need our productized CSV to keep this permission (otherwise the older productized versions of Kiali server will not be able to be installed by the productized operator).

We can remove this section only when all supported versions of Kiali server no longer support the iter8 extension.

jmazzitelli (changes requested):

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
lucasponce: Reverted.
Probably the risk of problems is very low as this is unused in the product but better to be conservative here.

manifests/kiali-community/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml
```diff
@@ -527,16 +527,6 @@ spec:
...
-        - apiGroups: ["iter8.tools"]
-          resources:
-          - experiments
-          verbs:
-          - get
-          - list
-          - watch
-          - create
-          - delete
-          - patch
```
jmazzitelli, line 539: I just noticed something. We already released 1.45 - see https://quay.io/repository/kiali/kiali?tab=tags and https://quay.io/repository/kiali/kiali-operator?tab=tags

That means these 1.45 CSV metadata files should be fixed to that release (and we need to publish these on operatorhub.io, which we have not done yet - we are only at 1.44 - see https://operatorhub.io/operator/kiali)

Therefore, we need to revert these changes to `manifests/kiali-community/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml` and `manifests/kiali-upstream/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml`

I'll create a new 1.46.0 version of both those files and then you can update them for this PR. I'll also have to publish 1.45 to operatorhub and the redhat registry.

manifests/kiali-upstream/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml
```diff
@@ -527,16 +527,6 @@ spec:
...
-        - apiGroups: ["iter8.tools"]
-          resources:
-          - experiments
-          verbs:
-          - get
-          - list
-          - watch
-          - create
-          - delete
-          - patch
```
jmazzitelli, line 539: revert this as per my previous comment

jmazzitelli (changes requested):
see my comments about reverting the 1.45 metadata and how we need new metadata files for this PR

manifests/kiali-community/1.45.0/manifests/kiali.v1.45.0.clusterserviceversion.yaml
lucasponce: Yes, thanks for catching it.
This PR was created for 1.45 but now we need to move it.

jmazzitelli: I updated my review comment above to explain what we need to do. I guess I should have just added a comment here. I'll do that now:

1. remove these changes to 1.45 OLM metadata
2. merge this PR with the new 1.46 OLM metadata - #478
3. Rebase your PR so you get the new 1.46 OLM metadata
4. Make your changes to the new 1.46 OLM metadata (i.e. the changes you revert from the 1.45 metadata need to be done to the 1.46 metadata).

lucasponce:
I think the changes won't change the feedback from @pbajjuri20. 
So, I think we can consolidate and fix it later if there is any new issue.

PR #480, opened 2022-01-27
update some things since we are now publishing the cr docs on kiali.io

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/4684

Do not merge this unless and until this is merged: https://github.com/kiali/kiali.io/pull/506


crd-docs/config/apigen-crd.template
```diff
@@ -1,91 +1,22 @@
...
-</dd>
-</dl>
-
 {{ if .VersionSchemas }}
 {{ range $versionName, $versionSchema := .VersionSchemas }}
 <div class="crd-schema-version">
-<h2 id="{{$versionName}}">Version {{$versionName}}</h2>
 
 {{with .ExampleCR}}
 <h3 id="crd-example-{{$versionName}}">Example CR (all values shown here are the defaults unless otherwise noted)</h3>
```
jshaughn, line 22: Can you move the `(all values shown here are the defaults unless otherwise noted)` to just be regular text below the heading?

jmazzitelli: change made - now looks like:
image

jshaughn (dismissed):
This looks fine.  I think when there is a change to the CR we should require the person to also perform the doc re-gen and create a kiali.io PR to go with the operator and server prs.

jshaughn (approved):
re-approving, horizontal rules look good between props.

jmazzitelli:
added new commit to remove the crd version from the url links since we dont' need that granularity and it makes the urls longer and ugly

jshaughn (approved):
re-approving after mazz took my suggestion to drop the unnecessary version and cleanup the links.

PR #484, opened 2022-02-01
Add 1.46 role

aljesusg:
1.46 will be replaced by 1.48 

Do not merge

### DIFF
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -102,8 +102,13 @@ spec:
     ### Kiali Custom Resource Configuration Settings
 
     For quick descriptions of all the settings you can configure in the Kiali
+<<<<<<< HEAD
     Custom Resource (CR), see
     [the kiali.io docs](https://kiali.io/docs/configuration/kialis.kiali.io/).
+=======
+    Custom Resource (CR), see the file
+    [kiali_cr.yaml](https://github.com/kiali/kiali-operator/blob/v1.46/deploy/kiali/kiali_cr.yaml)
+>>>>>>> c256fed (Add 1.46 role)
 
     ## Prerequisites for enabling this Operator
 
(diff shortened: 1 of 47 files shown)


manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
```diff
@@ -103,7 +103,7 @@ spec:
 
     For quick descriptions of all the settings you can configure in the Kiali
     Custom Resource (CR), see the file
-    [kiali_cr.yaml](https://github.com/kiali/kiali-operator/blob/v1.36/deploy/kiali/kiali_cr.yaml)
+    [kiali_cr.yaml](https://github.com/kiali/kiali-operator/blob/v1.46/deploy/kiali/kiali_cr.yaml)
```
jmazzitelli, line 106: This section (lines 104-106) is different now. 1.46 branch won't have the kiali_cr yaml - so you need to change this to look like what master looks like now: https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L104-L106

```
    For quick descriptions of all the settings you can configure in the Kiali
    Custom Resource (CR), see
    [the kiali.io docs](https://kiali.io/docs/configuration/kialis.kiali.io/).
```

jmazzitelli (changes requested):
Aside from the help text that needs to change in the CSV, this looks good.

But can you confirm that the `roles/v1.46` is simply a copy of the current `roles/default`? That's what it should be (I just want to make sure you didn't copy `roles/v1.36` - the new version should be a copy of the `roles/default` directory).

I also +1 the changes you made to add the new 1.46 to the default supported images. That is all correct.

aljesusg:
> Aside from the help text that needs to change in the CSV, this looks good.
> 
> 
> 
> But can you confirm that the `roles/v1.46` is simply a copy of the current `roles/default`? That's what it should be (I just want to make sure you didn't copy `roles/v1.36` - the new version should be a copy of the `roles/default` directory).
> 
> 
> 
> I also +1 the changes you made to add the new 1.46 to the default supported images. That is all correct.

I am 100% sure that I copied the 1.36 and not the default 🤣 let me change it 😊

jmazzitelli (approved):

aljesusg:
I think @jmazzitelli that if we are going to update this later to the version 1.48 we can avoid this merge and only do the PR to the branch 1.46 https://github.com/kiali/kiali-operator/pull/488 . What do you think?

jmazzitelli:
> only do the PR to the branch 1.46 #488 . What do you think?

yes, that's a good idea.

aljesusg:
Closed in favor https://github.com/kiali/kiali-operator/pull/488

PR #485, opened 2022-02-03
Support being able to reliably load dev images from the internal OpenShift image registry

jmazzitelli:
This is for when running molecule tests and you want to test with dev images that are pushed directly into the OpenShift internal registry.

Requires kiali repo changes that go with this.

fixes: https://github.com/kiali/kiali/issues/4691


jshaughn (approved):
Looks sane.

PR #489, opened 2022-02-07
api.namespaces.exclude defaults should have "starts with" regexes

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4714

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -19,11 +19,11 @@ spec:
   api:
     namespaces:
       exclude:
-      - "istio-operator"
-      - "kube-.*"
-      - "openshift.*"
-      - "ibm.*"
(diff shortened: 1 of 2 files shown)


jmazzitelli:
How to test:

1. Create a namespace `akube-foo`
2. Install Kiali Operator and Kiali (e.g. on dev machine, `make build cluster-push operator-create kiali-create`)
3. Log into the Kiali Console (`hack/kiali-port-forward.sh` can work if on minikube)
4. Confirm that the Graph page shows `akube-foo` option in the namespace dropdown; confirm you see that namespace in the Overview page.

jshaughn (approved):
I was thinking maybe we needed to update the doc as well, but the default values should now make it clear that `^` is necessary if you want "starts with" as opposed to "substring" regex matching.  LGTM

pbajjuri20:
I have deployed on OpenShift, It worked as expected.


operator_pr_498

pbajjuri20 (approved):

PR #493, opened 2022-02-08
[v1.36] Add skipRange to kiali-ossm CSV.

jmazzitelli:
This is backporting the kiali-ossm manifest change found in https://github.com/kiali/kiali-operator/pull/492

This is going to help customers that are in a disconnected environment - they will not have to upgrade all the individual operator versions.

There needs to be some discussion before deciding to merge this. We may decide to hold off from doing this (in which case, just close this PR without merging).


jmazzitelli:
We are going to defer this until later.

PR #496, opened 2022-02-28
Add role 1.48 for ossm 2.2

aljesusg:
Wait until release but prepare the role in operator to be able to make the build

- Add kiali 1.48
- Remove 1.12 ?


jmazzitelli:
This looks correct (looks like the v1.48 role is a copy of the current "default" role, which is what we want).
We aren't sure if 1.12 will no longer be supported, so we should wait to get confirmation that we can delete 1.12 before we merge this.

aljesusg:
Rebased after https://github.com/kiali/kiali-operator/pull/497

aljesusg:
I updated them after https://github.com/kiali/kiali-operator/pull/497 I'll check them this thursday to be sure.

aljesusg:
Update role 1.48 from master

jmazzitelli (approved):
LGTM

doing a `diff -r default/ v1.48/` shows no diff - so the v1.48 role is identical to the default role - that is what we want.

PR #508, opened 2022-03-30
Adjust molecule openshift-auth-test because of auth refactor

israel-hdez:
Related back-end PR: kiali/kiali#4834

* Cookie name changes from kiali-token to kiali-token-aes
* The JTW token is no longer returned in the JSON when auth succeeds.


jmazzitelli (approved):
this test passes.

PR #521, opened 2022-04-29
add remote openshift parameters

libesz:
Related to https://github.com/kiali/kiali/pull/5021

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -50,6 +50,7 @@ spec:
       username_claim: "sub"
     openshift:
       client_id_prefix: "kiali"
+      custom_ca: ""
       redirect_url: ""
       server_prefix: "https://kubernetes.default.svc/"
       use_system_ca: false
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -51,7 +51,7 @@ spec:
     openshift:
       client_id_prefix: "kiali"
       redirect_url: ""
-      server_prefix: ""
+      server_prefix: "https://kubernetes.default.svc/"
       use_system_ca: false
 
   # default: custom_dashboards is an empty list
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -51,7 +51,6 @@ spec:
     openshift:
       client_id_prefix: "kiali"
       custom_ca: ""
-      redirect_url: ""
       server_prefix: "https://kubernetes.default.svc/"
       use_system_ca: false
 
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -50,6 +50,9 @@ spec:
       username_claim: "sub"
     openshift:
       client_id_prefix: "kiali"
+      redirect_url: ""
+      server_prefix: ""
+      use_system_ca: false
 
   # default: custom_dashboards is an empty list
   custom_dashboards:
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -177,6 +177,9 @@ spec:
                       client_id_prefix:
                         description: "The Route resource name and OAuthClient resource name will have this value as its prefix. This value normally should never change. The installer will ensure this value is set correctly."
                         type: string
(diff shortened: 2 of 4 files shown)


crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -50,6 +50,9 @@ spec:
       username_claim: "sub"
     openshift:
       client_id_prefix: "kiali"
+      redirect_url: ""
+      server_prefix: ""
```
jmazzitelli, line 54: Under most circumstances, the values in this example file should show the default values (people refer to this file to see what the defaults are). Based on the `defaults/main.yml` file that this PR is modifying, this should be:
```suggestion
      server_prefix: "https://kubernetes.default.svc/"
```

jmazzitelli (changes requested):

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
libesz: So should we document here what is hardwired in the code for defaults, when the user does not set the value? https://github.com/kiali/kiali/pull/5021/commits/3ad3ff92122e0feae44ea33a388522fa359c03e0#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R532

jmazzitelli: In short, these config/config.go settings are what our unit tests see :) That's why we set them here.

libesz: Makes sense 😄

libesz:
@jmazzitelli I ran into an issue when I attempt to install Kiali on a K8S cluster, while the actual mesh cluster is Openshift (which is the purpose of this PR).

```
TASK [default/kiali-deploy : Confirm auth strategy is valid for Kubernetes environments] ***
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Invalid auth.strategy [openshift]! Must be one of either 'token', 'openid', 'anonymous', or 'header'"}
```
In this case this may not be a user error, so I wonder if I could eliminate it in case when i.e. the server_prefix is provided in the auth settings, suggesting that the oauth server (and so the cluster itself) is not the one where kiali being installed on.

israel-hdez:
> I haven't thought about this before. Is this even a valid use case? Installing both non-OpenShift and OpenShift clusters (having Kiali in one and the service mesh in the other)?

Once you start looking at setups spanning over more than one cluster, I may think that it makes sense in a more general perspective. So, you may have one cluster in each cloud (let's think of AKS, EKS, GKS and well... perhaps OpenShift-on-AWS). I guess you are free to choose where to install Kiali, but if you want OpenShift to handle auth, well... I think you should be capable of it.

libesz:
Yes, so this is for sure a valid production use-case for us. We are the provider of an Openshift as a Service offer, where the mesh itself is operating on the user's cluster (Openshift in this case), but there is also a management cluster which belongs to the provider (not accessible for the user on k8s level). This management cluster may be pure K8S or Openshift, hosting Openshift master components, istiod for the mesh, and now Kiali. I think it should not really matter for Kiali because it will only talk to the mesh cluster (at least for oauth purposes and k8s API resources).

I added a commit to allow the installation in this tricky case.

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -50,6 +50,9 @@ spec:
       username_claim: "sub"
     openshift:
       client_id_prefix: "kiali"
+      custom_ca: ""
+      server_prefix: "https://kubernetes.default.svc/"
+      use_system_ca: false
```
jmazzitelli, line 55: Do we need "use_system_ca"? Could we use the "custom_ca" field for both? By that I mean, if "custom_ca" is empty string (the default), wouldn't that assume you want to use the system CA (use_system_ca=true)? And if "custom_ca" is not an empty string, wouldn't that mean you are NOT using the system CA (i.e. use_system_ca=false)??

In other words: `use_system_ca = (custom_ca == "")`

jmazzitelli (commented):
I have a question about use_system_ca and custom_ca...

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
libesz: Yep, we need it because this now becomes a three way switch. The third case is the existing default behavior, when the CA is fetched from the local kubernetes injected file next to the API token and other stuff.
System CA however means kiali is using the public trusted CA certificate bundle packaged with the base image of the container.

jmazzitelli:
see discussion in https://github.com/kiali/kiali/pull/5021
We don't need this PR (at least not now). We can resurrect in the future if need be.

PR #523, opened 2022-04-29
add 1.50.0 olm metadata

jmazzitelli:
side note: we should probably get our new github action workflows to prepare these metadata files if they dont' already exist.
We forgot to add them prior to releasing 1.50.0. Not horrible, we'll just add them now and we can use these to publish to operatorhub.io, but it would be nice to automate the creation of these (or, perhaps after release, we create the metadata for the upcoming release - that would be the best thing to do.) I'll create another github issue for this.


jshaughn (approved):
looks bueno

PR #524, opened 2022-05-02
be able to add containers to Kiali pod

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5028


jmazzitelli:
molecule tests pass on minikube:
```
=====================
=== TEST RESULTS: ===
=====================

              accessible-namespaces-test... success [7m 12s]
     affinity-tolerations-resources-test... success [3m 36s]
                cluster-wide-access-test... success [8m 12s]
                      config-values-test... success [2m 33s]
                 container-security-test... success [3m 58s]
                  default-namespace-test... success [1m 41s]
                            grafana-test... success [3m 19s]
                        header-auth-test... success [1m 40s]
                      instance-name-test... success [3m 8s]
                             jaeger-test... success [2m 29s]
                            metrics-test... success [3m 27s]
                     null-cr-values-test... success [1m 38s]
                only-view-only-mode-test... success [2m 26s]
                             openid-test... success [1m 32s]
           remote-cluster-resources-test... success [2m 50s]
                              roles-test... success [4m 57s]
                    rolling-restart-test... success [2m 51s]
                              token-test... success [1m 56s]

Test logs can be found at: /tmp/kiali-molecule-test-logs.2025-08-25_06-06-10
```

jmazzitelli:
the new molecule tests pass (config-values-test, containers-security-test, init-containers-security-test) when helm installs the operator and when OLM installs the operator.

jmazzitelli:
I consolidated the 2 new molecule tests into one (now called containers-test). I explicitly ran the following molecule tests  (all the rest should be the same as before - they all passed earlier)

(helm installed operator)

=====================
=== TEST RESULTS: ===
=====================

                                 default... success [2m 42s]
                      config-values-test... success [3m 9s]
                         containers-test... success [4m 53s]
                             jaeger-test... success [3m 9s]
                        header-auth-test... success [2m 33s]
                             openid-test... success [2m 22s]
           remote-cluster-resources-test... success [3m 37s]
                     null-cr-values-test... success [2m 43s]
                only-view-only-mode-test... success [3m 9s]

jmazzitelli:
(tests with OLM-installed operator)

```
Skipping helm - will use the operator that is already installed

=====================
=== TEST RESULTS: ===
=====================

                      config-values-test... success [2m 26s]
                         containers-test... success [5m 21s]
                             openid-test... success [1m 32s]
           remote-cluster-resources-test... success [2m 55s]
```

jmazzitelli:
The tests will not run on OpenShift because of OpenShift's more restrictive security policies. Need to figure out how to get the new molecule tests to pass on OpenShift:

Here's an analysis of the config-values-test failure, but the same conclusion can be made for the containers-test:

# Analysis of the `config-values-test` Molecule Test Failure

## **Root Cause:**
The test is failing because of **OpenShift Security Context Constraints (SCCs)** that are much more restrictive than Kubernetes RBAC policies used in minikube.

## **What's Happening:**

1. **Test Environment Differences:**
   - **Minikube**: Uses standard Kubernetes security policies which are more permissive
   - **OpenShift**: Uses Security Context Constraints (SCCs) that are much more restrictive

2. **The Specific Failure:**
   The test is intentionally configuring containers with privileged security contexts to verify that:
   - `ALLOW_SECURITY_CONTEXT_OVERRIDE=true` allows custom security contexts to be preserved
   - The operator doesn't override user-specified security settings when this flag is enabled

3. **Test Configuration That's Causing Issues:**
   Looking at lines 306-353 in `converge.yml`, the test adds containers with these intentionally permissive settings:
   ```yaml
   securityContext:
     runAsUser: 1001                    # ❌ OpenShift requires UIDs in range [1000650000, 1000659999]
     runAsGroup: 1001                   # ❌ OpenShift requires GIDs in range [1000650000, 1000659999]  
     runAsNonRoot: false                # ❌ OpenShift requires runAsNonRoot: true
     privileged: true                   # ❌ OpenShift forbids privileged containers
     allowPrivilegeEscalation: true     # ❌ OpenShift forbids privilege escalation
     capabilities:
       add: ['SYS_ADMIN', 'NET_ADMIN']  # ❌ OpenShift forbids adding capabilities
   ```

4. **OpenShift SCC Violations:**
   The error message shows the pod is rejected because it violates the `restricted-v2` SCC:
   ```
   unable to validate against any security context constraint: [
     provider "anyuid": Forbidden: not usable by user or serviceaccount,
     provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 1001: must be in the ranges: [1000650000, 1000659999],
     provider restricted-v2: .initContainers[0].privileged: Invalid value: true: Privileged containers are not allowed,
     provider restricted-v2: .initContainers[0].capabilities.add: Invalid value: "SYS_ADMIN": capability may not be added,
     ...
   ]
   ```

## **Why It Works on Minikube but Fails on OpenShift:**

- **Minikube**: Uses standard Kubernetes security policies that generally allow these configurations or have no admission controllers enforcing strict security contexts
- **OpenShift**: Uses SCCs which are admission controllers that actively reject pods that don't meet security requirements

## **The Test's Intent vs. OpenShift Reality:**

The test is designed to verify that when `ALLOW_SECURITY_CONTEXT_OVERRIDE=true` is set, the operator preserves user-specified security contexts without modification. However, in OpenShift:

1. Even if the **operator** allows security context override
2. The **platform itself** (via SCCs) will reject the pod before it's even created
3. This creates a conflict between what the test expects to verify and what the platform allows

## **This is NOT a Bug in Kiali:**

This is actually expected behavior. The test failure reveals a fundamental difference in how security is handled:
- **Minikube**: Relies on pod security policies or no enforcement
- **OpenShift**: Has mandatory security context constraints that cannot be bypassed

## **Potential Solutions:**

1. **Platform-Specific Test Logic**: Modify the test to use different security contexts based on whether it's running on OpenShift vs. Kubernetes
2. **Test Skip Logic**: Skip the privileged container tests when running on OpenShift
3. **Use OpenShift-Compatible Security Contexts**: Modify the test to use security contexts that comply with OpenShift SCCs while still testing the security context override functionality

The most appropriate solution would be #1 - use platform-specific test configurations that test the same functionality but with security contexts that are appropriate for each platform.

jmazzitelli:
The need for SCCs has been addressed in the current molecule tests.

jmazzitelli:
after rebasing (which picked up the formal crd schema), I ran molecule tests. All looks good. Even ran OLM and helm install tests on OpenShift 4.19.

jshaughn (changes requested):
I think the test instructions are out of date. They passed for me in the original impl but they did not pass with the more recent changes. Instead, I tried running the molecule tests (containter-tests), which have been updated.  But it also failed. So, please take another look at the test instructions and/or the molecule tests. I'm adding the molecule logs... 
[containers-test.log](https://github.com/user-attachments/files/22261758/containers-test.log)

jmazzitelli:
I just rebased all PRs, pushed them all. I am now running molecule tests.

stand up test environment:
```bash
./hack/k8s-minikube.sh --dex-enabled true -mp ci start && ./hack/k8s-minikube.sh -mp ci istio
```
build and push:
```bash
make CLUSTER_TYPE=minikube MINIKUBE_PROFILE=ci build-ui build test cluster-push
```
run tests:
```bash
./hack/run-molecule-tests.sh --client-exe "$(which kubectl)" --cluster-type minikube --minikube-profile ci -udi true -hcrp false 
```

jmazzitelli:
The two molecule test failures are expected - once Istio 1.27.2 is released, they should pass. We are going to hold off on merging these anyway - perhaps get these in the next sprint release (not the current sprint release).

At that time, we'll have to remember to update this PR so 2.18 CSV metadata are updated, rather than 2.17

olsib:
Hi @jmazzitelli! 

We're awaiting this feature for Kiali Multi Cluster implementation


**The Challenge:**

The [Kiali multi-cluster documentation](https://kiali.io/docs/configuration/multi-cluster/) recommends using long-lived service account tokens for remote cluster authentication. However, **this approach doesn't meet our security and compliance requirements**:

**Our Use Case:**
- Kiali hub monitoring multiple remote EKS clusters
- Authentication via AWS IAM + EKS exec plugin (`aws eks get-token`)
- Need AWS CLI available in Kiali container

**Current Workaround:**
Custom Docker image with AWS CLI pre-installed (maintenance overhead).

**How This PR Helps:**
With `additional_pod_init_containers_yaml`, we could install AWS CLI at runtime via initContainer, eliminating custom images.

**Question:**
I see the Helm chart enforces `readOnlyRootFilesystem: true` for initContainers. Would this prevent copying AWS CLI to a shared emptyDir volume for the main container?

**Alternative:** Could we optionally disable `readOnlyRootFilesystem` for specific initContainers when needed?

This would be hugely valuable for AWS EKS users wanting to use native IAM authentication for multi-cluster instead of long-lived tokens.

**Alternative Approaches:** Are there any other recommended patterns for including external tools (like AWS CLI) in the Kiali container without custom images?

Thanks for your work on this! 

---

**Context:**
- Kiali v2.16.0 (Helm)
- AWS EKS 1.30+
- Cross-account hub-spoke with IRSA + IAM assume role
- Related: #8616, #5028

olsib:
Thanks @jshaughn, I see the Helm chart enforces `readOnlyRootFilesystem: true` for initContainers. Would this prevent copying AWS CLI to a shared emptyDir volume for the main container? Any alternatives you propose?

PR #525, opened 2022-05-02
new olm metadata for 1.51

jmazzitelli:
In the future we'll autogenerate this when we release a new kiali
version (see https://github.com/kiali/kiali/issues/5025)
but for now I'm generating it manually here.


jshaughn (approved):
LOK

PR #526, opened 2022-05-09
Nightly pipeline and refinements to Kiali operator release pipeline

leandroberetta:
This PR contains the following:

- Adds a nightly build pipeline to generate a latest image every weekday at 04:00 UTC.
- Removes the snapshots so a release pipeline only has "minor" and "patches" options (latest is left for the nightly, that can also be started manually)
- Adds metadata generation

Resolves: https://github.com/kiali/kiali/issues/5025.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,22 +1,18 @@
 name: Release
 
 on:
+  schedule:
+    # Every Friday at 19:00 (UTC)
+    - cron: '00 19 * * FRI'
   workflow_dispatch:
     inputs:
       release_type:
         description: 'Release type'     
         required: true
-        default: 'auto' 
         type: choice
         options:
-        - auto
-        - edge
         - minor
         - patch
-        - snapshot.0
-        - snapshot.1
-        - snapshot.2
-        - snapshot.3
       release_branch:
         description: Branch to release
         required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -221,6 +221,18 @@ jobs:
     - name: Create or update version branch      
       run: git push origin $(git rev-parse HEAD):refs/heads/$BRANCH_VERSION
 
+    - name: Configure Operator SDK
+      if: ${{ needs.initialize.outputs.release_type == 'minor' }}
+      run: |
+        export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)
+        export OS=$(uname | awk '{print tolower($0)}')
+
+        export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.20.0
```
jmazzitelli, line 230: this probably isn't good grabbing the hardcoded 1.20. We use very specific version of the SDK. Our Makefile has a target you can use that downloads the SDK we use: https://github.com/kiali/kiali-operator/blob/master/Makefile#L63

jmazzitelli (changes requested):

.github/workflows/release.yml
leandroberetta: Thanks for the recommendation. It's resolved now.

PR #527, opened 2022-05-09
[molecule] conditionals to check when reconciliation was done didn't work 100% of the time.

jmazzitelli:
These new conditionals should be more specific and should work.

The issue was there was multiple status condition blocks that had `reason=Successful` in them - which resulted on the json query string being "TrueTrue".

I changed the conditionals to look specifically for the conditional that indicates the reconciliation is complete.


leandroberetta (approved):
I did not test it but looks good.

PR #528, opened 2022-05-09
[v1.48] found the conditionals here didn't work 100% of the time.

jmazzitelli:
cherry pick of PR #527


jshaughn (approved):
OK, does not affect any runtime code, so does not require a new build.

PR #534, opened 2022-05-11
[molecule] fix metrics-test

jmazzitelli:
OSSM 2.2 still uses "kubernetes_namespace"
Upstream Istio Prom addon uses "namespace"
Related to https://github.com/kiali/kiali/pull/4778

This now fixes all the molecules tests when running on OSSM 2.2.


jshaughn (approved):
LOK

PR #547, opened 2022-06-30
Increase refresh interval default and introduce new 2m duration option

jshaughn:
Part-of https://github.com/kiali/kiali/issues/5269

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1025,9 +1025,9 @@ spec:
                                 type: string
                                 enum: ["none","sent","received","total"]
                       metrics_per_refresh:
-                        description: "Duration of metrics to fetch on each refresh. Value must be one of: `1m`, `5m`, `10m`, `30m`, `1h`, `3h`, `6h`, `12h`, `1d`, `7d`, or `30d`"
+                        description: "Duration of metrics to fetch on each refresh. Value must be one of: `1m`, '2m', `5m`, `10m`, `30m`, `1h`, `3h`, `6h`, `12h`, `1d`, `7d`, or `30d`"
                         type: string
-                        enum: ["1m","5m","10m","30m","1h","3h","6h","12h","1d","7d","30d"]
+                        enum: ["1m","2m","5m","10m","30m","1h","3h","6h","12h","1d","7d","30d"]
                       metrics_inbound:
                         description: |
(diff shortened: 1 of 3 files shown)


Xunzhuo (commented):
Why set 60s to default value not 30s?

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1025,9 +1025,9 @@ spec:
                                 type: string
                                 enum: ["none","sent","received","total"]
                       metrics_per_refresh:
-                        description: "Duration of metrics to fetch on each refresh. Value must be one of: `1m`, `5m`, `10m`, `30m`, `1h`, `3h`, `6h`, `12h`, `1d`, `7d`, or `30d`"
+                        description: "Duration of metrics to fetch on each refresh. Value must be one of: `1m`, '2m', `5m`, `10m`, `30m`, `1h`, `3h`, `6h`, `12h`, `1d`, `7d`, or `30d`"
```
jmazzitelli, line 1028: wrong ticks
```suggestion
                        description: "Duration of metrics to fetch on each refresh. Value must be one of: `1m`, `2m`, `5m`, `10m`, `30m`, `1h`, `3h`, `6h`, `12h`, `1d`, `7d`, or `30d`"
```

jmazzitelli (changes requested):

jshaughn:
> Why set 60s to default value not 30s?

@Xunzhuo Remember, this is not the scrape interval, this is just the default refresh interval for the Refresh dropdown menu.  15s is usually too frequent, except for maybe demos of a small mesh.  I chose 1m to reduce default refresh by 4x, something a little more meaningful.  Note that the default is often not even used, as the URL will override it. And, of course, it is easy to change via the dropdown.

Xunzhuo:
> > Why set 60s to default value not 30s?
> 
> 
> 
> @Xunzhuo Remember, this is not the scrape interval, this is just the default refresh interval for the Refresh dropdown menu.  15s is usually too frequent, except for maybe demos of a small mesh.  I chose 1m to reduce default refresh by 4x, something a little more meaningful.  Note that the default is often not even used, as the URL will override it. And, of course, it is easy to change via the dropdown.
> 
> 

Yes, I realized just now 😂, this is more like the frequency from front end to call backend to refresh the graph.

jmazzitelli (approved):

PR #549, opened 2022-07-04
Set release branch correctly when the pipeline is scheduled

leandroberetta:
This is an addition that is already present on the Kiali server pipeline. 

This wasn't an issue on the last release as the last scheduled release failed because of another problem, making me to push the button manually (avoiding this bug).

Today, the pipeline failed because of this when creating the PR, on the last step, because of a missing value on the RELEASE_BRANCH env var. I created the PR and merge it so for today, it's fixed. 

The branch value is also used in the checkout's ref parameter, which was set right because "master" is the default branch to fetch in the absence of a ref value.

For the next time, it's better to have the ref parameter properly set.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v3
       with:
-        ref: ${{ github.event.inputs.release_branch }}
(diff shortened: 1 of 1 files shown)


Xunzhuo (approved):
Looks good

jmazzitelli (approved):

PR #551, opened 2022-07-11
refactor large blob of code that converts yaml from snake case to camel case

jmazzitelli:
extract to a separate file to make it easier to maintain

This code was originally done as part of PR https://github.com/kiali/kiali-operator/pull/524 but we might not merge that, and because I didn't want to lose this refactoring, I am putting it in as this separate PR.


jmazzitelli:
This is enough to test this works. I ran it and all is OK:
`make -e MINIKUBE_PROFILE=ci -e CLUSTER_TYPE=minikube -e MOLECULE_SCENARIO="default affinity-tolerations-resources-test config-values-test" -e MOLECULE_USE_DEV_IMAGES="true" cluster-push molecule-test`

I also ran `make run-operator-playbook` and saw things run through the new task file. So all is good.

leandroberetta (approved):
LGTM

PR #552, opened 2022-07-13
remove perms no longer needed

jmazzitelli:
see https://issues.redhat.com/browse/OSSM-274


jmazzitelli:
You can see we no longer use this permission in any kiali roles by doing:

`grep -R "authentication.maistra.io" *` and `grep -R "rbac.maistra.io *`

These WERE used in the old 1.12 playbook - but the 1.12 playbook has now been deleted. This again you can confirm - look and compare at the 1.12 role.yaml openshift template versus the 1.48 template:

1.12 role where we did use these permissions:  https://github.com/kiali/kiali-operator/blob/v1.12/roles/default/kiali-deploy/templates/openshift/role.yaml#L62-L81

1.48 role where we no longer use these permissions (you will not see them in here, search for "maistra.io" to confirm): https://github.com/kiali/kiali-operator/blob/v1.48/roles/default/kiali-deploy/templates/openshift/role.yaml

leandroberetta (approved):
LGTM.

PR #558, opened 2022-08-19
For Tracing queries, add QueryScope support (analogous to Prometheus)

jshaughn:
https://github.com/kiali/kiali/issues/4635


jmazzitelli (approved):
This is OK, but I question the need to update the v1.48 role. Is this new query_scope support going to be backported to the v1.48 server? If not, you should remove this from the 1.48 role.  Otherwise, if backporting will happen, then you can merge as-is.

jshaughn:
> This is OK, but I question the need to update the v1.48 role

Good point, no planned backport, not sure why I updated that role.  I'll revert that change.

jmazzitelli (approved):

PR #562, opened 2022-08-25
remove seccompProfile

jmazzitelli:
part of https://github.com/kiali/kiali/issues/5409


jshaughn (approved):
I did not run through these tests but mazz did, and I read through the whole diatribe on the issue.  It seems necessary and looks visually good.

PR #565, opened 2022-08-29
Remove rhel8 from registry

aljesusg:
We need to remove the key rhel8 from registry


jmazzitelli (approved):
If this is going to be true starting today, then you should update this in the master branch, too.

aljesusg:
> If this is going to be true starting today, then you should update this in the master branch, too.

Yes...let check first if this is working as I expect. HoneyBadger team wanted to remove the "rhelX" key so...We'll need it

PR #569, opened 2022-09-13
Added apiGroups for K8s Gateway API gateway.networking.k8s.io

hhovsepy:
Required for https://github.com/kiali/kiali/pull/5345

### DIFF
--- a/manifests/kiali-community/1.57.0/manifests/kiali.v1.57.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.57.0/manifests/kiali.v1.57.0.clusterserviceversion.yaml
@@ -485,6 +485,7 @@ spec:
           - security.istio.io
           - extensions.istio.io
           - telemetry.istio.io
+          - gateway.networking.k8s.io
           resources: ["*"]
           verbs:
           - get
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -539,6 +539,7 @@ spec:
           - security.istio.io
           - extensions.istio.io
           - telemetry.istio.io
+          - gateway.networking.k8s.io
           resources: ["*"]
           verbs:
           - get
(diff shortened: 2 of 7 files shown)


manifests/kiali-upstream/1.56.0/manifests/kiali.v1.56.0.clusterserviceversion.yaml
```diff
@@ -485,6 +485,7 @@ spec:
           - security.istio.io
           - extensions.istio.io
           - telemetry.istio.io
+          - gateway.networking.k8s.io
```
jmazzitelli, line 488: we already released 1.56.0

jmazzitelli (changes requested):

hhovsepy:
thanks @jmazzitelli have fixed and rebased

jmazzitelli (approved):
this looks correct

manifests/kiali-community/1.57.0/manifests/kiali.v1.57.0.clusterserviceversion.yaml
```diff
@@ -485,6 +485,7 @@ spec:
           - security.istio.io
           - extensions.istio.io
           - telemetry.istio.io
+          - gateway.networking.k8s.io
```
nrfox, line 488: Does this throw errors if the gateway APIs are not deployed into the cluster? Since the gateway API CRDs are not required and bundled with Istio, I wonder if we first need to check if those types exist before granting permissions for them or if kube doesn't care if you grant permissions for a type that doesn't exist in the cluster yet.

hhovsepy: No error on a clean cluster without Gateway API CRD, Kiali installs and runs fine.

jmazzitelli: no - no APIs defined in roles like this are required to be deployed. This is why we can have one list that even contains openshift.io APIs (and it works on, say, minikube).

PR #570, opened 2022-09-14
Replace values by variables for build

aljesusg:
Well now we'll have these variables defined in our environment.

Let me check  if works correctly in my repo first


aljesusg:
ok, let me see it. 

I tested the change in my repo and works :)
I am bumping automatically the metadata version too !!! ...we will not need to change anything each build

jmazzitelli:
I see this PR is for the `ossm-latest` branch. If it works, we'll want to put this in `master` too

aljesusg:
Yep, I wanted to try a real test with a tag creation and if works we can make the changes in master

PR #572, opened 2022-09-22
support the use-case when istiod name is not "istiod"

jmazzitelli:


jmazzitelli:
I tested this with OSSM 2.2 and setting that istiod_deployment_name to 'istiod-custom'

jshaughn (approved):
LOK

PR #580, opened 2022-10-05
[molecule] use operator pod when querying prometheus

jmazzitelli:
We cannot use the kiali pod to perform the curl because the kiali pod container is distroless and does not have curl anymore. But we know the operator pod has it, and it can connect to prometheus.

### DIFF
--- a/molecule/common/query-prometheus.yml
+++ b/molecule/common/query-prometheus.yml
@@ -4,10 +4,11 @@
   when:
   - is_minikube == True
 
-- name: Query Prometheus from Kiali Pod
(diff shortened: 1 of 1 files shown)


jmazzitelli:
To test:
1. Start minikube and install Istio
2. `make -e CLUSTER_TYPE=minikube -e MOLECULE_SCENARIO="metrics-test" -e MOLECULE_USE_DEV_IMAGES="true" build cluster-push molecule-test`

jshaughn (approved):
Fine.

PR #582, opened 2022-10-21
be able to have the operator restrict accessible namespaces

jmazzitelli:
operator PR: https://github.com/kiali/kiali-operator/pull/582
helm PR: https://github.com/kiali/helm-charts/pull/175
doc PR: https://github.com/kiali/kiali.io/pull/593

PR #583, opened 2022-10-26
Be able to have the operator restrict accessible namespaces (v1.57 role)

jmazzitelli:
This is for the v1.57 role - just copying what is in the default role See: https://github.com/kiali/kiali-operator/pull/582 This will be cherry-picked over to the v1.57 branch.


jshaughn (approved):
Fine.

PR #585, opened 2022-10-28
secrets now injected into pod file system, not as env vars

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5591

### DIFF
--- a/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml
+++ b/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml
@@ -102,13 +102,6 @@ spec:
           value: "{{ kiali_vars.deployment.logger.sampler_rate }}"
         - name: LOG_TIME_FIELD_FORMAT
           value: "{{ kiali_vars.deployment.logger.time_field_format }}"
-{% for env in kiali_deployment_environment_variables %}
-        - name: {{ env }}
-          valueFrom:
-            secretKeyRef:
-              name: {{ kiali_deployment_environment_variables[env].secret_name }}
-              key: {{ kiali_deployment_environment_variables[env].secret_key }}
-{% endfor %}
         volumeMounts:
         - name: kiali-configuration
           mountPath: "/kiali-configuration"
@@ -116,6 +109,11 @@ spec:
           mountPath: "/kiali-secret"
         - name: kiali-cabundle
           mountPath: "/kiali-cabundle"
+{% for sec in kiali_deployment_secret_volumes %}
+        - name: {{ sec }}
+          mountPath: "/kiali-override-secrets/{{ sec }}"
+          readOnly: true
+{% endfor %}
 {% for secret in kiali_vars.deployment.custom_secrets %}
         - name: {{ secret.name }}
           mountPath: "{{ secret.mount }}"
@@ -138,6 +136,15 @@ spec:
         configMap:
           name: {{ kiali_vars.deployment.instance_name }}-cabundle
           optional: true
+{% for sec in kiali_deployment_secret_volumes %}
+      - name: {{ sec }}
+        secret:
+          secretName: {{ kiali_deployment_secret_volumes[sec].secret_name }}
+          items:
+          - key: {{ kiali_deployment_secret_volumes[sec].secret_key }}
+            path: value.txt
+          optional: false
+{% endfor %}
 {% for secret in kiali_vars.deployment.custom_secrets %}
       - name: {{ secret.name }}
         secret:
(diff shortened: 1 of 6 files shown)


jmazzitelli:
note [this k8s limitation](https://github.com/kubernetes/kubernetes/issues/81089) affects this PR.

jmazzitelli:
With this change, you should no longer see secrets injected into the pod as env vars via `secretKeyRef`. Instead you should see volume mounts for each secret. If all set, the yaml would look like:

```yaml
      volumeMounts:
      - mountPath: /kiali-override-secrets/prometheus-password
        name: prometheus-password
        readOnly: true
      - mountPath: /kiali-override-secrets/prometheus-token
        name: prometheus-token
        readOnly: true
      - mountPath: /kiali-override-secrets/tracing-password
        name: tracing-password
        readOnly: true
      - mountPath: /kiali-override-secrets/tracing-token
        name: tracing-token
        readOnly: true
      - mountPath: /kiali-override-secrets/grafana-password
        name: grafana-password
        readOnly: true
      - mountPath: /kiali-override-secrets/grafana-token
        name: grafana-token
        readOnly: true
      - mountPath: /kiali-override-secrets/login-token-signing-key
        name: login-token-signing-key
        readOnly: true
...
    volumes:
    - name: prometheus-password
      secret:
        defaultMode: 420
        items:
        - key: password
          path: value.txt
        optional: false
        secretName: mazz-secret
    - name: prometheus-token
      secret:
        defaultMode: 420
        items:
        - key: username
          path: value.txt
        optional: false
        secretName: mazz-secret
    - name: tracing-password
      secret:
        defaultMode: 420
        items:
        - key: password
          path: value.txt
        optional: false
        secretName: mazz-secret
    - name: tracing-token
      secret:
        defaultMode: 420
        items:
        - key: username
          path: value.txt
        optional: false
        secretName: mazz-secret
    - name: grafana-password
      secret:
        defaultMode: 420
        items:
        - key: password
          path: value.txt
        optional: false
        secretName: mazz-secret
    - name: grafana-token
      secret:
        defaultMode: 420
        items:
        - key: username
          path: value.txt
        optional: false
        secretName: mazz-secret
    - name: login-token-signing-key
      secret:
        defaultMode: 420
        items:
        - key: key
          path: value.txt
        optional: false
        secretName: kiali-signing-key
```

roles/default/kiali-deploy/templates/kubernetes/deployment.yaml
```diff
@@ -102,20 +102,18 @@ spec:
...
           mountPath: "/kiali-configuration"
         - name: kiali-secret
           mountPath: "/kiali-secret"
         - name: kiali-cabundle
           mountPath: "/kiali-cabundle"
+{% for sec in kiali_deployment_secret_volumes %}
+        - name: {{ sec }}
+          mountPath: "/kiali-override-secrets/{{ sec }}"
+          readOnly: true
+{% endfor %}
```
nrfox, line 116: Do these need to be separate volumes or can they be combined?

nrfox (dismissed):
Tested along with: https://github.com/kiali/kiali/pull/5592

roles/default/kiali-deploy/templates/kubernetes/deployment.yaml
jmazzitelli: Because these are separate secrets, I believe they need to be mounted separately. While I did not actually try it, it looks like the volumes need to be in their own locations. For example, I don't think we can put everything in mountPath "/kiali-override-secrets" because the keys in multiple secrets could conflict and thus cause a problem since the filenames would be the same (or we'd have to somehow figure out a way to create non-conflicting names). For example, if you had two secrets and both have a key "password" in them, if they are both mounted in the same /kiali-override-secrets path, we have no easy way to avoid that conflict. The way this is now, it is easy to configure without worry about conflicts and it ensures each secret is separately mounted in their own location.

jmazzitelli: for the record, I just tried it to see what would happen if I had multiple secrets mounted on the same mountPath (I set all the mountPath settings to "/kiali-override-secrets") but make sure I set all the path values to a unique name (to avoid name conflicts) ... it appears the last one wins. I was expecting to see 7 files, but I only see 1.

jmazzitelli: looks like there is this "projected" feature that might allow to consolidate this all - but I don't see a need to investigate this any further. This is a hidden implementation detail that should not concern the user and it is fine having subdirectories for each secret. Not much benefit here by consolidating this further.

nrfox (approved):

PR #589, opened 2022-11-07
deprecated api updated

krouser:
deprecated ingress api updated

### DIFF
--- a/roles/v1.24/kiali-deploy/templates/kubernetes/ingress.yaml
+++ b/roles/v1.24/kiali-deploy/templates/kubernetes/ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: kiali


jmazzitelli:
Please ensure any submitted PR has an associated github issue for it - the issue needs to document the problem the PR is attempting to solve. The Kiali team will help assess the issue and determine if a fix is warranted.

Before submitting proposed code changes, it is best to first ask a question in Kiali github Q&A Discussion forum. See: https://github.com/kiali/kiali/discussions/new?category=q-a

In this case, this PR has several problems:
1. It is changing the `v1.24` role which is not maintained/supported for upstream community consumption. See the [Kiali support matrix]([url](https://kiali.io/docs/installation/installation-guide/prerequisites/#version-compatibility)) - v1.24 was for Istio 1.7 and that version of Istio has long been out of support.
2. Only the `default` ansible role is supported (the `spec.version` of the Kiali CR must be "default" when installing for upstream Istio installations - see [the Kiali docs](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.version) where it says "The only supported value today is default.").
3. This PR only changes the `kubernetes` but not the `openshift` template.
4. There is no associated PR change for the Kiali server helm chart.

In the future, any Kiali resource changes a PR suggests should be done following the developer checklist found here: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-resource

I'm closing this PR as it is incomplete and we are not updating the v1.24 ansible role regardless since it is no longer maintained and no supported upstream Istio version works with the Kiali installed via the v1.24 role.

Regarding the issue of the deprecated Ingress API - there was a previous issue already created where this was addressed. See this issue https://github.com/kiali/kiali/issues/3706 and its 3 related PRs documented in that issue's description text. (there actually was an even earlier [issue](https://github.com/kiali/kiali/issues/3343), but the Ingress API changed again after that which resulted in that second change).

If you are still installing the old v1.24 Kiali but on a newer K8s version that supports the new API, the suggested solution is to create your own Ingress and tell Kiali not to install its own Ingress via the old v1.24 `spec.ingress_enabled` setting set to `false`. But I would question if Kiali 1.24 and Istio 1.7 would even install/run properly on the newer k8s versions (it may, I just don't think that was tested).

PR #595, opened 2022-11-30
Add new configuration option istio_api_enabled

josunect:
REF https://github.com/kiali/kiali/pull/5642

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -277,6 +277,7 @@ spec:
       envoy_admin_local_port: 15000
       gateway_api_class_name: ""
       # default: istio_canary_revision is undefined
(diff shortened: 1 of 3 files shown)


crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -354,6 +354,8 @@ spec:
     # default: private_key_file is undefined
     private_key_file: ""
 
+  istio_api_enabled: true
```
jmazzitelli, line 357: this doesn't seem to need to be a whole new top-level section. This should go under `external_services.istio` with the rest of the istio-specific configuration settings.

josunect: Thanks for your feedback, @jmazzitelli ! I will push a new commit.

jmazzitelli (approved):

PR #602, opened 2023-01-10
Remove kiali_1_24

aljesusg:


jmazzitelli (approved):
We should also remove the v1.24 role from both `master` and `v1.57` branches as well:

These:
https://github.com/kiali/kiali-operator/tree/v1.57/roles/v1.24
https://github.com/kiali/kiali-operator/tree/default/roles/v1.24

aljesusg:
> We should also remove the v1.24 role from both `master` and `v1.57` branches as well:
> 
> These: https://github.com/kiali/kiali-operator/tree/v1.57/roles/v1.24 https://github.com/kiali/kiali-operator/tree/master/roles/v1.24

Ok, I didn't know if we are using these roles for upstream releases. I am going to do it

jmazzitelli:
> I didn't know if we are using these roles for upstream releases. I am going to do it

Upstream releases only support `default` role (i.e. the latest-n-greatest).

We [document that here](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.version) where it says, "The only supported value today is `default`"

PR #610, opened 2023-01-27
Add a new configuration for jaeger query timeout

RaiAnandKr:
Kiali's query to 'jaeger-query' service for fetching traces times out occasionally and the timeout value is hardcoded.
Add a new configuration: 'query_timeout' to enable users to configure that timeout.

Closes https://github.com/kiali/kiali/issues/5751

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -332,6 +332,7 @@ spec:
       query_scope:
         mesh_id: "mesh-1"
         cluster: "cluster-east"
+      query_timeout: 5
       url: ""
       use_grpc: true
       whitelist_istio_system: ["jaeger-query", "istio-ingressgateway"]
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -220,6 +220,7 @@ kiali_defaults:
       is_core: false
       namespace_selector: true
       query_scope: {}
+      query_timeout: 5
       url: ""
       use_grpc: true
       whitelist_istio_system: ["jaeger-query", "istio-ingressgateway"]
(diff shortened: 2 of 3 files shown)


jmazzitelli (approved):
This looks good to me.

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -220,6 +220,7 @@ kiali_defaults:
       is_core: false
       namespace_selector: true
       query_scope: {}
+      query_timeout: 4
```
jshaughn, line 223: This should match the timeout set in `crd-docs/cr/kiali.io_v1alpha1_kiali.yaml`.   The current hardcoded value is 4s for grpc, which is the default api we use.   For some reason we have 5s for http.  And you are looking for a longer setting for your env.  How about if we just set the default to 5s.  I don't want to change from 4s to 10s as part of this change, let's just add the configurability.

jshaughn (changes requested):
Need to make one change and then this should be good.   I am going to generate a kiali/kiali PR that actually uses the setting.  This should all make it into v1.64, scheduled for release on Feb 20.

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -332,6 +332,7 @@ spec:
       query_scope:
         mesh_id: "mesh-1"
         cluster: "cluster-east"
+      query_timeout: 10
```
jmazzitelli, line 335:
```suggestion
      query_timeout: 5
```

roles/default/kiali-deploy/defaults/main.yml
jmazzitelli, line 223:
```suggestion
      query_timeout: 5
```

jmazzitelli (changes requested):
commit suggestions attached to this review - move the defaults to 5s to conform to the desired default (which also matches the 5s over in the server PR).

roles/default/kiali-deploy/defaults/main.yml
jshaughn: Thanks, I don't know why we had 4s for gRPC and 5s for HTTP.  But because you were hitting timeouts I thought the higher value would be better as the default for both.   My understanding is that the values in `crd-docs/cr/kiali.io_v1alpha1_kiali.yaml` should reflect the actualy defaults in `roles/default/kiali-deploy/defaults/main.yml`.  We may have to take a look and audit the crd-docs file, but not in the PR.

jmazzitelli (approved):

jshaughn (approved):
LG

PR #612, opened 2023-01-30
[v1.63] the 1.63.1 manifest - this is the first release for istio 1.17

jmazzitelli:
part of https://github.com/kiali/kiali/issues/5773

this is for the v1.63 branch for when we build 1.63.1. We need this also in the master branch

### DIFF
--- a/manifests/kiali-community/1.63.1/bundle.Dockerfile
+++ b/manifests/kiali-community/1.63.1/bundle.Dockerfile
@@ -1,0 +1,11 @@
+FROM scratch
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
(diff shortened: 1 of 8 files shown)


leandroberetta:
@jmazzitelli isn't the pipeline creates this files? Or that is just for minor releases?.

leandroberetta:
> @jmazzitelli isn't the pipeline creates this files? Or that is just for minor releases?.

I checked and only generates manifests for minor releases

jshaughn (commented):
LOK

jmazzitelli:
Yeah, @leandroberetta this needs to have the 1.63.1 metadata in the branch because when I produce the actual OLM PRs, I do it from the branch. so these need to be in there

PR #615, opened 2023-01-31
Remove the kubernetes_config.cache_enabled configuration

israel-hdez:
The kiali back-end will no longer honor the kubernetes_config.cache_enabled configuration (kube cache is going to be mandatory). Thus, this is removing all references in the operator to such configuration, as it will no longer be valid.

Related issue kiali/kiali#5599
Related server PR: kiali/kiali#5779


jmazzitelli (approved):
lgtm

PR #616, opened 2023-02-01
OSSM-3317 allow for cluster-wide setting of **

jmazzitelli:


jshaughn (approved):
This is fine but do we even need this env var?

jmazzitelli:
> This is fine but do we even need this env var?

we need it because its a security valve. We want admins to be given the ability to disable someone asking to install a Kiali with full cluster permissions. see helm chart, too: https://github.com/kiali/helm-charts/blob/master/kiali-operator/values.yaml#L77-L82

security, security, security!

PR #622, opened 2023-02-23
Add new config option skip_wildcard_gateway_hosts

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5711

Added new option "skip_wildcard_gateway_hosts" to skip "KIA0301 More than one Gateway for the same host port combination" validation for Gateways with wildcard in hostname.

Related core PR https://github.com/kiali/kiali/pull/5869

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -415,6 +415,7 @@ spec:
       refresh_interval: "60s"
     validations:
       ignore: ["KIA1201"]
+      skip_wildcard_gateway_hosts: false
 
   kubernetes_config:
     burst: 200
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -779,6 +779,9 @@ spec:
                       root_namespace:
                         description: "The namespace to treat as the administrative root namespace for Istio configuration."
                         type: string
+                      skip_wildcard_gateway_hosts:
+                        description: "To skip KIA0301 validation for Gateways with wildcards in hosts."
+                        type: boolean
                       url_service_version:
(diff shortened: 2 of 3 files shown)


crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -415,6 +415,7 @@ spec:
       refresh_interval: "60s"
     validations:
       ignore: ["KIA1201"]
+      skip_wildcard_gateway_hosts: false
```
jmazzitelli, line 418: Is there a reason why we add a separate config and not just tell people to add  KIA301 to the ignore list above? This seems confusing to allow to ignore validations by ID but then add a setting for a specific setting.

hhovsepy: There is a community user request to keep the validations of KIA0301 but to ignore them for wildcard only hostnames: https://github.com/kiali/kiali/issues/5711
As Istio itself is not conflicting with wildcard hosts but leaving them as a last Gateway to be used.

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -779,6 +779,9 @@ spec:
                       root_namespace:
                         description: "The namespace to treat as the administrative root namespace for Istio configuration."
                         type: string
+                      skip_wildcard_gateway_hosts:
+                        description: "To skip KIA0301 validation for Gateways with wildcards in hosts."
```
jmazzitelli, line 783: This needs a lot more detail to explain what this is all about. It is confusing because one wonders why I can't just add KIA301 to the ignore list. So explain what this setting does and when you would want to set it.

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
hhovsepy: Adding a new validation code would make a confusion in two quite similar codes, so I decided to make a small change in existing one. Agree to have it well documented in crd schema.

PR #620, opened 2023-02-17
use k8s_cluster_info to get api version availability from the cluster

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5857

PR #626, opened 2023-03-13
Add 1.65 version and removed 1.36

aljesusg:
Adding roles and 1.65 version and removing 1.36

Must we removed 1.36 in 2.4 right? 
This means that we are going to support 2.2 (1.48), 2.3(1.57) and 2.4(1.65)

Verify this please @jshaughn @jmazzitelli 

I am going to make the changes in the branch


jmazzitelli (approved):
This looks correct.

I do not know if 1.36 should be removed or not.

jmazzitelli:
Something happened with this PR that isn't correct.

@aljesusg when you created this v1.65 role, do you remember how you did it? Because I think something went wrong.

What is supposed to happen is you copy the "default" role and name the new role "v1.65" - but something went wrong because we are missing code.

This went in a month before into the default role - https://github.com/kiali/kiali-operator/pull/620/files
But this code is missing (at least the code for the task `Set default HPA api_version` - I did not check what else is missing).

I wrote [an issue to fix this](https://github.com/kiali/kiali/issues/6201) - but we need to figure out how this happened so we can avoid this in the future.

PR #627, opened 2023-03-13
Add role in 1.65 and remove 1.36

aljesusg:
Adding 1.65 and removing 1.36 version for ossm 2.4


jmazzitelli (approved):
This looks good.

I do not know if 1.36 should be removed or not.

PR #631, opened 2023-03-29
fix the typo in the new ansible setting name

jmazzitelli:
The new setting name is `callbacks_enabled` (in the plural).

Also keep the deprecated name around since currently we are using an ansible version that doesn't know about the new one.

fixes https://github.com/kiali/kiali/issues/4338


jmazzitelli:
i confirmed molecule tests can still pass with this (I'm not even sure if we use the junit plugin in our tests but whatev).

PR #633, opened 2023-04-04
use include_tasks, no longer use deprecated include

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5931


jmazzitelli:
ran the "default" molecule test - all worked. So we know the change is valid.

jshaughn (approved):
LOK as long as this is OK for any version the operator will deal with, which I guess is fine for upstream.

PR #634, opened 2023-04-11
fix the CRD schema and have validation tool print out errors

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6015

To test this, start a cluster and in the operator repo, run this make target: `make validate-cr` -- it should now pass


jshaughn (approved):
LOK

jshaughn (approved):
Ran against the updated Ci and looks OK.

PR #642, opened 2023-05-01
make the namespace-detected playbook faster

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/6081

### DIFF
--- a/playbooks/kiali-new-namespace-detected.yml
+++ b/playbooks/kiali-new-namespace-detected.yml
@@ -15,21 +15,11 @@
   #
   # This playbook is essentially a no-op when there are no Kiali CRs installed in the cluster.
 
-  - name: Get all Kiali CRs
-    ignore_errors: yes
-    k8s_info:
-      api_version: kiali.io/v1alpha1
-      kind: Kiali
-    register: kiali_cr_list
-
-  # end very quickly and quietly if there are no Kiali CRs in the cluster - there is nothing for us to do in that case
-  - meta: end_play
-    when:
-    - kiali_cr_list is not defined or kiali_cr_list.resources is not defined or kiali_cr_list.resources | length == 0
-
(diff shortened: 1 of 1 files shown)


jmazzitelli:
This makes it a little faster. But if you have the normal 70 or so namespaces in an out of box openshift, it still takes a while to process them all. I don't think there is anything we can do - because this is just the time it takes to spawn off an ansible playbook run.

jmazzitelli:
How to test (these assume minikube, but you can try it on OpenShift, just set the `-e` options accordingly):

1. Create a cluster - no need to have Istio in it, we will only test the operator startup.
2. Make sure you check out this PR branch in your local operator repo.
3. Create and push the images: `make -e CLUSTER_TYPE=minikube build build-ui cluster-push`
4. Install the operator with the profiler enabled, and immediately start watching the operator logs:
```
make -e OPERATOR_PROFILER_ENABLED=true -e CLUSTER_TYPE=minikube operator-create \
  && kubectl logs -n kiali-operator -l app.kubernetes.io/name=kiali-operator -f
```
Just make sure no errors  occur. You can look at the profiler results and see how fast each namespace is processed at startup. e.g. processing this namespace took 0.19s, with the ansible playbook taking an overall .287s (this is on minikube).

```
Wednesday 03 May 2023  14:08:33 +0000 (0:00:00.194)       0:00:00.287 ********* 
=============================================================================== 
Touch the Kiali CRs that were created before the new namespace [oauth2-proxy] --- 0.19s
```

jmazzitelli:
For comparison to see how much faster it goes with this PR, here's results of processing a namespace with master build:

```
Wednesday 03 May 2023  15:11:24 +0000 (0:00:01.405)       0:00:01.523 ********* 
=============================================================================== 
Get all Kiali CRs ------------------------------------------------------- 1.30s
```

So 1.3 seconds vs. 0.19s.

PR #644, opened 2023-05-11
Add new server config clusterName

josunect:
Fixes https://github.com/kiali/kiali/issues/6137

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1200,7 +1200,7 @@ spec:
                     description: "This Kiali cache is a list of namespaces per user. This is typically a short-lived cache compared with the duration of the namespace cache defined by the `cache_duration` setting. This is specified in seconds."
                     type: integer
                   cluster_name:
-                    description: "The cluster name. It is usually set up in multi cluster environments. By default, it will try to auto detect the cluster name from the Istiod deployment."
+                    description: "The name of the cluster Kiali is deployed in. This is only used in multi cluster environments. If not set, Kiali will try to auto detect the cluster name from the Istiod deployment or use the default 'Kubernetes'."
                     type: string
                   excluded_workloads:
                     description: "List of controllers that won't be used for Workload calculation. Kiali queries Deployment, ReplicaSet, ReplicationController, DeploymentConfig, StatefulSet, Job and CronJob controllers. Deployment and ReplicaSet will be always queried, but ReplicationController, DeploymentConfig, StatefulSet, Job and CronJobs can be skipped from Kiali workloads queries if they are present in this list."
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1200,7 +1200,7 @@ spec:
                     description: "This Kiali cache is a list of namespaces per user. This is typically a short-lived cache compared with the duration of the namespace cache defined by the `cache_duration` setting. This is specified in seconds."
                     type: integer
                   cluster_name:
-                    description: "The cluster name. It is usually set up in multi cluster environments. By default, it will try to auto detect the cluster name in the Istiod deployment."
+                    description: "The cluster name. It is usually set up in multi cluster environments. By default, it will try to auto detect the cluster name from the Istiod deployment."
                     type: string
                   excluded_workloads:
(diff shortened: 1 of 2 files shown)


roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -315,6 +315,7 @@ kiali_defaults:
     cache_namespaces:
     - ".*"
     cache_token_namespace_duration: 10
+    cluster_name: ""
```
jmazzitelli, line 318: Is this cluster name defined by Istio? The docs say it can be auto-discovered via istiod, so I assume this is an Istio setting. If so, it should be considered an istio configuration, not a kubernetes_config setting (i.e. if this is an Istio-specific cluster name, it should be `external_services.istio.cluster_name`.)

josunect: Yes, that's right. And looking into this: https://github.com/kiali/kiali/blob/master/kiali.go#L239 and this discussion https://github.com/kiali/kiali/pull/6019#discussion_r116322873 looks like it is just discovered from Istiod, so probably it would be better into external_services.istio

I will ping @hhovsepy and @leandroberetta in case they have any objection, as it will also require changes in the kiali server.

hhovsepy: Is the cluster name Istio specific or it is cluster specific before Istio is installed?
If is it Istio specific, probably we need a method of getting this config instead of directly accessing to in from everywhere config.Get().KubernetesConfig.ClusterName (a bunch of places).

josunect: So now, what it is doing, is get the cluster name from the configuration, but, if it is empty, it tries to get the name from the Istio Config. If it fails, it uses the default cluster name. But it always sets the name into the config, so it is always available in the same place. 
It uses the config one because the Istio cluster name is only in the deployment as an env var, so if Istio is not available when Kiali starts, it will fail.

josunect: I was going to update this but I see this specific comment: 

`// ClusterName is the name of the kubernetes cluster that Kiali is running in.`

https://github.com/kiali/kiali/blob/master/config/config.go#L305 

So maybe it makes more sense to leave this setting in the kubernetes config. Because, the name could come from Istio, but maybe don't, in the case for example of a primary remote, In case of Istio the cluster name should be specified in the control plane but if this is not present, i think It should be the same as the kubernetes cluster name? 

image

jmazzitelli: That is true - the remote cluster names come from the remote secrets, not from any Istio API.

Whatever people think is best. I just want to avoid seeing so many config settings all over the place. But if it makes more sense under kubernetes, then leave it there.

josunect: I think it would make more sense under kubernetes. Anyway, I will also ping @nfox to confirm.

nrfox: It was put under `kubernetes_config` because it is the name of the kubernetes cluster. Kubernetes, afaik, has no concept of cluster name but Istio needs some identifier for the clusters. So it could probably go under either the kube config or istio since it's related to both and I chose the kube config since that was where I'd expect to find it.

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1199,6 +1199,9 @@ spec:
                   cache_token_namespace_duration:
                     description: "This Kiali cache is a list of namespaces per user. This is typically a short-lived cache compared with the duration of the namespace cache defined by the `cache_duration` setting. This is specified in seconds."
                     type: integer
+                  cluster_name:
+                    description: "The cluster name. It is usually set up in multi cluster environments. By default, it will try to auto detect the cluster name from the Istiod deployment."
```
nrfox, line 1203:
```suggestion
                    description: "The name of the cluster Kiali is deployed in. This is only used in multi cluster environments. If not set, Kiali will try to auto detect the cluster name from the Istiod deployment or use the default 'Kubernetes'."
```

roles/default/kiali-deploy/defaults/main.yml
nrfox: It could also go under `spec.deployment` since it's really the name of the cluster that Kiali is deployed in?

josunect: Probably it would be more obvious to find in the Kubernetes config. But I don't have a strong opinion about this. 

I've taken some notes that I think it would be important to include in the docs - following the discussion in slack: 
- The cluster name should be set when Kiali is deployed in a cluster in the Mesh and Kiali does not rely on Istio to auto discover the cluster name (May fail is istiod is restarting). 
- Should match what it is set in Istio so it can match the Telemetry
- For the remote clusters, Kiali reads the secret set from the kubeconfig.

nrfox (approved):

PR #655, opened 2023-05-30
fix bug that caused error when only one consolelink was to be removed

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6216

PR #656, opened 2023-05-30
[v1.65] fix bug that caused error when only one consolelink was to be removed

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/6216

cherry-pick of: https://github.com/kiali/kiali-operator/pull/655


aljesusg (approved):
LGFM

PR #659, opened 2023-06-06
default to HPA v2 which is supported on OpenShift 4.10+

jmazzitelli:
fixes https://issues.redhat.com/browse/OSSM-4069


jshaughn (approved):
LOK

PR #666, opened 2023-07-10
fix remote cluster secret label default

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6344

The default is defined here - you can see it is "kiali.io" not "istio" prefix: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L251


leandroberetta (approved):
LG!

PR #677, opened 2023-08-29
ARM identifier is arm64

jmazzitelli:
The arch name is based on GOARCH, so it has to be `arm64`:

https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/


jshaughn (approved):
LOK

PR #684, opened 2023-09-05
Added support of multiple Gateway API classes.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/6429a

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -745,9 +745,18 @@ spec:
                       envoy_admin_local_port:
                         description: "The port which kiali will open to fetch envoy config data information."
                         type: integer
-                      gateway_api_class_name:
-                        description: "The K8s Gateway API GatewayClass's Name used in Istio. If empty, the default value 'istio' is used."
-                        type: string
+                      gateway_api_classes:
+                        description: "The K8s Gateway API GatewayClass's implementation name and class name used in Istio. If empty, the default value 'Istio/istio' is used."
+                        type: array
+                        items:
+                          type: object
+                          properties:
+                            name:
+                              description: "The name of the Gateway API implementation."
+                              type: string
+                            class_name:
+                              description: "The name of the GatewayClass."
+                              type: string
                       istio_api_enabled:
                         description: "Indicates if Kiali has access to istiod. true by default."
                         type: boolean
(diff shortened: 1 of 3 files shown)


jmazzitelli:
This breaks backward compatibility. Are we sure we want to do this?

hhovsepy:
Before this PR, the only supported GatewayClass was "istio", which was the default if no configuration touched, but now, when several implementations of Gateway should be supported, if still no configuration touched, the default is "istio". So we need this new way of config.

jmazzitelli:
> Before this PR, the only supported GatewayClass was "istio", which was the default if no configuration touched, but now, when several implementations of Gateway should be supported, if still no configuration touched, the default is "istio". So we need this new way of config.

OK. Sounds like this was rarely touched by a user. We should just document this change in the release notes (I think there is a "Breaking change" section somewhere - or something like that) where we should mention it.

jshaughn:
> OK. Sounds like this was rarely touched by a user. We should just document this change in the release notes (I think there is a "Breaking change" section somewhere - or something like that) where we should mention it.

+1, we can document the config change.  But because the previous attribute supported only one value, the default, it should be a no-op to replace it.

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -745,9 +745,18 @@ spec:
                       envoy_admin_local_port:
                         description: "The port which kiali will open to fetch envoy config data information."
                         type: integer
-                      gateway_api_class_name:
-                        description: "The K8s Gateway API GatewayClass's Name used in Istio. If empty, the default value 'istio' is used."
-                        type: string
+                      gateway_api_classes:
+                        description: "The K8s Gateway API GatewayClass's implementation name and class name used in Istio. If empty, the default value 'Istio/istio' is used."
```
jmazzitelli, line 749: Since this is now an array, we should re-word this to make that clear.
```suggestion
                        description: "A list declaring all the Gateways used in Istio. If left empty or undefined, the default is a single list item whose name is `Istio` and class_name is `istio`."
```

jmazzitelli (changes requested):

jshaughn (commented):
Looks good, new config was accepted:

jshaughn (approved):

jmazzitelli (approved):

PR #690, opened 2023-10-06
Fix error in jaeger test

josunect:
Fix error https://github.com/kiali/kiali/actions/runs/6316367574/job/17151097502
```
TASK [Assert that there is an error related to Jaeger] *************************
fatal: [localhost]: FAILED! => {
    "assertion": "jaeger_error is defined",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

Ref. https://github.com/kiali/kiali/issues/6537

### DIFF
--- a/molecule/jaeger-test/converge.yml
+++ b/molecule/jaeger-test/converge.yml
@@ -27,7 +27,7 @@
     set_fact:
       tracing_error: "{{ item.status }}"
     loop: "{{ status_response.json }}"
-    when: "item.name == 'jaeger'"
+    when: "item.name == 'tracing'"
   
   - name: Assert that there is no error related to Jaeger
     assert:
--- a/molecule/jaeger-test/converge.yml
+++ b/molecule/jaeger-test/converge.yml
@@ -25,14 +25,14 @@
   # If Jaeger is in the list, it means that there is an error (Unreacheable)
   - name: Check if Jaeger status is present
     set_fact:
-      jaeger_error: "{{ item.status }}"     
+      tracing_error: "{{ item.status }}"
     loop: "{{ status_response.json }}"
     when: "item.name == 'jaeger'"
   
   - name: Assert that there is no error related to Jaeger
     assert:
       that:
-      - jaeger_error is not defined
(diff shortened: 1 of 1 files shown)


molecule/jaeger-test/converge.yml
```diff
@@ -25,14 +25,14 @@
   # If Jaeger is in the list, it means that there is an error (Unreacheable)
   - name: Check if Jaeger status is present
     set_fact:
-      jaeger_error: "{{ item.status }}"     
+      tracing_error: "{{ item.status }}"
     loop: "{{ status_response.json }}"
     when: "item.name == 'jaeger'"
```
jmazzitelli, line 30:
```suggestion
    when: "item.name == 'tracing'"
```

molecule/jaeger-test/converge.yml
```diff
@@ -57,11 +57,11 @@
 
   - name: Check if Jaeger status is present
     set_fact:
-      jaeger_error: "{{ item.status }}"     
+      tracing_error: "{{ item.status }}"
     loop: "{{ status_response.json }}"
     when: "item.name == 'jaeger'"
```
jmazzitelli, line 62:
```suggestion
    when: "item.name == 'tracing'"
```

jmazzitelli (changes requested):
Changing the ansible variable names doesn't actually fix anything :)

See my suggested changes that should fix the tests.

josunect:
> Changing the ansible variable names doesn't actually fix anything :)
> 
> See my suggested changes that should fix the tests.

Oh, I should have read a bit more about ansible first ... I thought it was related to a type that is named also jaeger_error and was renamed...

jmazzitelli (approved):
I haven't had a change to run this, but this looks correct.

We'll know it works when the  CI "Test Molecule" action runs ;)

josunect:
> I haven't had a chance to run this, but this looks correct.
> 
> We'll know it works when the CI "Test Molecule" action runs ;)

So, I'll merge then :)

PR #703, opened 2023-10-25
Fix upgrade with no label_selector_include (#6778)

sylvainOL:
fixes: kiali/kiali#6778

When `label_selector_include` is not set in Kiali CR, upgrade still tries to patch some namespaces as variable used to retrieve it has empty string as value.

Make sure it's not the case anymore

### DIFF
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -788,6 +788,7 @@
   - current_cluster_wide_access is defined
   - current_cluster_wide_access|bool == False
   - current_label_selector_include is defined
+  - current_label_selector_include != ""
   - kiali_vars.api.namespaces.label_selector_include is defined
   - current_label_selector_include != kiali_vars.api.namespaces.label_selector_include
 
@@ -884,6 +885,7 @@
   when:
   - no_longer_accessible_namespaces is defined
   - current_label_selector_include is defined
+  - current_label_selector_include != ""
 
 - name: Create additional Kiali label on all accessible namespaces
   vars:


jmazzitelli:
@sylvainOL this looks good! Thanks!

jmazzitelli:
I'm going to look more closely at the code and test this manually to make sure there isn't something else wrong here that is  being missed. I'll merge this when I'm confident nothing else is wrong.

roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -884,6 +884,7 @@
   when:
   - no_longer_accessible_namespaces is defined
   - current_label_selector_include is defined
+  - current_label_selector_include != ""
```
jmazzitelli, line 887: Can you make this same change after line 790 as well? That is the only thing I see missing that would make this PR complete. So this:

```
line 789:  - current_cluster_wide_access|bool == False
line 790:  - current_label_selector_include is defined
           - current_label_selector_include != ""          <<<===== add this line
line 791:  - kiali_vars.api.namespaces.label_selector_include is defined
```

I'd add that as a github suggestion, but github won't let me add a comment on a line that isn't changed in the PR.

Thanks.

jmazzitelli (changes requested):
One addition is needed. If we get this in now, we can merge this and get it into the release that will be published this Monday.

roles/default/kiali-deploy/tasks/main.yml
sylvainOL: I can make this change (I actually thought about it) but this will change the behavior of how it's done today (which may not be what the behavior you should want):

Today, I you set a `label_selector_include` and then remove it, the task `Do not allow user to change label_selector_include` will fail.

If we add `current_label_selector_include != "" ` at the list of conditions, it won't be the case.

in case `label_selector_include` is not set and it is on `kiali_vars.api.namespaces.label_selector_include`, the it will fail (comparison between empty string and a string) and if kiali_vars.api.namespaces.label_selector_include` is not set, line 791 will do the trick.

So if you think it's the way to go, I'm OK to add the line but I just wanted you to understand that the behavior will change when  `label_selector_include` was set and is removed (before it'll fail, after it'll pass).

sylvainOL:
I've also change the title because I mixed  `label_selector_include` and  `label_selector_exclude` sorry

roles/default/kiali-deploy/tasks/main.yml
jmazzitelli: I think it is OK. The reason I think that is that it is more likely a person will not have set that value at first, and then decide later they want their own label. I think the reverse is less-likely (it is less likely a person will set their own label but then later decide they don't want it and remove it to pick up the default Kiali label). So let's code this up for the most likely scenario.

If it really happens that a person does the less likely scenario, the workaround is simply to delete the Kiali CR, let the operator uninstall Kiali fully, and then install the new Kiali CR. That's the basic workaround for this bug anyway. So that person isn't dead in the water, they can workaround it.

sylvainOL: ok, so I add the line too!

thanks

sylvainOL: done

jmazzitelli (approved):

jmazzitelli:
Thanks, @sylvainOL !

sylvainOL:
no problem, I should have done sooner :)

PR #704, opened 2023-10-25
if OS console is not enabled, then either skip doing things or fail with a human readable error

jmazzitelli:
fixes: https://issues.redhat.com/browse/OSSM-2134

PR #705, opened 2023-10-25
[1.73] if OS console is not enabled, then either skip doing things or fail with a human readable error

jmazzitelli:
fixes: https://issues.redhat.com/browse/OSSM-2134

cherry pick of https://github.com/kiali/kiali-operator/pull/704


jshaughn (approved):
Visual check looks like a straight cherry-pick.  Did not test.

PR #717, opened 2023-11-02
explicitly document the version fields for display in the OS Console UI

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6813

this also corrects some issues with other metadata


jshaughn (approved):
LOK

PR #723, opened 2023-11-03
purge maistra related stuff

jmazzitelli:
part of https://github.com/kiali/kiali/issues/6816

see also: https://github.com/kiali/kiali/pull/6819

to test, see test procedures [here](https://github.com/kiali/kiali/pull/6819#issuecomment-1805693667)


jshaughn (approved):
This looks fine.   Do we have any maistra-specific content on kiali.io (other than the compatibility chart)?

jmazzitelli:
> This looks fine. Do we have any maistra-specific content on kiali.io (other than the compatibility chart)?

Only other place is https://kiali.io/docs/configuration/namespace-management/#accessible-namespaces
```
Maistra supports multi-tenancy and the accessible_namespaces extends that feature to Kiali. However, explicit naming of accessible namespaces can benefit non-Maistra installations as well - with it Kiali does not need cluster roles and the Kiali Operator does not need permissions to create cluster roles.
```

jshaughn:
> Only other place is https://kiali.io/docs/configuration/namespace-management/#accessible-namespaces

I think we can remove this blurb as well, but I can do it as part of the next release notes PR, which also has to update a couple of other things.  This will remain in the 1.73 version of the docs (and all other maistra release versions of the docs).

jmazzitelli:
> > Only other place is https://kiali.io/docs/configuration/namespace-management/#accessible-namespaces
> 
> I think we can remove this blurb as well, but I can do it as part of the next release notes PR, which also has to update a couple of other things. This will remain in the 1.73 version of the docs (and all other maistra release versions of the docs).

I created a PR for this already, let it be attached to this issue:

https://github.com/kiali/kiali.io/pull/719

PR #728, opened 2023-11-28
Add new settings for the otel collector

josunect:
Ref. [#6572](https://github.com/kiali/kiali/issues/6572) 
Ref. https://github.com/kiali/kiali/pull/6886

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1298,7 +1298,7 @@ spec:
                             type: object
                             properties:
                               ca_name:
-                                description: "The name of the ca name containing when `tls_enabled` is `true` and `skip_verify` is `false`."
+                                description: "The name of the CA cert; this is used when `tls_enabled` is `true` and `skip_verify` is `false`."
                                 type: string
                               protocol:
                                 description: "Protocol. Supported values are: `http`, `https` or `grpc`."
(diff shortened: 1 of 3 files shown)


crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1297,9 +1297,18 @@ spec:
...
                             type: object
                             properties:
+                              ca_name:
+                                description: "The name of the ca name containing when `tls_enabled` is `true` and `skip_verify` is `false`."
+                                type: string
                               protocol:
                                 description: "Protocol. Supported values are: `http`, `https` or `grpc`."
                                 type: string
+                              skip_verify:
+                                description: "Skip tls verification when using self signed certificates. Insecure option recommended just for testing."
```
jmazzitelli, line 1307: This doesn't just skip verification only if using self-signed certs. If you set this to true, NO cert verification will be done, regardless of where the CA came from. We should change this to make that more clear.
Also, capitalize TLS.

```suggestion
                                description: "If true, TLS certificate verification will not be performed. This is an unsecure option and is recommended only for testing."
```

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1297,9 +1297,18 @@ spec:
...
+                                type: boolean
+                              tls_enabled:
+                                description: "Enable tls for the collector. It must be used when `protocol` is `https` or `grpc`. It needs to set `ca_name` or `skip_verify`."
```
jmazzitelli, line 1310: Capitalize TLS and tell the reader it "must be *specified*".  Also, be more clear about what else is required.

```suggestion
                                description: "Enable TLS for the collector. This must be specified when `protocol` is `https` or `grpc`. When you set this to `true`, you must also set a `ca_name` or set `skip_verify` to true."
```

jmazzitelli (approved):
lgtm

PR #731, opened 2023-12-12
while processing OSSMConsole CR, operator should retry a few times before giving up getting the kiali version

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6950

### DIFF
--- a/roles/default/ossmconsole-deploy/tasks/main.yml
+++ b/roles/default/ossmconsole-deploy/tasks/main.yml
@@ -188,12 +188,31 @@
   when:
   - ossmconsole_vars.kiali.servicePort == 0
 
+- name: Ask Kiali for information about itself
+  vars:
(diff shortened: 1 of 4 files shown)


jmazzitelli:
Tested and it works.

Quick test is to install Istio and then Kiali. Then scale down the deployment to 0 pods so a kiali pod isn't running.  Then install an OSSMConsole CR and look at the operator logs. You'll see it start to loop retrying to get kiali version info. Scale up the kiali deployment to 1 so a pod starts and then eventually you should see the operator continue processing the OSSMConsole CR and the CR's status field will show the kiali version.

jshaughn (approved):
This makes sense, OSSMC depends on Kiali, it should do its best to wait for Kiali, but not wait forever, just in case someone is determined to install the plugin first.

PR #746, opened 2024-03-05
Add i18n settings to kiali operator

ferhoyos:
Add i18n settings to kiali operator.

```
kiali_feature_flags:
  ui_defaults:
    i18n:
      default: string
      languages: array
```

By default only supported language is English.

To test the PR, build and deploy the kiali-operator in your cluster and modify Kiali CR to add "Chinese" to available languages:
```
  i18n:
    default: zh
    languages: 
    - en
    - zh 
```

After operator reconcile process, Kiali application should show chinese language as default and show language switch component:

image

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -68,7 +68,7 @@ spec:
 
   deployment:
     # default: accessible_namespaces is undefined
-    accessible_namespaces: [ "my-mesh.*" ]
+    accessible_namespaces: ["my-mesh.*"]
     # default: additional_service_yaml is empty
     additional_service_yaml:
       externalName: "kiali.example.com"
@@ -371,7 +371,7 @@ spec:
   istio_labels:
     app_label_name: "app"
     injection_label_name: "istio-injection"
-    injection_label_rev:  "istio.io/rev"
+    injection_label_rev: "istio.io/rev"
     version_label_name: "version"
 
   kiali_feature_flags:
@@ -396,18 +396,21 @@ spec:
         - description: "Find: slow edges (> 1s)"
           expression: "rt > 1000"
         - description: "Find: unhealthy nodes"
-          expression:  "! healthy"
+          expression: "! healthy"
         - description: "Find: unknown nodes"
-          expression:  "name = unknown"
+          expression: "name = unknown"
         hide_options:
         - description: "Hide: healthy nodes"
           expression: "healthy"
         - description: "Hide: unknown nodes"
-          expression:  "name = unknown"
+          expression: "name = unknown"
         traffic:
           grpc: "requests"
           http: "requests"
-          tcp:  "sent"
+          tcp: "sent"
+      i18n:
+        default: "en"
+        languages: ["en"]
       metrics_per_refresh: "1m"
       # default: metrics_inbound is undefined
       metrics_inbound:
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -410,7 +410,7 @@ spec:
           tcp: "sent"
       i18n:
         language: "en"
-        show_selector: true
+        show_selector: false
       list:
         include_health: true
         include_istio_resources: true
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -409,8 +409,13 @@ spec:
           http: "requests"
           tcp: "sent"
       i18n:
-        default: "en"
-        languages: ["en"]
+        language: "en"
+        show_selector: true
+      list:
+        include_health: true
+        include_istio_resources: true
+        include_validations: true
+        show_include_toggles: false
       metrics_per_refresh: "1m"
       # default: metrics_inbound is undefined
       metrics_inbound:
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -140,7 +140,7 @@ spec:
 
                       When empty, this value will default to `openshift` on OpenShift and `token` on other Kubernetes environments.
                     type: string
-                    enum: ["","anonymous","token","openshift","openid","header"]
+                    enum: ["", "anonymous", "token", "openshift", "openid", "header"]
                   openid:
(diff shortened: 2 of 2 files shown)


aljesusg:
I wouldn't add the languages list. 
We could have some users asking that a language is not supported. 
I would only add the default language option, and in case that user set in default a language that is not supported then show him english and a warning message in console. User should be able to see all languages supported by default.

ferhoyos:
My idea with languages list is to show all languages supported by default but a way to differentiate between "beta" and "stable" languages. Currently Chinese is in progress (just few labels translated) so we cannot show it as supported language. When Chinese is finished we will add 'zh' to default languages list. 

This list also provides some flexibility if any organization wants to show only one language or a new language whose translation is in progress (beta). If you think this is too much flexibility and we want a more opinionated configuration in production, we can get rid of languages list for kiali-operator. Anyway I would maintain the languages list in the config map since a language translation can take several sprints and we need some mechanism to show/hide in progress languages from Kiali configuration.

aljesusg:
IMO with another tools...there is not beta/stable languages. If there is not a label in a X langauge then default is showed in that label. It's hard to have the 100% covered, if we add something new we are not going to translate that to all languages.

About languages supported by a tool, this is the first time that I see this perspective. I don't see the goal from a org perspective to disable/enable languages. The option to see an information in a specific langauge is from the user perspective not from the company, another thing would be if we are storing information in a database for example in that language where the data can be affected.

jmazzitelli:
I agree with @aljesusg ... this is requiring the user to tell Kiali what languages it supports - how does the user know that?

The user usually just wants to tell Kiali what default to use. I don't see it as a common use case for a user to enable a sub-set of languages and ignoring others, it's either all-or-nothing.

If you want to user to have the ability to hide the language selector, and the UI only show a default language (this is probably the more common use-case than having a sub-set of languages to choose from) you can add an option here:

```yaml
kiali_feature_flags:
  ui_defaults:
    i18n:
      default: en
      hide_selector: true
```

This would make the UI default to English and not show the language selector (effectively making the UI English-only). If someone wants the UI to be only Chinese, just change to `default: zh`.

(Note: this all assumes the Kiali Server and UI knows all the languages it supports without the user telling it via config.)

If a language isn't fully "supported" (meaning, if not everything is translated) then just show the default (English). That's how I've seen other UIs work. I've never worked on a project (that actually had i18n) that had ever single word available in all languages. There was always some words missing their translations (especially when new words are added to the code and none of the developers have the native language expertise).

ferhoyos:
Ok, no problem, I will remove language list setting if you don't think is needed.

Regarding `hide_selector` feature, I think it should be disabled by default (or even don't add this flag to settings).

jshaughn:
Late to the discussion but I agree that it's not necessary to worry about in-progress languages.  Although I think it may be more clear to just have:

```
kiali_feature_flags:
  ui_defaults:
    i18n:
      languages: array
```

The default is the first language in the list.  If only one language then don't show the selector.

jmazzitelli:
> The default is the first language in the list. If only one language then don't show the selector.

That requires the user to know what languages are even supported. And if they put a bad one in there, then what?

The Kiali Server should know all the languages it has. No need for the user to tell it that.

I think it's best just to be able to set the default language and forget the list of languages. That's confusing and not necessary for the vast majority of use cases.

ferhoyos:
If it is not necessary to worry about in-progress languages, then list of languages don't make sense to me.

What about simplest setting for default language?

```
kiali_feature_flags:
  ui_defaults:
    language: string
```

If the language is not supported, just show English language.

Also ok with @jmazzitelli option with default and hide selector settings.

jmazzitelli:
> If it is not necessary to worry about in-progress languages, then list of languages don't make sense to me.
> 
> What about simplest setting for default language?
> 
> ```
> kiali_feature_flags:
>   ui_defaults:
>     language: string
> ```
> 
> If the language is not supported, just show English language.
> 
> Also ok with @jmazzitelli option with default and hide selector settings.

This is good to me - especially with it being our initial release of this feature. If the community wants more (like the ability to hide/show the selector) then we can do that later.

My main point was - don't make the user provide Kiali with a list of language names.

jmazzitelli:
Oh, but... I would still like the "i18n" section.  Because it (a) makes it clear what the "language" setting is about and (b) it provides the extension point when we do add things later.

```yaml
kiali_feature_flags:
  ui_defaults:
    i18n:
      language: string
```

jshaughn:
I'd say we need the option to have the selector show the supported languages.  It can default to not being shown, but it should be available.

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -409,8 +409,13 @@ spec:
           http: "requests"
           tcp: "sent"
       i18n:
-        default: "en"
-        languages: ["en"]
+        language: "en"
+        show_selector: true
+      list:
```
jmazzitelli, line 414: what is "list" ??

ferhoyos:
From all your feedback I have defined i18n settings like:

```
kiali_feature_flags:
  ui_defaults:
    i18n:
      default: en
      show_selector: true
```

I have renamed `hide_selector` to `show_selector` to be aligned with other Kiali CR similar features like `show_include_toggles`.

I will modify i18n PR of Kiali repo to reflect this change.

aljesusg:
> From all your feedback I have defined i18n settings like:
> 
> ```
> kiali_feature_flags:
>   ui_defaults:
>     i18n:
>       default: en
>       show_selector: true
> ```
> 
> I have renamed `hide_selector` to `show_selector` to be aligned with other Kiali CR similar features like `show_include_toggles`.
> 
> I will modify i18n PR of Kiali repo to reflect this change.

show_selector => show_language_selector ? 
I don't know if user know about **_i18n_**....I would change that to **language**:

ferhoyos:
Sorry I had a copy paste mistake in the comment, it is `language` instead of `default`

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -396,18 +396,26 @@ spec:
...
-          expression:  "name = unknown"
+          expression: "name = unknown"
         traffic:
           grpc: "requests"
           http: "requests"
-          tcp:  "sent"
+          tcp: "sent"
+      i18n:
+        language: "en"
+        show_selector: true
```
jshaughn, line 413: I think this should be false.  We don't want to show the selector by default, most users will just want English.

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
jshaughn: As for the original question, the `list` section can be added, it is a documented feature.

jshaughn (approved):
LGTM!

jmazzitelli (approved):

PR #751, opened 2024-03-19
Add IPv6 support to ossmconsole nginx configuration

ferhoyos:
This PR allows OSSM console plugin to be deployed in OpenShift IPv6 clusters.

Closes https://github.com/kiali/openshift-servicemesh-plugin/issues/281


jmazzitelli (approved):
LGTM.. this should be backported to the 1.73 branch also.

PR #757, opened 2024-03-29
[molecule] dynamically determine the openshift identity provider

jmazzitelli:
This is so we can run this test on different types of clusters (CRC specifically)


jmazzitelli:
This doesn't affect anything other than running a specific molecule test on openshift. This does pass on my internal openshift test machine. I'm merging.

PR #758, opened 2024-04-01
[molecule] fix flake in "os-console-links-test"

jmazzitelli:
I think this should fix it. I do not see this happening on my local machine. This PR also has the positive side effect of cleaning up the test even on failure (before it was leaving test namespaces in the cluster on failure).

fixes: https://github.com/kiali/kiali/issues/7243


jshaughn (approved):
Flake fixes are easy to review because it's hard to make it worse.

PR #760, opened 2024-04-17
Give server OAuth client read permissions

nrfox:
Removes the unused openshift auth settings and grants the kiali server permission to read the kiali oauth client CR.

Relates to https://github.com/kiali/kiali/pull/7280


jmazzitelli:
Reminder of some things that will be needed (helm charts and operator CSV changes for example)... see checklist: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission-that-applies-to-resources-in-all-accessible-namespaces

Since openshift-auth-test is going away, we can remove mention of it here: https://github.com/kiali/kiali/blob/master/hack/run-molecule-tests.sh#L179-L181

nrfox:
The operator already has enough `oauthclient` permissions to grant what is needed to the kiali server: https://github.com/kiali/kiali-operator/blob/ca42ca3063c10294f1b002a1b8331b6439974042/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L483-L493.

jmazzitelli (changes requested):
This should be removed (this is for the Kiali CR schema reference documentation):

* https://github.com/kiali/kiali-operator/blob/master/crd-docs/crd/kiali.io_kialis.yaml#L190-L192
* https://github.com/kiali/kiali-operator/blob/master/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L54

(this stuff is published here  https://kiali.io/docs/configuration/kialis.kiali.io/ -- changing those two locations will remove them from this published Kiali CR reference)

jmazzitelli (approved):

PR #761, opened 2024-04-24
Add a dedicated clusterrole for oauthclient

nrfox:
OAuthClient is a clusterscoped resource so must have a ClusterRole to be granted permissions for it. When `cluster_wide_access = false`, the operator will create Roles/Rolebindings instead of ClusterRole/ClusterRoleBinding and this won't grant any permissions for the OAuthClient resource.


jmazzitelli:
Ok, this is going to some additional effects. Does this mean the operator will always need to create a cluster role/binding? The Kiali operator helm chart has a security lock-down feature for users who don't want to give the operator cluster role creation permission (because this will ultimately mean giving the operator effectively cluster-admin rights -- and some users do not want to give the operator that much power). See:
* https://kiali.io/docs/faq/installation/#operator-fails-due-to-cannot-list-resource-clusterroles-error
* https://github.com/kiali/helm-charts/blob/v1.83.0/kiali-operator/values.yaml#L41-L45

Note that if you install Kiali via OLM, the operator will not have this security lock-down feature. Only the operator helm chart does.

jmazzitelli:
Also, make sure to test when deleting the Kiali CR that these cluster roles are deleted. I believe you might have to add something to the remove role (https://github.com/kiali/kiali-operator/tree/master/roles/default/kiali-remove/tasks)

nrfox:
> Ok, this is going to some additional effects. Does this mean the operator will always need to create a cluster role/binding? 

Yes for openshift at least since `OAuthClient`s are cluster scoped resources and the operator needs to be able to grant permissions to the server to view its respective `OAuthClient`.

> because this will ultimately mean giving the operator effectively cluster-admin rights

Yes but just to get/list `OAuthClient`s. The kiali-server will only be granted permission to get its own `OAuthClient` and nobody else's but the operator will need permissions to `list/get` all.

nrfox:
> because this will ultimately mean giving the operator effectively cluster-admin rights

That's not what we want. I think I can rework this a bit to still enable that use case.

nrfox:
> Ok, this is going to have some additional effects. Does this mean the operator will always need to create a cluster role/binding? The Kiali operator helm chart has a security lock-down feature for users who don't want to give the operator cluster role creation permission (because this will ultimately mean giving the operator effectively cluster-admin rights -- and some users do not want to give the operator that much power). See:

I'm not sure we can get this to work exactly how we'd want it to. My goal with having a separate `ClusterRole` per kiali instance was to ensure that each kiali server can only get/view its own `OAuthClient` and not other `OAuthClients` either for other Kiali instances or for other apps. Using the `resourceNames` field you can restrict which specific `OAuthClient` the `ClusterRole` grants you access to. So in order to restrict a specific Kiali instance's service account to only its `OAuthClient`, we need to create a separate `ClusterRole` for each kiali instance.

To support locking down the operator to not give it permissions to create `ClusterRole`s, we can instead include a separate `ClusterRole` with the operator helm chart for `get/list` on `OAuthClient`s. The operator will update the `ClusterRole` and add each kiali instance it creates to the list of `resourceNames` and the bind this `ClusterRole` to each instance's service account. This means that each kiali instance's service account won't be able to see other app's `OAuthClient`s but they will be able to see other kiali instance's `OAuthClient`s. So it gets us halfway I'd say.

Given that the operator can only create/update `ClusterRole`s that it already has permissions to: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#restrictions-on-role-creation-or-update, is it such a concern to give it permissions to create `ClusterRole`s? In other words, the operator cannot create a `ClusterRole` for itself to escalate privileges beyond what it has already been granted.

jmazzitelli:
> Given that the operator can only create/update `ClusterRole`s that it already has permissions to ... is it such a concern to give it permissions to create `ClusterRole`s?

At some point in the distant past :) I seem to recall there was an issue with escalation being a worry. Maybe there are more locked down features in later versions of kubernetes that did not exist at the time. But regardless, it is still the case that the Kiali Operator has a wide swath of permissions that, granted at the cluster level, would be extensive.

Because this is security related, I'm simply paranoid of a hidden security leak here. That ability to disable cluster role creation was a result of this paranoia. But as I always say, you aren't paranoid if people really are out to get you. So we always need to be careful with the way we expand the permissions we give to the operator.

I don't know the answer here. I'm only saying "tread lightly" and think long and hard before adding cluster roles and cluster role permissions.

nrfox:
> I'm only saying "tread lightly" and think long and hard before adding cluster roles and cluster role permissions.

We're only adding permissions to the kiali server here that the kiali operator already has but as you pointed it this won't work when you've set `clusterRoleCreator=false`. Since the `OAuthClient` resources are only created on openshift this would only affect users who:

1. Use the kiali operator on openshift
2. Don't use OLM
3. Use that `clusterRoleCreator=false` setting

Is that anyone ^^ or can we just say that if you are using the operator on openshift then you either need to use olm or else set `clusterRoleCreator=true`?

nrfox:
> would only affect users who:
> 
>     1. Use the kiali operator on openshift
> 
>     2. Don't use OLM
> 
>     3. Use that `clusterRoleCreator=false` setting

If we want to support this, I think a reasonable compromise is this one:
> To support locking down the operator to not give it permissions to create ClusterRoles, we can instead include a separate ClusterRole with the operator helm chart for get/list on OAuthClients. The operator will update the ClusterRole and add each kiali instance it creates to the list of resourceNames and the bind this ClusterRole to each instance's service account. This means that each kiali instance's service account won't be able to see other app's OAuthClients but they will be able to see other kiali instance's OAuthClients. So it gets us halfway I'd say.

jmazzitelli:
> Is that anyone ^^

I'm guessing more than 0 and less than 2. :)

> can we just say that if you are using the operator on openshift then you either need to use olm or else set `clusterRoleCreator=true`?

I think that's good. In fact, I would add something else - it is also only when using "openshift" auth. If you are using, say, "anonymous" or "header" or "token" auth, I believe this also doesn't come into play (correct?)

nrfox:
> I think that's good. In fact, I would add something else - it is also only when using "openshift" auth. If you are using, say, "anonymous" or "header" or "token" auth, I believe this also doesn't come into play (correct?)

Should be. There's no reason for the operator to create the `OAuthClient` if you are not using `openshift` auth and the server will only try to get the `OAuthClient` if you are using `openshift` auth.

roles/default/kiali-deploy/templates/openshift/clusterrole.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
jmazzitelli, line 1: can you rename this so it is more specific - in [a similar way we have specific names for some role templates](https://github.com/kiali/kiali-operator/tree/master/roles/default/kiali-deploy/templates/openshift) this is for supporting oauth so name this `clusterrole-oauth.yaml`.

roles/default/kiali-deploy/templates/openshift/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
jmazzitelli, line 1: can you rename this so it is more specific - in [a similar way we have specific names for some role binding templates](https://github.com/kiali/kiali-operator/tree/master/roles/default/kiali-deploy/templates/openshift) this is for supporting oauth so name this `clusterrolebinding-oauth.yaml`.

roles/default/kiali-deploy/tasks/openshift/os-main.yml
```diff
@@ -26,6 +26,8 @@
     - "templates/openshift/role-controlplane.yaml"
     - "templates/openshift/rolebinding.yaml"
     - "templates/openshift/rolebinding-controlplane.yaml"
+    - "templates/openshift/clusterrole.yaml"
+    - "templates/openshift/clusterrolebinding.yaml"
```
jmazzitelli, line 30: would like these renamed:
```suggestion
    - "templates/openshift/clusterrole-oauth.yaml"
    - "templates/openshift/clusterrolebinding-oauth.yaml"
```

jmazzitelli (changes requested):
I'm going to test this, but I would like to ask to rename a couple of those files.

roles/default/kiali-deploy/templates/openshift/clusterrole.yaml
nrfox: I kept it intentionally generic in case there were other cluster scoped objects that we wanted to manage in the future similar to how `role.yaml` and `rolebinding.yaml` are named but I can rename if that's not a concern.

nrfox: If we're renaming the files should the clusterrole/binding themselves also be renamed to `<instance>-<namespace>-oauth`?

jmazzitelli: Not a concern about the file names - these are even openshift specific so hopefully we won't need to provide more openshift-specific perms. But if we do, we'll worry about it later.

Might not be a bad idea to rename the resources themselves too - makes it clear what their purpose is for.

roles/default/kiali-deploy/tasks/openshift/os-main.yml
```diff
@@ -26,6 +26,8 @@
     - "templates/openshift/role-controlplane.yaml"
     - "templates/openshift/rolebinding.yaml"
     - "templates/openshift/rolebinding-controlplane.yaml"
+    - "templates/openshift/clusterrole-oauth.yaml"
+    - "templates/openshift/clusterrolebinding-oauth.yaml"
```
jmazzitelli, line 30: Only create the oauth cluster role stuff if we need them (that is, if auth strategy is `openshift`):
```suggestion
    - "{{ 'templates/openshift/clusterrole-oauth.yaml' if kiali_vars.auth.strategy == 'openshift' else '' }}"
    - "{{ 'templates/openshift/clusterrolebinding-oauth.yaml' if kiali_vars.auth.strategy == 'openshift' else '' }}"
```
Here's how to test this:

1. In kiali/kiali repo, `make/Makefile.operator.mk` set operator-cluster-role-creator to false (we never made this configurable via env var - probably should do that one day :)). This change:
```diff
diff --git a/make/Makefile.operator.mk b/make/Makefile.operator.mk
index 317f0251d..450881220 100644
--- a/make/Makefile.operator.mk
+++ b/make/Makefile.operator.mk
@@ -83,7 +83,7 @@ operator-create: .ensure-operator-repo-exists .ensure-operator-helm-chart-exists
     --helm-set                      "image.pullSecrets={${OPERATOR_IMAGE_PULL_SECRET_NAME}}" \
-    --operator-cluster-role-creator "true" \
+    --operator-cluster-role-creator "false" \
     --operator-image-name           "${CLUSTER_OPERATOR_INTERNAL_NAME}" \
```
2. Push and deploy the operator to the cluster - this operator will not have permission to create cluster roles: `make cluster-push operator-create`
3. Now try to install Kiali server with auth strategy of openshift - this should fail with permission errors: `make -e AUTH_STRATEGY="openshift" -e ACCESSIBLE_NAMESPACES="istio-system" kiali-create`
4. You can confirm step 3 fails by watching the progress in the Kiali CR - you will eventually see an error logged in the status field: `oc get kiali kiali -n kiali-operator -oyaml -w`
5. Now update that Kiali CR so that the auth strategy is anonymous: `make -e AUTH_STRATEGY="anonymous" -e ACCESSIBLE_NAMESPACES="istio-system" kiali-create`
6. Eventually, Kiali should be installed. Wait for it in istio-system -- `oc get pods -n istio-system -w`
7. Navigate to the Kiali route and see that you get to the UI without logging in, confirming anonymous mode was set.
8. Notice there are no cluster roles/bindings created for the server; you should only have the clusterroles for the operator itself:
```
 $ oc get clusterroles,clusterrolebindings -o name | grep kiali
clusterrole.rbac.authorization.k8s.io/kiali-operator
clusterrolebinding.rbac.authorization.k8s.io/kiali-operator
```

jmazzitelli (changes requested):

jmazzitelli (approved):

hhovsepy:
Issue fixed on OCP cluster.

PR #762, opened 2024-04-29
be able to enable/disable namespace watching

jmazzitelli:
This disables namespace watching by default. But it provides a way for the user to enable it if that feature is desired.

fixes: https://github.com/kiali/kiali/issues/7322

To test (you could test on an non-OpenShift cluster but the steps below are for OpenShift - I used CRC):

1. Be logged into an OpenShift cluster and make sure Istio is installed
2. Log into the image registry of the cluster (see the `make cluster-status` output for the command to use to login)
3. Push the dev images to the cluster via `make cluster-push`
4. Install the operator via OLM using `make olm-operator-create`
5. At this point, the default behavior should NOT watch namespaces. Look at the operator logs and you should not see the namespace processing playbook being run for every namespace in the cluster. The operator logs should be relatively quiet. The operator is located in the `openshift-operators` namespace (so you do something like `oc logs -n openshift-operators <operator pod name>`).
6. Now turn on the namespace watching. The FAQ explains the details here: https://kiali.io/docs/faq/installation/#operator-configuration but below is what you want to do. You basically want to set the operator Deployment's env var `WATCHES_FILE` to `watches-os-ns.yaml` (or `watches-k8s-ns.yaml` if testing on non-k8s but for this test we are using OpenShift). You can do this via this set of commands:
```bash
ENV_NAME="WATCHES_FILE"

ENV_VALUE="watches-os-ns.yaml"

OPERATOR_NAMESPACE="openshift-operators"
oc -n ${OPERATOR_NAMESPACE} patch $(oc -n ${OPERATOR_NAMESPACE} get csv -o name | grep kiali) --type=json -p "[{'op':'replace','path':"/spec/install/spec/deployments/0/spec/template/spec/containers/0/env/$(oc -n ${OPERATOR_NAMESPACE} get $(oc -n ${OPERATOR_NAMESPACE} get csv -o name | grep kiali) -o jsonpath='{.spec.install.spec.deployments[0].spec.template.spec.containers[0].env[*].name}' | tr ' ' '\n' | cat --number | grep ${ENV_NAME} | cut -f 1 | xargs echo -n | cat - <(echo "-1") | bc)/value",'value':"\"${ENV_VALUE}\""}]"
```
7. Do the same as step 5 - look at the operator logs - but this time you should see namespace processing happening, once for each namespace in the cluster.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -12,6 +12,8 @@ COPY roles/ ${HOME}/roles/
 COPY playbooks/ ${HOME}/playbooks/
 COPY watches-k8s.yaml ${HOME}/watches-k8s.yaml
 COPY watches-os.yaml ${HOME}/watches-os.yaml
+COPY watches-k8s-ns.yaml ${HOME}/watches-k8s-ns.yaml
+COPY watches-os-ns.yaml ${HOME}/watches-os-ns.yaml
 
 COPY requirements.yml ${HOME}/requirements.yml
 RUN ansible-galaxy collection install -r ${HOME}/requirements.yml \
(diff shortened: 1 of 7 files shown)


hhovsepy (approved):
Verified on a remote OCP cluster, tested default disable behavior: no logs, enable it: logs are back.

PR #768, opened 2024-05-07
be able to set DNS policy and config on the Kiali pod

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7150


jmazzitelli:
`config-values-test` molecule test is testing this operator feature.

```
./hack/run-molecule-tests.sh --client-exe "$(which kubectl)"  --minikube-exe $(which minikube) --minikube-profile ci --cluster-type minikube --use-dev-images true -at config-values-test
```

jshaughn (approved):
molecule testing passed.

PR #772, opened 2024-06-05
Add config for the server timeout

josunect:
Added new option to make the server timeout as a config, in case there is the need to increase it. 
It is called just timeout for simplicity, but it is just affecting the server write time out (Which I think it is usually the problematic). 

Ref PR: https://github.com/kiali/kiali/pull/7389

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1402,6 +1402,9 @@ spec:
                       enabled:
                         description: "When 'true', the profiler will be enabled and accessible at /debug/pprof/ on the Kiali endpoint."
                         type: boolean
+                  timeout:
+                    description: "The write timeout in seconds for the server. Default is 30."
+                    type: integer
                   web_fqdn:
                     description: "Defines the public domain where Kiali is being served. This is the 'domain' part of the URL (usually it's a fully-qualified domain name). For example, `kiali.example.org`. When empty, Kiali will try to guess this value from HTTP headers. On non-OpenShift clusters, you must populate this value if you want to enable cross-linking between Kiali instances in a multi-cluster setup."
                     type: string
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1402,9 +1402,6 @@ spec:
                       enabled:
                         description: "When 'true', the profiler will be enabled and accessible at /debug/pprof/ on the Kiali endpoint."
                         type: boolean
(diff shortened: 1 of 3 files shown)


crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1402,6 +1402,9 @@ spec:
                       enabled:
                         description: "When 'true', the profiler will be enabled and accessible at /debug/pprof/ on the Kiali endpoint."
                         type: boolean
+                  timeout:
+                    description: "The write timeout in seconds for the server. Default is 30."
```
jmazzitelli, line 1406: I'm not quite sure what this setting is for or does based on the description here. Can we add more description here to be clear what it is. I'm not sure what "write timeout" means... is this the browser client timeout?

jmazzitelli (changes requested):

crd-docs/crd/kiali.io_kialis.yaml
jmazzitelli: OK, I see it: 
```go
	httpServer := &http.Server{
		Addr:         fmt.Sprintf("%v:%v", conf.Server.Address, conf.Server.Port),
		TLSConfig:    tlsConfig,
		ReadTimeout:  30 * time.Second,
		WriteTimeout: writeTimeout,
```
We should name this "write_timeout" to make it specific... it will help people better understand what this "timeout" is.

jmazzitelli: Here's the Go description: https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/net/http/server.go;l=2856-2861

Note there are other timeouts here (ReadTimeout, ReadHeaderTimeout, etc..), so calling our setting "timeout" will be ambiguous to users. Hence why we should be specific here and call it "write_timeout".

jmazzitelli: The description should be something like:

```
description: "The maximum duration, in seconds, before timing out writes of the HTTP response back to the client. Default is 30." 
```

josunect: It is the server timeout itself. If the server takes longer than this value to process a request, the response to the client will be aborted. See this for further info: https://github.com/kiali/kiali/blob/master/server/server.go#L86

jmazzitelli: > It is the server timeout itself.

There is more than one "server timeout" though - see my comments above.

jmazzitelli (approved):

PR #778, opened 2024-06-17
Remove 1.48

aljesusg:
Working in ossm 2.6

We should support:
2.6 => 1.73
2.5 => 1.73
2.4 => 1.65

So...we should remove 1.48 and 1.57 right ? 

cc @jmazzitelli @jshaughn


jmazzitelli:
Do not remove 1.57. We want to "unofficially" support the N-3rd (just like we did for 1.48)

So just remove 1.48.

aljesusg:
Ok Thanks !
 I assumed they were ossm releases not kiali releases let me make the changes

jmazzitelli (approved):

PR #779, opened 2024-06-20
Add role 1.86

aljesusg:
Add role 1.86 ~~and remove 1.57~~ in master for ossm 3.0


jmazzitelli (changes requested):
add 1.86 to the OSSMC file: https://github.com/kiali/kiali-operator/blob/master/playbooks/ossmconsole-default-supported-images.yml

jmazzitelli (approved):

PR #784, opened 2024-06-27
Remove config map and deployment name defaults

nrfox:


jmazzitelli (commented):
Are these no longer actually used anywhere? If so, remove them from the CRD docs:

* example CR: https://github.com/kiali/kiali-operator/blob/v1.86.3/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L298-L313
* CRD docs: https://github.com/kiali/kiali-operator/blob/v1.86.3/crd-docs/crd/kiali.io_kialis.yaml#L795-L849

nrfox:
> Are these no longer actually used anywhere? If so, remove them from the CRD docs:

You still can set them manually but this change is removing the operator defaulting them so that the kiali server will autodetect them. When you have these set the kiali server will respect their values and try to use what you have set instead of autodetecting.

jmazzitelli (approved):
As long as the operator never uses these, and as long as the server detects a missing value as an empty string, then this is OK.

if we think there is a potential for some kind of nil pointer or if the operator ever needs to use these, then I would recommend leaving the values set in defaults/main.yml, just change the defaults to empty string `""`.

I'm OK either way so long as no potential nil value issues are posssible, and approving.

nrfox:
> if we think there is a potential for some kind of nil pointer or if the operator ever needs to use these, then I would recommend leaving the values set in defaults/main.yml, just change the defaults to empty string `""`.

These are all plain strings in kiali not pointers so there's no risk of nil pointers there and I can't see why the operator would need to use these again in the future.

PR #788, opened 2024-06-28
Added Istio Ingress and Egress Gateway workload label selectors into configs.

hhovsepy:
Added configurations for `egress_gateway_label` and `ingress_gateway_label`. Before they were hardcoded in the code.

Issue https://github.com/kiali/kiali/issues/7232
Core PR https://github.com/kiali/kiali/pull/7492

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -389,6 +389,8 @@ spec:
 
   istio_labels:
     app_label_name: "app"
+    egress_gateway_label: "istio=egressgateway"
+    ingress_gateway_label: "istio=ingressgateway"
     injection_label_name: "istio-injection"
     injection_label_rev: "istio.io/rev"
     version_label_name: "version"
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1057,6 +1057,12 @@ spec:
                   app_label_name:
                     description: "The name of the label used to define what application a workload belongs to. This is typically something like `app` or `app.kubernetes.io/name`."
                     type: string
+                  egress_gateway_label:
+                    description: "The selector label for Egress Gateway workload. This is typically `istio=egressgateway`."
+                    type: string
+                  ingress_gateway_label:
+                    description: "The selector label for Ingress Gateway workload. This is typically `istio=ingressgateway`."
+                    type: string
                   injection_label_name:
                     description: "The name of the label used to instruct Istio to automatically inject sidecar proxies when applications are deployed."
                     type: string
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -250,6 +250,8 @@ kiali_defaults:
 
   istio_labels:
     app_label_name: "app"
+    egress_gateway_label: "istio=egressgateway"
+    ingress_gateway_label: "istio=ingressgateway"
     injection_label_name: "istio-injection"
     injection_label_rev: "istio.io/rev"
     version_label_name: "version"


jmazzitelli (commented):
We already have something like this in the config in another place:

https://github.com/kiali/kiali-operator/blob/v1.86.0/roles/default/kiali-deploy/defaults/main.yml#L173-L177

@nrfox is in the process of trying to remove some of these istio settings. I think we should try not to add more, especially when they seem to be already defined in another place. Because now if the ingress or egress labels are different than the default, the user will now have to know to change TWO different places in the config. We should try to at least consolidate into a single config setting (if not be able to auto-discover these).

nrfox:
> @nrfox is in the process of trying to remove some of these istio settings. I think we should try not to add more, especially when they seem to be already defined in another place. Because now if the ingress or egress labels are different than the default, the user will now have to know to change TWO different places in the config. We should try to at least consolidate into a single config setting (if not be able to auto-discover these).

I agree having two places where you need to set the label for the same thing is not good. Putting the setting under `istio_labels` probably makes more sense to me though than `external_services.istio.component_status.components[*].app_label`. I think the `external_services.istio.component_status.components` settings are just for monitoring the status of the different istio components anyways? Whereas the config that you are adding would apply to more things like creating gateways through the wizard?

hhovsepy:
> > @nrfox is in the process of trying to remove some of these istio settings. I think we should try not to add more, especially when they seem to be already defined in another place. Because now if the ingress or egress labels are different than the default, the user will now have to know to change TWO different places in the config. We should try to at least consolidate into a single config setting (if not be able to auto-discover these).
> 
> I agree having two places where you need to set the label for the same thing is not good. Putting the setting under `istio_labels` probably makes more sense to me though than `external_services.istio.component_status.components[*].app_label`. I think the `external_services.istio.component_status.components` settings are just for monitoring the status of the different istio components anyways? Whereas the config that you are adding would apply to more things like creating gateways through the wizard?

`external_services.istio.component_status.components[*].app_label` is for Istio core component's health check, to check the health of workload which has "app: istio-ingressgateway" label, this is not mandatory field. 
Where the added config is to check whether the workload has label "istio: ingressgateway" and thus is Gateway (used in wizards and in details pages). 
The 'istio-ingressgateway' workload itself has both of those labels in a same time.
So we cannot unify those 2 types of labels into a single config.

jmazzitelli:
> The 'istio-ingressgateway' workload itself has both of those labels in a same time.

Is there ANY way we can reduce the labels we care about to just 1, and have it configured in 1 place (and better yet, have it auto-discovered without needing a config)? Since the workload has both, we can just look at one or the other.

@nrfox is in the process of refactoring the server in order that users are not burdened with having to configure so many Istio settings in the Kiali CR (look at [this list](https://github.com/kiali/kiali-operator/blob/v1.86.0/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L281-L313) - we want to reduce that). He removed a couple already, but now this PR is adding another one, which is duplicating the identification of the gateways (albeit using different labels - which makes it even more confusing and burdensome to users who have to configure Kiali to work with their Istio setup). Two steps forward, one step back :)

hhovsepy:
> @hhovsepy are these labels also present when you create gateway API gateways?

No, those are for Istio gateways only.

hhovsepy:
> Is there ANY way we can reduce the labels we care about to just 1, and have it configured in 1 place (and better yet, have it auto-discovered without needing a config)? Since the workload has both, we can just look at one or the other.
> 

In Istio those labels are also 2 separate ones, one is for app label, and another is for service name, and they are taken from command line arguments separately.
IMO if we make them single label in Kiali CR and reuse in both places this can be wrong, also implementation of status component which accepts lists of labels and checks in loops  should be totally redesigned, which is out of scope.
So IMO solution is either keep the label `istio=ingressgateway` as hardcoded (make a constant) or make a separate config line then redesign the component status check in another task. Also for instance `app=istiod` label is hardcoded in code (as a constant), but it also set is in `external_services.istio.component_status.components[*].app_label`.

jmazzitelli:
@hhovsepy I suggest you write up a second github issue so we don't forget about this. We need to come up with a way to combine the two.

I was thinking rather than require "app" label, have the component check be able to define a custom label NAME as well as label VALUE.. so add app_label_name like this:

```yaml
component_status:
  components:
  - app_label: "ingressgateway"
    app_label_name: "istio"     <---- add this
    is_core: true
    is_proxy: true
```

That is backward compatible, but still requires someone to maintain that same label in two different places in the Kiali CR. So I'm still not crazy about that. We do want to collapse the config as much as we can so a person only has to configure the ingress/egress labels in one place, not multiple (my solution above doesn't do that).

hhovsepy:
I agree @jmazzitelli , have created a separate issue for that https://github.com/kiali/kiali/issues/7524 as it requires a rework of code and configuration to have single place of setting labels.

jmazzitelli (approved):
I approve. We will fix the duplication of the label config in another PR.

PR #790, opened 2024-07-03
Disable tracing by default

josunect:
Set default for `external_services.tracing.enabled: false`

Ref. [#7332](https://github.com/kiali/kiali/issues/7332)

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -349,7 +349,7 @@ spec:
       # default: custom_headers is empty
       custom_headers:
         customHeader1: "customHeader1Value"
-      enabled: true
+      enabled: false
       grpc_port: 9095
       health_check_url: ""
       in_cluster_url: ""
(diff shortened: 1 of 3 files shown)


jmazzitelli (approved):

jmazzitelli:
There is a jaeger-test molecule test... make sure that passes before merging :)

jmazzitelli:
Molecule test needs a change... go to this file: https://github.com/kiali/kiali-operator/blob/master/molecule/jaeger-test/kiali-cr.yaml and add `spec.external_services.tracing.enabled` with value `true` to the bottom so the test turns on tracing since it is now off by default. The diff looks like this:

```diff
index 4a5b10c..879b97d 100644
--- a/molecule/jaeger-test/kiali-cr.yaml
+++ b/molecule/jaeger-test/kiali-cr.yaml
@@ -18,3 +18,6 @@ spec:
     image_version: "{{ kiali.image_version }}"
     accessible_namespaces: {{ kiali.accessible_namespaces }}
     service_type: {{ 'LoadBalancer' if is_kind else 'NodePort' }}
+  external_services:
+    tracing:
+      enabled: true
```

jmazzitelli (changes requested):
see my change request in my last comment

jmazzitelli (changes requested):

jmazzitelli (approved):

PR #791, opened 2024-07-04
Upgrade github actions to Node 20

ferhoyos:
Currently, github actions use v3 of actions/checkout, which internally runs on Node 16 (https://github.com/actions/checkout/blob/v3/action.yml#L90). 

This PR upgrades github actions to v4, which internally runs Node on 20 (https://github.com/actions/checkout/blob/v4/action.yml#L102)


jmazzitelli (approved):
i don't see any issues with this. minor updates.

PR #797, opened 2024-07-23
Discovery Selectors

jmazzitelli:
Implement the new discovery selector support in the operator.

part of: https://github.com/kiali/kiali/issues/7546
part of KEP: https://github.com/kiali/kiali/blob/master/design/KEPS/namespace-discovery/proposal.md

server PR: https://github.com/kiali/kiali/pull/7592
operator PR: https://github.com/kiali/kiali-operator/pull/797
helm chart PR: kiali/helm-charts#274
docs PR: https://github.com/kiali/kiali.io/pull/807


jmazzitelli:
Some things I know I still need to get the operator to do:
* attempt to auto-detect istio configmap
* set the discovery_selectors in the configmap to select only those namespaces that the operator gave kiali access to

jmazzitelli:
As of right now, I can confirm the following molecule tests pass (minus the server-side testing - mainly calls into the server's namespaces API):

```
accessible-namespaces-test
affinity-tolerations-resources-test
cluster-wide-access-test
config-values-test
default
default-namespace-test
instance-name-test
null-cr-values-test
only-view-only-mode-test
os-console-links-test
read-certs-secrets-test
roles-test
```

Next up the server has to be refactored to move away from accessible_namespaces and go use discovery selectors. Once the server is ready, the following molecule tests then need to be tested (along with the ones above, since they were not testing server-side functionality at the time I got them to pass).
```
grafana-test
header-auth-test
jaeger-test
metrics-test
openid-test
rolling-restart-test
token-test
```

jmazzitelli:
Two more things added to this PR.

First, when discovery selectors are defined AND cluster wide access is FALSE, the operator will set `discovery_selectors.default` in the ConfigMap to be a list of simple match_labels, one for each namespace. They will match on the namespace name label `kubernetes.io/metadata.name`. So, in other words, a user can provide a complex set of selectors in the Kiali CR - the operator will find all namespaces that match those selectors at the time the Kiali CR is reconciled. The operator will then replace the `discovery_selectors.default` field in the ConfigMap - it won't have those complex set of selectors; instead, it will have a list of simple match_labels selectors - one `kubernetes.io/metadata.name` selector per namespace. This way the server will be ensured to select those namespaces (and only those namespaces) in which it was given permission to see (the operator's Roles will ensure permissions are set up correctly for those namespaces, but only those namespaces).

Note that if cluster_wide_access is True, the operator simply leaves the ConfigMap's discovery selectors to those complex set of selectors the user put in the Kiali CR. This is because Kiali is given permission to see any and all namespaces in the cluster, so if new namespaces are created and match those selectors, the Kiali Server will see them (assuming it queries for namespaces again using the selectors). There is no need to fix the Kiali Server on a set list of namespaces. This behavior is the same as the old `accessible_namespaces` of `["**"]`.

Secondly, the operator will attempt to find the Istio ConfigMap (regardless of its name) in the istio control plane namespace. It will look at all ConfigMaps, and the first it finds that has "data.mesh" field, that will be used as the Istio ConfigMap. It will then look inside to see if there are any discoverySelectors for the operator to use. Note that the operator only ever looks for the Istio ConfigMap if there is no `discovery_selectors.default` setting in the Kiali CR.

jmazzitelli:
reminder to me to document the labels that the operator creates. things like `kiali.io/kiali.home` and other labels.

jmazzitelli:
namespace label (created when the operator makes the namespace accessible to the kiali server):
* `kiali.io/<instance-name>.home` == `<kiali deployment.namespace>`

remote cluster secret annotation (created by script that creates remote cluster secrets):
* `kiali.io/cluster` == `<cluster name>`

kiali deployment annotation (created/updated by the operator when it restarts the kiali server pod)
* `operator.kiali.io/last-updated` == `<date/time of last update>`

ossmconsole deployment annotation (created/updated by the operator when it restarts the ossmconsole server pod)
* `ossmconsole.kiali.io/last-updated` == `<date/time of last update>`

nrfox (approved):

jmazzitelli:
I made a change - now that the server supports multiple "values" for a matchExpressions, I change the operator so it creates those instead of multiple matchLabels (makes reading the configmap easier and probably more efficient on the server side).

Molecule tests all pass after making this change:
```
=====================
=== TEST RESULTS: ===
=====================

              accessible-namespaces-test... success [6m 36s]
     affinity-tolerations-resources-test... success [3m 20s]
                cluster-wide-access-test... success [7m 30s]
                      config-values-test... success [2m 23s]
                  default-namespace-test... success [1m 46s]
                            grafana-test... success [3m 8s]
                        header-auth-test... success [1m 37s]
                      instance-name-test... success [2m 10s]
                             jaeger-test... success [2m 33s]
                            metrics-test... success [3m 31s]
                     null-cr-values-test... success [1m 33s]
                only-view-only-mode-test... success [2m 20s]
                             openid-test... success [1m 36s]
                 read-certs-secrets-test... success [1m 33s]
                              roles-test... success [4m 46s]
                    rolling-restart-test... success [2m 15s]
                              token-test... success [1m 53s]
```

jshaughn (approved):
Change is fine by me, more efficient and easier to look at.

nrfox (approved):

PR #800, opened 2024-08-03
fix processing around replicas when HPA is involved

jmazzitelli:
Testing on CRC/openshift, there was a lag when HPA adjusts replicas on the Deployment. Add a check in the molecule test so it waits for the replicas to reach steady state before continuing.

After I fixed that, I noticed another race condition in the code related to HPA that I fixed here.

### DIFF
--- a/molecule/affinity-tolerations-resources-test/converge.yml
+++ b/molecule/affinity-tolerations-resources-test/converge.yml
@@ -159,6 +159,21 @@
       - hpa_resource_raw.resources | length == 1
       fail_msg: "HPA was not created successfully: {{ hpa_resource_raw }}"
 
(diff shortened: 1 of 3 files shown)


jmazzitelli:
to see the test fixed...

1. *OpenShift*: start a CRC cluster, install Istio, then run: `make -e MOLECULE_SCENARIO=affinity-tolerations-resources-test molecule-test`
2. *Minikube*: start minikube, install Istio, then run: `make -e CLUSTER_TYPE=minikube MOLECULE_SCENARIO=affinity-tolerations-resources-test molecule-test`

aljesusg (approved):
No tested but code seems ok . LGFM

PR #801, opened 2024-08-13
check Kiali version when installing OSSMC

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7619

### DIFF
--- a/manifests/kiali-community/1.89.0/manifests/kiali.v1.89.0.clusterserviceversion.yaml
+++ b/manifests/kiali-community/1.89.0/manifests/kiali.v1.89.0.clusterserviceversion.yaml
@@ -286,6 +286,8 @@ spec:
                   value: "true"
                 - name: ACCESSIBLE_NAMESPACES_LABEL
                   value: ""
+                - name: OSSMC_SKIP_VERSION_CHECK
+                  value: "false"
                 - name: PROFILE_TASKS_TASK_OUTPUT_LIMIT
                   value: "100"
                 - name: ANSIBLE_DEBUG_LOGS
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -296,6 +296,8 @@ spec:
                   value: "true"
                 - name: ACCESSIBLE_NAMESPACES_LABEL
                   value: "maistra.io/member-of"
+                - name: OSSMC_SKIP_VERSION_CHECK
+                  value: "false"
                 - name: PROFILE_TASKS_TASK_OUTPUT_LIMIT
                   value: "100"
                 - name: ANSIBLE_DEBUG_LOGS
--- a/manifests/kiali-upstream/1.89.0/manifests/kiali.v1.89.0.clusterserviceversion.yaml
+++ b/manifests/kiali-upstream/1.89.0/manifests/kiali.v1.89.0.clusterserviceversion.yaml
@@ -235,6 +235,8 @@ spec:
                   value: "true"
                 - name: ACCESSIBLE_NAMESPACES_LABEL
                   value: ""
+                - name: OSSMC_SKIP_VERSION_CHECK
+                  value: "false"
                 - name: PROFILE_TASKS_TASK_OUTPUT_LIMIT
                   value: "100"
                 - name: ANSIBLE_DEBUG_LOGS
--- a/roles/default/ossmconsole-deploy/tasks/main.yml
+++ b/roles/default/ossmconsole-deploy/tasks/main.yml
@@ -240,6 +240,27 @@
         serviceNamespace: "{{ ossmconsole_vars.kiali.serviceNamespace }}"
         servicePort: "{{ ossmconsole_vars.kiali.servicePort }}"
 
+# The OSSMC plugin must talk to a Kiali of the same version.
+# We asked the Kiali Server what version it is, so we already have that.
+# To know the OSSMC version, we assume even if the user overrides the imageName/imageVersion, that that image must
+# support the same major.minor version of the playbook (i.e. spec.version). We look up the version of the playbook
+# in supported_ossmconsole_images and obtain the imageVersion that that playbook supports and assume that is the version
+# of OSSMC that is being installed - so that is the version we compare with the Kiali version. Note that if imageVersion
+# is the string "operator_version", the supported OSSMC version is the same version as the operator itself.
+# Note that we ignore any "v" prefix that may be in front of the version numbers; always just compare major.minor (v or no-v).
+# If, for some reason, the user runs into an edge case where OSSMC needs to be installed even if the versions don't match,
+# set the env var OSSMC_SKIP_VERSION_CHECK=true in the operator.
+- name: Ensure Kiali major.minor version is supported by the version of OSSMC being installed
+  vars:
+    kiali_version_to_match: "{{ kiali_version | default('') | regex_replace('^[v]?([0-9]+.[0-9]+).*', '\\1') }}"
+    ossmc_version_expected: "{{ supported_ossmconsole_images[ossmconsole_vars.version].imageVersion | default ('') }}"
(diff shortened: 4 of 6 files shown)


jmazzitelli:
molecule tests still pass:
1. `make cluster-push` from the ossmc local repo, and `make cluster-push` from the kiali/kiali repo
2. `hack/run-molecule-tests.sh --client-exe "$(which oc)" -dorp podman --cluster-type openshift -udi true -at "ossmconsole-default ossmconsole-config-values-test"`

```
=====================
=== TEST RESULTS: ===
=====================

                     ossmconsole-default... success [2m 3s]
          ossmconsole-config-values-test... success [2m 9s]
```

jmazzitelli:
How to test (of course, before you start, make sure this PR is checked out into your local operator repo and your Kiali dev environment is setup):

1. Get an OpenShift cluster up with Istio installed (`hack/crc-openshift.sh start` then `hack/istio/install-istio-via-istioctl.sh -c oc`)
2. Build, push, and deploy the operator: `make cluster-push-operator operator-create` (you MUST do this before any followup steps - do not do the next step (install Kiali) first; do that after you install the operator).
3. Install Kiali v1.73.0, but do so using the server helm chart (this is just to see this works with some random Kiali installation, not one necessarily installed by the operator):
```sh
# add or update your kiali helm repo
helm repo add kiali https://kiali.org/helm-charts && helm repo update
# install server v1.73.0
helm install --set auth.strategy=anonymous --namespace istio-system --version 1.73.0 kiali-server kiali/kiali-server
```
4. Now attempt to install OSSMC using `spec.version` of `default`:
```sh
make ossmconsole-create
```
5. Because our current master is not 1.73 but something like 1.89 (i.e. 1.89-SNAPSHOT), this should fail. Look at the status field of the OSSMC CR to confirm the failure:
```sh
oc get ossmconsole ossmconsole -n ossmconsole -o jsonpath='{.status}' | jq
```
should have an error message in there that looks like this:
```
"message": "The version of the Kiali Server [1.73] does not match the
OSSMC version [1.89]. OSSMC will not be installed. To skip this check
and allow for the install to continue, set the operator environment
variable OSSMC_SKIP_VERSION_CHECK to 'true'.",
```
6. Now edit the OSSMC CR so that its `spec.version` field is `v1.73` - thus matching the version of Kiali:
```sh
oc patch ossmconsole ossmconsole -n ossmconsole --type='json' -p='[{"op": "replace", "path": "/spec/version", "value": "v1.73"}]'
```
7. Now look at the OSSMC CR status and things should be successful. You won't have an error message in there, it will have finished its progress, and you should have a ossmconsole plugin running:
```sh
# this shows all good
oc get ossmconsole ossmconsole -n ossmconsole -o jsonpath='{.status}' | jq

# this shows running pod
oc get pods -n ossmconsole 
```
8. Uninstall OSSMC by removing the OSSMC CR: `oc delete ossmconsole ossmconsole -n ossmconsole`
9. Install OSSMC v1.86 - our makefile can set the spec.version field with the correct env var set:
```sh
make -e OSSMCONSOLE_CR_SPEC_VERSION=v1.86 ossmconsole-create
```
10. Again, this should fail because the 1.86 does not match the 1.73 version of the server. Confirm by looking at the status field of the OSSMC CR and see the failsure message:
```sh
oc get ossmconsole ossmconsole -n ossmconsole -o jsonpath='{.status}' | jq
```
11. Now tell the operator to skip the version check and install OSSMC anyway. This is a pull-in-case-of-emergency switch where a user can tell the operator to just install OSSMC even if for some reason the version check fails but knows OSSMC is compatible. This is to cover unforeseen edge cases when the operator is deployed out in the field:
```sh
oc set env deployment/kiali-operator -n kiali-operator OSSMC_SKIP_VERSION_CHECK=true
```
12. Wait for the operator to restart and reconcile the OSSMC CR - eventually, you should see OSSMC install. Look at the OSSMC CR status and confirm it is successful, then notice you have the OSSMC pod running:
```sh
# this shows all good
oc get ossmconsole ossmconsole -n ossmconsole -o jsonpath='{.status}' | jq

# this shows running pod
oc get pods -n ossmconsole 
```

To clean up this test, `helm uninstall kiali-server -n istio-system && make operator-delete` or just tear down the CRC cluster if you don't need it.

jmazzitelli:
molecule tests still pass:
1. `make cluster-push` from the ossmc local repo, and `make cluster-push` from the kiali/kiali repo
2. `hack/run-molecule-tests.sh --client-exe "$(which oc)" -dorp podman --cluster-type openshift -udi true -at "ossmconsole-default ossmconsole-config-values-test"`

```
=====================
=== TEST RESULTS: ===
=====================

                     ossmconsole-default... success [2m 3s]
          ossmconsole-config-values-test... success [2m 9s]
```

ferhoyos:
@jmazzitelli what is the best value to specify the version of OSSMC? I know you can do it with `spec.version` and `spec.deployment.imageVersion`. According to https://kiali.io/docs/configuration/ossmconsoles.kiali.io/#.spec.version, the only supported value for `spec.version` in the OSSMConsole CR is `default`, and it suggests not using this field to set the version of OSSMC. However, the `Version` field still appears in the OSSMConsole CR form in the OpenShift console (as mentioned here https://kiali.io/docs/installation/installation-guide/creating-updating-ossmconsole-cr/#creating-the-ossmconsole-cr-to-install-the-ossmc-plugin). I find this a bit confusing.

ferhoyos (approved):
LGTM. Tested locally

jmazzitelli:
> @jmazzitelli what is the best value to specify the version of OSSMC? I know you can do it with `spec.version` and `spec.deployment.imageVersion`. According to https://kiali.io/docs/configuration/ossmconsoles.kiali.io/#.spec.version, the only supported value for `spec.version` in the OSSMConsole CR is `default`, and it suggests not using this field to set the version of OSSMC. However, the `Version` field still appears in the OSSMConsole CR form in the OpenShift console (as mentioned here https://kiali.io/docs/installation/installation-guide/creating-updating-ossmconsole-cr/#creating-the-ossmconsole-cr-to-install-the-ossmc-plugin). I find this a bit confusing.

First, let me make it clear what those two settings are.

* `spec.version` is not really specifying the version of the image to install. It really is specifying the name of the Ansible role to install (one of [these](https://github.com/kiali/kiali-operator/tree/master/roles)). But if we named it `spec.role` or `spec.ansible_role`, people wouldn't understand it and it would be even more confusing. :) When you read [the spec.version docs](https://kiali.io/docs/configuration/ossmconsoles.kiali.io/#.spec.version), we do mention this (though we say it refers to an Ansible `playbook`, but it is really the role... again, if we said "role" in the docs it would confuse people making them think it has something to do with permissions, or Kubernetes Roles, so we just say "playbook" since it is close enough to get the point across).
* `spec.deployment.imageVersion` is literally the version of the image as pulled down from the image registry. If you do NOT specify this, then the playbook that is run (and that is determined by `spec.version`) will give you a version by default. So `spec.version` does indirectly specify what version you install since it defines the default. You can override that default, though - overriding the default is what this `spec.deployment.imageVersion` does.

Now, the confusion comes in when you consider upstream versus downstream (OpenShift production) users. The reason we say this in the docs:  `the only supported value for spec.version in the OSSMConsole CR is default` is because *upstream* doesn't support running any playbooks other than the latest one (i.e. `default`) -- this is true even for the Kiali CR as you [see here](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.version).

For OpenShift users, they are 9 times out of 10 (probably more) downstream users so  they can use different settings for `spec.version`.

And, technically, upstream users can, too. We just want to make it clear the upstream operator does not officially support any other playbook other than the "default" one so they should not assume an earlier version of the playbook will work. But, 99% of the time, it will work. The docs are just trying to set people's expectations for production use.

With all of that said, I think we can re-word the upstream documentation to help clear up some of the confusion you had (maybe just use some of what I said above in the updated docs). I will create a separate issue and submit another PR for docs and re-word some of this. But that would be separate from this PR. I'll merge this PR and work on the docs separately.

PR #810, opened 2024-08-19
change supported versions from v1.86 to v1.89

jmazzitelli:
followed the checklist: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-adding-support-for-a-new-ansible-role-version


aljesusg (approved):
LGFM

PR #811, opened 2024-08-19
[v1.89] change supported versions from v1.86 to v1.89

jmazzitelli:
cherry pick of https://github.com/kiali/kiali-operator/pull/810


aljesusg (approved):
LGFM

PR #813, opened 2024-08-30
system namespaces are now filtered out when no discovery selectors are set

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/7687

no changes to the operator need to be made - we only need to change tests now that system namespaces are being filtered out when no discovery selectors are set.

### DIFF
--- a/molecule/accessible-namespaces-test/converge.yml
+++ b/molecule/accessible-namespaces-test/converge.yml
@@ -13,17 +13,20 @@
     vars:
       namespace_list: []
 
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
This looks good.  I started on minikube with just the usual things + `bookinfo` and all of the namespaces were visible.  I added `openshift-foo` and `openshft-foo` and only the misspelled one appeared.  I think created a DS that allowed only `bookinfo` and `openshift-foo`, and after reconciling that's all I saw, plus `istio-system`.  Bully!

PR #814, opened 2024-09-03
install jmespath py library as root

jmazzitelli:
We are seeing issues on OpenShift not finding this jmespath library. I don't know if this fixes the problem, but installing that py library as root might be a solution. Need to test this.


hhovsepy:
works fine on jenkins pipelines on OCP cluster

jmazzitelli:
OK. @hhovsepy you can approve and merge this then. Let's get this in now.

hhovsepy (approved):

PR #816, opened 2024-09-04
Add READ permissions for mutating webhooks

nrfox:
Necessary for https://github.com/kiali/kiali/pull/7695

PR #817, opened 2024-09-12
fix typo that was missed in PR #816

jmazzitelli:
related to https://github.com/kiali/kiali-operator/pull/816


jmazzitelli:
Only one instance of the typo - so this PR fixes that one:

```
$ grep -R mutatating *
roles/default/kiali-deploy/templates/kubernetes/role-viewer.yaml:  - mutatatingwebhookconfigurations
```

nrfox (approved):

PR #821, opened 2024-09-21
include task has been deprecated in ansible - use include_tasks instead

jmazzitelli:
I don't know how this one sneaked through, but all others have already been replaced. This needs to be replaced, too.

This isn't necessary right now (so no need to backport) but this will be needed in the future when we get a new ansible operator base image that has Ansible 2.16+ in it. So let's get this in now.


ferhoyos (approved):
Not tested, but the change LGTM. The rest of the ansible files use `include_tasks` except these three files.

PR #823, opened 2024-09-25
Certificate information indicators removal

leandroberetta:
Given that Kiali will search the certificate information in a config map, this features is no longer required.


jmazzitelli:
After resolving the conflicts locally, I was able to run the "default" molecule test as a basic smoke test on minikube... all worked.

jmazzitelli (approved):
LGTM

PR #822, opened 2024-09-25
support namespaces with just numerals in their names

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7773

PR #824, opened 2024-09-25
[v1.89] support namespaces with just numerals in their names

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7773

cherry pick of https://github.com/kiali/kiali-operator/pull/822

test procedures are the same for https://github.com/kiali/kiali-operator/pull/822 except you just want to try setting `spec.version` to other supported versions (I tried all of them and all worked: `v1.89`, `v1.73`, `v1.65`, `v1.57`) and set `accessible_namespaces` to include `12345` rather than set discovery selectors:
```yaml
spec:
  deployment:
    accessible_namespaces: ["12345"]
```


aljesusg (approved):
LGFM

PR #827, opened 2024-10-02
[molecule] operator no longer creates the second role in control plane ns

jmazzitelli:
that second role/binding was removed as part of this work https://github.com/kiali/kiali/issues/7456


leandroberetta (approved):

jmazzitelli:
all molecule tests pass now

PR #829, opened 2024-10-16
be able to set custom env vars on server pod

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7569


jshaughn (approved):
I tested this the manual way and it worked.

PR #834, opened 2024-10-22
clean up some stuff now that istio secrets are no longer accessed

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7855

I checked the helm charts, everything is purged from over there. Looks like we just missed OLM metadata and the operator defaults.


leandroberetta (approved):
lgtm

PR #837, opened 2024-11-05
Add FBC resources for kiali operator

jmazzitelli:
This is replicating what was done here: https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/5455


jmazzitelli:
This does NOT include 2.1 in the new FBC stuff. This is just representative of the currently released metadata (which goes up to 2.0). I will work on figuring out what needs to be done for new versions that get introduced.

PR #838, opened 2024-11-05
add 2.1.0 to catalogs; when creating new versions, update the catalog templates

jmazzitelli:
This is to support FBC (file-based catalogs) which is being forced upon us.

Note: yq was used to generate the new stuff in the templates; yq indents differently than the on-boarding make target that was used to originally generate the templates. That's why there looks to be a large diff here, but its mainly whitespace. Do a "git diff -w" to see the true diff which simply adds 2.1.0 metadata. Note that the use of `yq` is taken directly from the [OLM FBC documentation here](https://olm.operatorframework.io/docs/advanced-tasks/catalog-update-formulary/#adding-a-new-replaces-link-between-two-existing-bundles), hence why it will be used moving forward to update the catalog templates.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -251,10 +251,11 @@ jobs:
     - name: Create or update version branch      
       run: git push origin $(git rev-parse HEAD):refs/heads/$BRANCH_VERSION
 
-    - name: Configure Operator SDK
+    - name: Configure Operator SDK and OPM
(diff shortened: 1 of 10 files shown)


jmazzitelli:
This works, I tested the action on my own fork. See the github action results here: https://github.com/jmazzitelli/kiali-operator/actions/runs/11696385693 which resulted in this PR here (which, again, is all correct): https://github.com/jmazzitelli/kiali-operator/pull/2

jshaughn (approved):
LG2M

PR #848, opened 2024-11-14
[Tempo] Add tempo cache config option

josunect:
Added a Tempo cache, these 2 options will allow to: 
- disable/enable
- Modify cache capacity

Ref PR: https://github.com/kiali/kiali/pull/7854

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1102,7 +1102,7 @@ spec:
                             description: "When `cache_enabled` is true, the number of traces saved in the cache. 200 by default."
                             type: integer
                           cache_enabled:
-                            description: "A FIFO cache with the last 'cache_capacity' traces viewed."
+                            description: "A FIFO cache with the last `cache_capacity` traces viewed."
                             type: boolean
                           datasource_uid:
                             description: "The unique identifier (uid) of the Tempo datasource in Grafana."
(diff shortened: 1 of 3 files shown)


crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1098,6 +1098,12 @@ spec:
                         description: "Settings used to configure the access url to the Tempo Datasource in Grafana."
                         type: object
                         properties:
+                          cache_capacity:
+                            description: "When 'cache_enabled' is true, the number of traces saved in the cache. 200 by default."
```
jmazzitelli, line 1102: use back-ticks so when we publish this on kiali.io, it uses the emphasized font we use for all the other settings
```suggestion
                            description: "When `cache_enabled` is true, the number of traces saved in the cache. 200 by default."
```

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1098,6 +1098,12 @@ spec:
...
+                            type: integer
+                          cache_enabled:
+                            description: "A FIFO cache with the last 'cache_capacity' traces viewed."
```
jmazzitelli, line 1105:
```suggestion
                            description: "A FIFO cache with the last `cache_capacity` traces viewed."
```

jmazzitelli (changes requested):

jmazzitelli (approved):

PR #851, opened 2024-11-22
Add new graph setting for the animation style

jshaughn:
Part-of https://github.com/kiali/kiali/issues/7934

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1255,7 +1255,7 @@ spec:
                               animation:
                                 description: "The traffic animation style. Value must be one of: `dash` or `point`. Default is `point`."
                                 type: string
-                                enum: ["dash", "point" ]
+                                enum: ["dash", "point"]
                           traffic:
                             description: "These settings determine which rates are used to determine graph traffic."
                             type: object
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1248,6 +1248,14 @@ spec:
                                 expression:
                                   type: string
                                   description: "The hide expression."
+                          settings:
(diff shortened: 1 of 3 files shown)


crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1248,6 +1248,14 @@ spec:
...
                                   type: string
                                   description: "The hide expression."
+                          settings:
+                            description: "Various presentation options."
+                            type: object
+                            properties:
+                              animation:
+                                description: "The traffic animation style. Value must be one of: `dash` or `point`. Default is `point`."
+                                type: string
+                                enum: ["dash", "point" ]
```
jmazzitelli, line 1258: Egad! No leading space in the front, but leading space at the end? UG-LY.

```suggestion
                                enum: ["dash", "point"]
```

jmazzitelli (changes requested):
You forgot to add the default to the example CR, and I just noticed we are missing all of the settings here. So you should add them. This example CR file is used for 2 things  (a) for an example CR that users can refer to, each value in the example CR file is the default, so users can use this as a reference for all default values, and
(b) this file is used to test the schema validation - our CI action runs a validation test on the example CR which confirms the schema and example files are both valid.

So in this file: https://github.com/kiali/kiali-operator/blob/v2.1.0/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L437

you need to add this:
```yaml
        settings:
          animation: "point"
          font_label: 13
          min_font_badge: 7
          min_font_label: 10
```

jshaughn:
> You forgot to add the default to the example CR, and I just noticed we are missing all of the settings here

I will add the new one, but the others are actually purposefully unadvertised, and also will go away with the cytoscape deprecation.

jmazzitelli (approved):
validation passes. lgtm

PR #855, opened 2024-12-11
[2.1]Change To candidate

aljesusg:
I have some doubts...

- We not need to change it in the community or upstream. right?
-  Do we need to change it here? 

https://github.com/kiali/kiali-operator/blob/v2.1/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L69

### DIFF
--- a/manifests/kiali-ossm/bundle.Dockerfile
+++ b/manifests/kiali-ossm/bundle.Dockerfile
@@ -4,8 +4,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=kiali-ossm
-LABEL operators.operatorframework.io.bundle.channels.v1=stable
-LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
+LABEL operators.operatorframework.io.bundle.channels.v1=candidate
+LABEL operators.operatorframework.io.bundle.channel.default.v1=candidate
 
 COPY manifests /manifests/
 COPY metadata /metadata/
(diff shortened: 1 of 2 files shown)


jmazzitelli:
No need to change in community (which we aren't publishing to anymore) or upstream (which only has our sprint releases, nothing to do with TPs or things like that).

This PR only affects the 2.1 branch, so I guess it is OK. I really do not know what will happen when we release the OSSM 3 GA version of Kiali - that will still only have "stable" channel. That is probably what we will want, but I do not know how OLM behaves when you have different channels for different builds like this.

jmazzitelli (approved):

jshaughn:
> No need to change in community (which we aren't publishing to anymore) or upstream (which only has our sprint releases, nothing to do with TPs or things like that).

If it's easier to make the change in upstream github, and pull it down, that should also be fine.  It is scoped to OSSM.

> This PR only affects the 2.1 branch, so I guess it is OK. I really do not know what will happen when we release the OSSM 3 GA version of Kiali - that will still only have "stable" channel. 

I'm fairly sure it won't impact 3.0 GA in any way.  The 3.0 operator won't even know about 2.1 in it's "previous versions".

jshaughn (approved):
LG!

jmazzitelli:
> The 3.0 operator won't even know about 2.1 in it's "previous versions".

FWIW, it isn't OUR operator I'm worried about. It's OLM itself. It's going to have this Kiali operator of this specific version with some metadata about a different channel. When we publish a new Kiali operator with a different version and different metadata, how will OLM like it? Maybe it will work, I dunno. Have we tested this? No.

aljesusg:
> > The 3.0 operator won't even know about 2.1 in it's "previous versions".
> 
> FWIW, it isn't OUR operator I'm worried about. It's OLM itself. It's going to have this Kiali operator of this specific version with some metadata about a different channel. When we publish a new Kiali operator with a different version and different metadata, how will OLM like it? Maybe it will work, I dunno. Have we tested this? No.

I'm worried if this isn't going to work.

jshaughn:
If it doesn't work as-is then perhaps we need to remove all previous version information so that this 2.1 is really just an island.

PR #856, opened 2024-12-16
increase signing key length

jmazzitelli:


jmazzitelli:
To test, push the operator dev build to a cluster and deploy Kiali, then confirm the auto-generated signing key has a length of 32 now:

_(The following assumes you have OpenShift cluster - but you can easily test this on, say, minikube if you set CLUSTER_TYPE appropriately.)_

1. Build everything, deploy operator, install Kiali:
```
make build build-ui cluster-push operator-create kiali-create
```
2. Confirm the signing key is the new length of 32:
```
kubectl get secret kiali-signing-key -n istio-system -ojsonpath='{.data.key}' | base64 -d | wc -c
```
That command should output "32" to indicate the signing-key string has a length of 32. You can see what the signing-key looks like via:
```
kubectl get secret kiali-signing-key -n istio-system -ojsonpath='{.data.key}' | base64 -d ; echo
```
That returns the random string that will look something like `rsuYTY156qMaqrCZp7Yj8LqgUqPikV33`

jshaughn (approved):
Followed the directions and the key looks as expected:
```
$ kubectl get secret kiali-signing-key -n istio-system -ojsonpath='{.data.key}' | base64 -d | wc -c
32
```

PR #860, opened 2024-12-27
do not create things that need route if ingress is disabled

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8023

### DIFF
--- a/molecule/common/tasks.yml
+++ b/molecule/common/tasks.yml
@@ -111,6 +111,8 @@
   when:
   - is_openshift == True
   - kiali_configmap.deployment.remote_cluster_resources_only == False
+  - kiali_route is defined
+  - kiali_route.resources | length > 0
 
 # To avoid problems with Ingress/Minikube conflicts, if installing multiple kiali instances set web_root to the instance name
 - name: Determine the Kiali Ingress URL on minikube
--- a/molecule/os-console-links-test/converge.yml
+++ b/molecule/os-console-links-test/converge.yml
@@ -13,6 +13,20 @@
       namespace_list:
       - "{{ istio.control_plane_namespace }}"
 
+  # The ingress is disabled at the start of this test - without this, the Route is not created and thus no ConsoleLinks are created
+  - name: Assert that there is no Route (and no ConsoleLinks) because the ingress was disabled
+    assert:
+      that:
+      - "{{ kiali_route.resources | length == 0 }}"
+      - "{{ query('k8s', kind='ConsoleLink', api_version='console.openshift.io/v1', errors='warn') | length == 0 }}"
+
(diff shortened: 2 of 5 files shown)


jmazzitelli:
This needs to skip creating the clusterrole/bindings for the OAuthClient - probably just need to add it to "Process OpenShift OAuth client" task. I'll work on that next

jmazzitelli:
The molecule tests that should show this doesn't break anything:

1. Get an OpenShift cluster - make sure you have user credentials kiali/kiali (use `hack/crc-openshift.sh` to make it easy)
2. Install Istio (e.g. `hack/istio/install-istio-via-istioctl.sh -cp openshift`)
3. `cd` to your ossmc local repo and build and push the ossmc plugin image (one of the molecule tests installs it): `make cluster-push`
4. Build and push server/operator dev images: `make build build-ui cluster-push`
5. Run all molecule tests: `./hack/run-molecule-tests.sh --client-exe "$(which oc)" --cluster-type openshift -udi true`

jmazzitelli:
After you run the test instructions in the previous comment, you can do additional manual testing:

1. Install operator and server: `make operator-create kiali-create`
2. Confirm everything is installed: `oc get pods -n istio-system -l app.kubernetes.io/name=kiali`
3. Patch the Kiali CR such that the ingress is disabled: `oc patch kiali kiali -n kiali-operator --type=json -p='[{"op": "replace", "path": "/spec/deployment/ingress/enabled", "value": false}]'`
4. Check that the operator aborted the reconciliation with the correct error message:
```
$ oc get kiali kiali -n kiali-operator -o jsonpath='{.status}' | jq

...
      "lastTransitionTime": "2024-12-29T16:15:22Z",
      "message": "The auth.strategy is 'openshift' which requires a Route, but deployment.ingress.enabled is false. Aborting.",
      "reason": "Failed",
      "status": "True",
      "type": "Failure"
...
```
5. Now patch the Kiali CR and turn the auth.strategy to `anonymous`: `oc patch kiali kiali -n kiali-operator --type=json -p='[{"op": "replace", "path": "/spec/auth/strategy", "value": "anonymous"}]'`
6. The install should succeed - you should see no errors in the Kiali CR status and there should be a Kiali server pod that has just been created:
```
$ oc get kiali kiali -n kiali-operator -o jsonpath='{.status}' | jq
...<there should be no errors>...
$ oc get pods -n istio-system -l app.kubernetes.io/name=kiali
NAME                     READY   STATUS    RESTARTS   AGE
kiali-668c44b795-rkb4z   1/1     Running   0          10s
```

josunect (approved):
LGTM:

Molecule tests passed locally in crc: 

image

Manual tests with expected results: 
Enable ingress:
image

anonymous strategy:
image

kiali pod:
image

PR #865, opened 2025-01-13
be able to configure /api endpoint to require credentials

jmazzitelli:
When auth.strategy is not anonymous, users can still see the /api endpoint version and config information. This new config allows users to disable that unauthenticated access, requiring users to authenticate with the server in order to see the /api info.

```yaml
server:
  require_auth: true
```

server PR: https://github.com/kiali/kiali/pull/8055

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1490,6 +1490,9 @@ spec:
                       enabled:
                         description: "When 'true', the profiler will be enabled and accessible at /debug/pprof/ on the Kiali endpoint."
                         type: boolean
+                  require_auth:
+                    description: "When true, the /api endpoint will require users to authenticate themselves. When false, users need not authenticate with Kiali in order to get basic runtime info about the server via the /api endpoint. This setting is ignored if auth.strategy is 'anonymous'."
+                    type: boolean
                   web_fqdn:
                     description: "Defines the public domain where Kiali is being served. This is the 'domain' part of the URL (usually it's a fully-qualified domain name). For example, `kiali.example.org`. When empty, Kiali will try to guess this value from HTTP headers. On non-OpenShift clusters, you must populate this value if you want to enable cross-linking between Kiali instances in a multi-cluster setup."
                     type: string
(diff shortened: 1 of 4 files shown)


jmazzitelli:
for test procedures, see https://github.com/kiali/kiali/pull/8055#issuecomment-2588309549

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1490,6 +1490,9 @@ spec:
                       enabled:
                         description: "When 'true', the profiler will be enabled and accessible at /debug/pprof/ on the Kiali endpoint."
                         type: boolean
+                  require_auth:
+                    description: "When true, the /api endpoint will require users to authenticate themselves. When false, users need not authenticate with Kiali in order to get basic runtime info about the server via the /api endpoint. This setting is ignored if auth.strategy is 'anonymous'."
```
jshaughn, line 1494:
```suggestion
                    description: "When true, the /api endpoint (and all other endpoints), will require users to authenticate themselves. When false, users need not authenticate with Kiali in order to get basic runtime info about the server via the /api endpoint. This setting is ignored if auth.strategy is 'anonymous'."
```

jshaughn (approved):
Approved but I suggest one small doc change.

crd-docs/crd/kiali.io_kialis.yaml
jmazzitelli: That is not true. NOT all other endpoints... specifically, there are /api/auth/XX endpoints that still are (and are required to me) unauthorized.

Right now this fix is very specific and only affects that single `api/` endpoint

jshaughn: hmmm, OK.  Then maybe I don't love the setting name.  But not worth changing.

jmazzitelli: I thought about the  name of the setting, but I was expecting something in the future where that setting might also affect (like maybe we add some other "version" or "status" time endpoints that we also want unauth'ed by default but allow a user to put it behind auth. So I tried to name it more generic as opposed to something like "make_api_endpoint_secure" or something like that.

PR #876, opened 2025-02-06
Remove sidecar injector setting from defaults

nrfox:
This should not be hardcoded. It is autodetected and when you set this to a specific value, Kiali can only get the sidecar injector configmap for a single revision at a time.

### DIFF
--- a/roles/v2.4/kiali-deploy/defaults/main.yml
+++ b/roles/v2.4/kiali-deploy/defaults/main.yml
@@ -181,7 +181,6 @@ kiali_defaults:
       istio_identity_domain: "svc.cluster.local"
       istio_injection_annotation: "sidecar.istio.io/inject"
       istio_sidecar_annotation: "sidecar.istio.io/status"
-      istio_sidecar_injector_config_map_name: "istio-sidecar-injector"
       istiod_pod_monitoring_port: 15014
       root_namespace: ""
     prometheus:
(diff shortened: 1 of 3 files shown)


nrfox:
This needs backport to 2.4

roles/v2.4/kiali-deploy/defaults/main.yml
```diff
@@ -181,7 +181,6 @@ kiali_defaults:
       istio_identity_domain: "svc.cluster.local"
       istio_injection_annotation: "sidecar.istio.io/inject"
       istio_sidecar_annotation: "sidecar.istio.io/status"
-      istio_sidecar_injector_config_map_name: "istio-sidecar-injector"
```
jmazzitelli, line 184: we need to delete this from the "default" role, too, correct?

https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L185

jmazzitelli (changes requested):
as per the [checklist](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting), need to update the example CR now that we are changing the default: https://github.com/kiali/kiali-operator/blob/v2.5.0/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L340

This:
```
      istio_sidecar_injector_config_map_name: "istio-sidecar-injector"
```

should now be this:

```
      # default: istio_sidecar_injector_config_map_name is undefined
```

When we release in the next sprint, this change will end up in the generated example that goes in kiali.io docs here: https://kiali.io/docs/configuration/kialis.kiali.io/#example-cr

jmazzitelli:
https://github.com/nrfox/kiali-operator/pull/1

nrfox:
> [nrfox#1](https://github.com/nrfox/kiali-operator/pull/1)

Merged thank you!

jmazzitelli (approved):

PR #892, opened 2025-03-19
optionally mount secret volume kiali-multi-cluster-secret

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7877


jmazzitelli:
I ran the molecule test for remote cluster secrets and it still passes, so nothing broke:

```
           remote-cluster-resources-test... success [3m 36s]
```

jmazzitelli:
To test this, create a remote cluster secret and put it in the namespace where Kiali will be deployed (e.g. `istio-system`). Make the name of the secret `kiali-multi-cluster-secret`. It does not have to have any labels or annotations on it. You can have any number of clusters defined in your secret.

See my example in the issue comment for details on getting this secret and what it can look like: https://github.com/kiali/kiali/issues/7877#issuecomment-2737571621

Then install Kiali CR and see that the remote cluster secret is loaded.

PR #894, opened 2025-03-20
Remove cytoscape related fields

jshaughn:
- remove graph.impl
- remove ui_defaults.graph.settings.font_label
- remove ui_defaults.graph.settings.min_font_badge
- remove ui_defaults.graph.settings.min_font_label
- change ossmc molecule test to update a different setting

part of https://github.com/kiali/kiali/issues/8053

### DIFF
--- a/molecule/ossmconsole-config-values-test/converge.yml
+++ b/molecule/ossmconsole-config-values-test/converge.yml
@@ -60,4 +60,6 @@
 
   - name: Confirm the servicePort is now changed to 20001
     assert:
-      that: "{{ ossmconsole_configmap.servicePort == 20001 }}"
+      that:
+      - ossmconsole_consoleplugin.spec.proxy[0].endpoint.service.port == (current_ossmconsole_cr.status.kiali.servicePort | int)
+      - ossmconsole_consoleplugin.spec.proxy[0].endpoint.service.port == 20001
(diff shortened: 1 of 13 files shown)


molecule/ossmconsole-config-values-test/converge.yml
```diff
@@ -35,20 +35,16 @@
...
   # This test will change some config settings to make sure things work like we expect.
   # We alter the current CR with new config and deploy that new CR.
 
   # Change the existing CR to get a new config
 
-  - name: Set to use cy graph impl
+  - name: Set to use different service port
     vars:
-      new_graph_impl: "cy"
+      new_service_port: 9444
```
jmazzitelli, line 45: I didn't run it - but if 9444 isn't valid, this is probably going to fail somewhere when trying to install OSSMC and thus the molecule test will fail because it won't see that the OSSMC CR processed successfully.

Did you run the molecule test? Just start CRC, build/push the dev images, install Istio, and run the test:

1. Start CRC: `hack/crc-openshift.sh start -p $HOME/openshift-pull-secret.txt` (note: if you don't have the pull secret in a file, I think you can omit the -p option and CRC will ask for it on the command line)
2. Login to the cluster: `oc login -u kubeadmin -p kiali https://api.crc.testing:6443 --insecure-skip-tls-verify`
3. Login to the cluster image registry: `podman login --tls-verify=false -u kiali -p $(oc whoami -t) default-route-openshift-image-registry.apps-crc.testing`
4. Install Istio: `hack/istio/install-istio-via-istioctl.sh -cp openshift`
5. Build the dev images and push them to the cluster: `make CLUSTER_TYPE=openshift build build-ui cluster-push`
6. Run the molecule test that you changed: `./hack/run-molecule-tests.sh --client-exe "$(which oc)" --cluster-type openshift -udi true -at ossmconsole-config-values-test`

jmazzitelli (changes requested):
I think you need to change the service port (maybe even just setting to 20001 would work)

jmazzitelli (approved):

PR #902, opened 2025-04-25
Added holdApplicationUntilProxyStarts in pod_annotations

hhovsepy:
Make sure pod_annotations `proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'` is always set to be ready to install Kiali on auto-injection=enabled namespaces.

Issue https://github.com/kiali/kiali/issues/8259

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -215,7 +215,7 @@ spec:
       nodeSelector: "nodeSelectorValue"
     # default: pod_annotations is empty
     pod_annotations:
-      podAnnotation: "podAnnotationValue"
+      proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'
     # default: pod_labels is empty
     pod_labels:
       sidecar.istio.io/inject: "true"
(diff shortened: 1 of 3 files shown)


jmazzitelli:
Note - we need to add to the docs that if the user adds their own pod_annotations - it will overwrite this default. They may want to override it for whatever reason, but if they do not, they have to know they need to add it with the rest of their pod annotations.

So make sure you mention this here: https://github.com/kiali/kiali-operator/blob/v2.8.0/crd-docs/crd/kiali.io_kialis.yaml#L601

jmazzitelli:
BTW: I like how you plan on doing this (adding it as a default rather than hardcoding it in the template (because if you hardcoded it, it will ALWAYS be there and can NEVER be removed)... I'm all about configurability because you never know in the future if something breaks because that annotation exists and there is no way for the user to remove it. The way you are doing it here, the annotation is removable.

hhovsepy:
@jmazzitelli thanks for the review, have tuned up.
Is it necessary to backport to older revisions?

jmazzitelli (approved):
looks good!

And, no, we do not need to backport. We have a workaround that users on older versions can do to fix the problem (they just have to define the annotation themselves).

PR #903, opened 2025-04-25
Add kiali_internal unstructured section

josunect:
Ref. https://github.com/kiali/kiali/issues/8292

### DIFF
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1427,6 +1427,10 @@ spec:
                         description: "The KIA0301 validation checks duplicity of host and port combinations across all Istio Gateways. This includes also Gateways with '*' in hosts. But Istio considers such a Gateway with a wildcard in hosts as the last in order, after the Gateways with FQDN in hosts. This option is to skip Gateways with wildcards in hosts from the KIA0301 validations but still keep Gateways with FQDN hosts."
                         type: boolean
 
+              kiali_internal:
+                description: "Unstructured section for internal testing and debugging features"
+                type: object
+
               kubernetes_config:
                 description: "Configuration of Kiali's access of the Kubernetes API."
                 type: object
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1430,6 +1430,7 @@ spec:
               kiali_internal:
                 description: "Unstructured section for internal testing and debugging features."
                 type: object
+                x-kubernetes-preserve-unknown-fields: true
 
               kubernetes_config:
                 description: "Configuration of Kiali's access of the Kubernetes API."
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1428,7 +1428,7 @@ spec:
                         type: boolean
 
               kiali_internal:
-                description: "Unstructured section for internal testing and debugging features"
+                description: "Unstructured section for internal testing and debugging features."
                 type: object
 
               kubernetes_config:


jmazzitelli (changes requested):
The schema is not valid. Here's how you can check the schema:

1. In the example CR (`crd-docs/cr/kiali.io_v1alpha1_kiali.yaml`) make a local change (don't commit this, just change this on your local machine) by adding a kiali_internal section.. for example, I did this:
```diff
diff --git a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
index a4dc39c..c7cfe31 100644
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -10,6 +10,9 @@ spec:
     annotation: "kiali.io/api-spec"
     icon_annotation: "kiali.io/api-type"
 
+  kiali_internal:
+    foo: bar
+
   installation_tag: ""
 ```

2. Have a kubernetes cluster running (I was on minikube). And then run the "make validate-cr" target (this make target is in the operator git repo - so run this after you "cd" to your operator repo directory):
```
make validate-cr
```
This error results (`unknown field "spec.kiali_internal.foo"`):
```
$ make validate-cr 

========== Validating the Kiali CR ==========
/home/jmazzite/source/kiali/kiali-operator/crd-docs/bin/validate-kiali-cr.sh --kiali-cr-file /home/jmazzite/source/kiali/kiali-operator/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
=== SETTINGS ===
KIALI_CRD_LOCATION=
KIALI_CR_FILE=/home/jmazzite/source/kiali/kiali-operator/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
KIALI_CR_NAME=
NAMESPACE=default
PRINT_CRD=false
=== SETTINGS ===
Using 'oc' located here: /home/jmazzite/bin/oc
...
Validating the CR:
----------
Error from server (BadRequest): error when creating "STDIN": TestKiali in version "v1alpha1" cannot be handled as a TestKiali: strict decoding error: unknown field "spec.kiali_internal.foo"
----------
ERROR! Validation failed for Kiali CR [/home/jmazzite/source/kiali/kiali-operator/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml]
make: *** [Makefile:92: validate-cr] Error 1
```

The reason is the new kiali_internal field does not indicate what kind of object it is (it needs to be `x-kubernetes-preserve-unknown-fields: true` because we are going to allow this field to be any valid yaml).

So the diff to fix this is:
```diff
index 6d5715c..35307b7 100644
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1430,6 +1430,7 @@ spec:
               kiali_internal:
                 description: "Unstructured section for internal testing and debugging features."
                 type: object
+                x-kubernetes-preserve-unknown-fields: true
 
               kubernetes_config:
                 description: "Configuration of Kiali's access of the Kubernetes API."
```

With that in place, now run the `make validate-cr` and see that it is successful.

josunect:
Thanks, @jmazzitelli ! I did the `make validate-cr` but without any of the new fields, now I learn how to validate properly.

jmazzitelli:
> Thanks, @jmazzitelli ! I did the `make validate-cr` but without any of the new fields, now I learn how to validate properly.

Yeah, that's why when we create new settings, we add the default values (or some value) to the example CR - so the validate-cr test runs and makes sure the schema is correct (that gets run by GitHub Actions for operator PRs).

But this is a very special case where we don't even want to show the kiali_internal in the example CR (since it really isn't for users to use) so that's why this wasn't caught.

jmazzitelli (approved):
lgtm

PR #905, opened 2025-04-28
Fix the pod_annotations

hhovsepy:
Fixing error in kiali-operator pod:
` 'dict object' has no attribute 'pod_annotations'`


jshaughn (approved):
LG

PR #908, opened 2025-04-29
fix pod_annotations null test - needs to fallback to the default

jmazzitelli:
This fixes the problem detected by the null-values-test molecule test. When a user wants to delete their pod_annotations from the Kiali CR, we need to ensure we fallback to the default value (which is the new annotation we just added).

To see this work now:

1. Start minikube, install Istio:
```
./hack/k8s-minikube.sh start && hack/k8s-minikube.sh istio
```
2. Build and push the dev images into minikube (make sure this operator PR is checked out into your local operator branch):
```
make CLUSTER_TYPE=minikube build build-ui test cluster-push
```
3. Now run the molecule test, making sure to pass `-udi true` so the test uses the dev image you just pushed. If you set `-udi false` you will see the test fail (because its using the latest image which doesn't have the fix).
```
./hack/run-molecule-tests.sh --client-exe "$(which kubectl)" --cluster-type minikube -at "null-cr-values-test" -udi true
```

### DIFF
--- a/molecule/null-cr-values-test/converge.yml
+++ b/molecule/null-cr-values-test/converge.yml
@@ -16,12 +16,12 @@
   # Assert that we have defaults since the CR set most settings to null.
   # The defaults come from the operator's roles/kiali-deploy/defaults/main.yaml
   # This just checks some of the settings to see that defaults are getting set.
-  # Note that accessible namespaces is being set to "[**]" so some settings depend on that.
   - assert:
       that:
       - kiali_configmap.installation_tag == ""
       - kiali_configmap.additional_display_details | length == 1
       - kiali_configmap.custom_dashboards | length == 0
+      - kiali_configmap.deployment.configmap_annotations | length == 0
       - kiali_configmap.deployment.discovery_selectors | length == 0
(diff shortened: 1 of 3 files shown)


hhovsepy (commented):
Regression tested on OCP cluster via Jenkins job.
However do not have local minikube to run molecule tests.

jmazzitelli:
> However do not have local minikube to run molecule tests

you can run molecule tests on OCP:

1. `oc login` to the cluster
2. Build and push the dev images into the cluster (make sure this operator PR is checked out into your local operator branch):
```
make build build-ui test cluster-push
```
3. Now run the molecule test, making sure to pass `-udi true` so the test uses the dev image you just pushed. If you set `-udi false` you will see the test fail (because its using the latest image which doesn't have the fix).
```
./hack/run-molecule-tests.sh --client-exe "$(which oc)" --cluster-type openshift -at "null-cr-values-test" -udi true
```

hhovsepy (approved):
Verified on OCP cluster:
Screenshot from 2025-04-30 12-29-10

PR #916, opened 2025-05-25
refactor to not use kubernetes.core.k8s_cluster_info task

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8459


jshaughn (approved):
This looks OK to me.  It looks like we do already have some HPA testing in the molecule tests, so I think we're good.

PR #922, opened 2025-06-11
add v2.11 support

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8504

checklists:
* https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-adding-support-for-a-new-ansible-role-version


jmazzitelli:
To test, create a Kiali CR with `spec.version` of `default`. After it installs, edit the CR and change `spec.version` to `v2.4`. After that installs, change `spec.version` to `v2.11`. This should all work.

jshaughn (approved):
Sanity check looks OK

PR #924, opened 2025-06-11
need to set CWA default in remove role

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8505

### DIFF
--- a/roles/default/kiali-remove/defaults/main.yml
+++ b/roles/default/kiali-remove/defaults/main.yml
@@ -2,6 +2,7 @@ kiali_defaults_remove:
   istio_namespace: ""
 
   deployment:
+    cluster_wide_access: true
     hpa:
       api_version: ""
     instance_name: "kiali"
--- a/roles/v2.4/kiali-remove/defaults/main.yml
+++ b/roles/v2.4/kiali-remove/defaults/main.yml
@@ -2,6 +2,7 @@ kiali_defaults_remove:
   istio_namespace: ""
 
   deployment:
+    cluster_wide_access: true
     hpa:
       api_version: ""
     instance_name: "kiali"


jmazzitelli:
how to test:

1. Install a cluster with Istio
2. build, push and install the operator: `make CLUSTER_TYPE=minikube cluster-push-operator operator-create`
3. Create a Kiali CR with `spec.version` set to `default` (for this test, `deployment.image_version` needs to be `latest`):
```sh
cat <<EOM | kubectl apply -f -
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  name: kiali
  namespace: istio-system
spec:
  version: default
  auth:
    strategy: anonymous
  deployment:
    image_version: latest
EOM
```
4. Now change spec.version to `v2.4`:
```sh
kubectl patch kiali kiali -n istio-system --type=merge -p '{"spec":{"version":"v2.4"}}'
```
5. After the operator finishes, grep its output looking for the task that was spitting out the error and confirm no error is getting logged now:
```sh
kubectl logs -n kiali-operator deployments/kiali-operator | grep -A 4 "Get namespaces that have Kiali roles in them"
```
That command should return something like this (notice no error now - because CWA is defaulted to true now, that task is just skipped normally):
```
TASK [default/kiali-remove : Get namespaces that have Kiali roles in them] *****
skipping: [localhost] => {"changed": false, "false_condition": "kiali_vars_remove.deployment.cluster_wide_access == False", "skip_reason": "Conditional result was False"}

TASK [default/kiali-remove : Delete Kiali roles] *******************************
skipping: [localhost] => {"changed": false, "false_condition": "namespaces_with_kiali_roles is defined", "skip_reason": "Conditional result was False"}
```

PR #927, opened 2025-06-18
Add UI defaults trace limit

josunect:
Ref. 
https://github.com/kiali/kiali/issues/8517
https://github.com/kiali/kiali/pull/8521


josunect:
> I know this is still in draft, but so far this looks good and can be approved if everything works on server side.

Thanks, @jmazzitelli ! 

There is a discussion about limiting the values to the same ones in the UI, in that case I think that should be also included here.

jmazzitelli (approved):

josunect:
Waiting for server PR to be merged

PR #928, opened 2025-06-20
Add schemas to CRDs, move to CRD version v1

jmazzitelli:


jmazzitelli:
> we are going to do this we might as well write/deploy a conversion webhook with the operator. It would be the same amount of work but a much better user experience.

But then that now becomes a maintenance burden on us going forward when we only need this for a one-time event. I really don't want to start shipping another productized binary and more OLM configuration for something we hardly need.

jmazzitelli:
another molecule test to fix...  `roles-test` failed with:

```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"Kiali.kiali.io \\\\\"kiali\\\\\" is invalid: spec.deployment.discovery_selectors.default: Invalid value: \\\\\"string\\\\\": spec.deployment.discovery_selectors.default in body must be of type array: \\\\\"string\\\\\"\",\"reason\":\"Invalid\",\"details\":{\"name\":\"kiali\",\"group\":\"kiali.io\",\"kind\":\"Kiali\",\"causes\":[{\"reason\":\"FieldValueTypeInvalid\",\"message\":\"Invalid value: \\\\\"string\\\\\": spec.deployment.discovery_selectors.default in body must be of type array: \\\\\"string\\\\\"\",\"field\":\"spec.deployment.discovery_selectors.default\"}]},\"code\":422}\\n'", "reason": "Unprocessable Entity"}
```

nrfox:
> But then that now becomes a maintenance burden on us going forward when we only need this for a one-time event. I really don't want to start shipping another productized binary and more OLM configuration for something we hardly need.

Maintenance burden either falls on us or on users. But besides we can get around all of that by keeping the version the same (v1alpha1) and just shipping a schema for the version. Users won't need to remove their CRs and fields that we remove when we ship the schema will be dropped through pruning: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning

jmazzitelli:
checkpoint - latest has a rebase and setting of defaults/constraints in the CRD - molecule tests still pass.

Now that there is a baseline schema, next up is to see what we can move around, change, or eliminate. This is the time to look at the Kiali CR and see what changes we want to make. Remember, we want to make this schema easily extensible in the future without requiring a bump in schema version.

jmazzitelli:
> > look at the Kiali CR and see what changes we want to make
> 
> There will be some changes based on the external-kiali work, so we may want to wait for that to be completed before releasing this into the wild.

Right - this will NOT be going in this sprint release (notice I did not attach this PR or the helm PR to the current Github Project).

jmazzitelli:
After re-organizing the spec.deployment and spec.server settings of the CRD, all molecule tests pass.

However, I'm not convinced we want to do this - the changes are huge, and I'm not sure if the gain is worth the potential pain. I don't think we have full 100% coverage of all these settings and I'm afraid this is missing something and we won't know until this is out in the field. It might be best to not re-arrange the settings (it would also avoid some pain to end users from having to convert their old CRs). But I will examine this further to see if I can gain more confidence in this change.

```
=====================
=== TEST RESULTS: ===
=====================

              accessible-namespaces-test... success [7m 56s]
     affinity-tolerations-resources-test... success [4m 24s]
                cluster-wide-access-test... success [9m 9s]
                      config-values-test... success [3m 4s]
                  default-namespace-test... success [2m 44s]
                            grafana-test... success [4m 5s]
                        header-auth-test... success [2m 22s]
                      instance-name-test... success [3m 42s]
                             jaeger-test... success [3m 19s]
                            metrics-test... success [4m 28s]
                     null-cr-values-test... success [2m 16s]
                only-view-only-mode-test... success [3m 9s]
                             openid-test... success [2m 24s]
           remote-cluster-resources-test... success [3m 29s]
                              roles-test... success [5m 27s]
                    rolling-restart-test... success [3m 2s]
                              token-test... success [3m 1s]
```

jmazzitelli:
I reverted back and am re-doing the re-org of the schema.

jmazzitelli:
> No IMO we shouldn't move settings with this change.

Yeah, I'm coming to that conclusion I'm going to back all of that out. Any re-organization of the schema is going to have massive changes even within the Kiali repos (kiali.io docs, helm charts, demos, Makefiles, hacks and other scripts, etc etc). Even trying to get AI to find all the places where changes are required is a challenge. :)

I don't think we'll ever want to move things around, at least on a wholesale level.

manifests/kiali-ossm/manifests/kiali.crd.yaml
```diff
@@ -21,4 +22,1707 @@ spec:
...
+                      operator itself was configured.
+                    type: string
+                  ingress:
+                    description: "Configures if/how the Kiali endpoint should be exposed externally."
+                    type: object
+                    properties:
+                      additional_labels:
+                        description: "Additional labels to add to the Ingress (or Route if on OpenShift). These are added to the labels that are created by default; these do not override the default labels."
+                        type: object
+                        x-kubernetes-preserve-unknown-fields: true
```
nrfox, line 536: Why are we preserving unknown fields for nearly all objects in the CRD?

nrfox: I don't think you need this. This is from the kube API itself:
```
          "labels": {
            "description": "Map of string keys and values that can be used to organize and categorize (scope and select) objects. May match selectors of replication controllers and services. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels",
            "type": "object",
            "additionalProperties": {
              "type": "string",
              "default": ""
            }
          },
```

nrfox: My only reason to be nitpicky is that once we publish the schema it will be hard to change it. With `x-kubernetes-preserve-unknown-fields` I think you could have something like
```
additionalLabels:
  0: false
  more:
    unknownfields
```

jmazzitelli: So, `type: object` would be all that is needed? Just remove `x-kubernetes-preserve-unknown-fields: true` .. I will test that out.

nrfox: > So, type: object would be all that is needed? Just remove x-kubernetes-preserve-unknown-fields: true .. I will test that out.

I believe so. At least that's what upstream types look like from the `/openapi/v3` endpoint.

nrfox:
I tested this out with an existing CR with extra fields and some fields with wrong types e.g. `spec.deployment.ingress.additional_labels= 123: 123` and after applying the schema the extra fields were dropped and the wrong types were coerced into the correct type e.g. `spec.deployment.ingress.additional_labels="123":"123"`.

Do we want to take the opportunity to prune deprecated fields before merging this?

jmazzitelli:
> Do we want to take the opportunity to prune deprecated fields before merging this?

Now is the time to start doing that kind of thing. If we don't need backward compatibility for a setting, and the setting is deprecated, we should consider removing them.

Everyone should take a look and suggest the changes we should make related to these deprecated or outdated settings.

jmazzitelli:
This has been rebased on master and has the CRD with schema in the 2.14 metadata.

jmazzitelli:
The golden copies of the CRDs (and the example CRs which gets placed into the kiali.io docs) are these four files - so these are the only four we need to maintain. The CRDs in other places are managed via `make sync-crds` (make `validate-crd-sync` makes sure everything is synced).

* `crd-docs/crd/kiali.io_kialis.yaml`
* `crd-docs/crd/kiali.io_ossmconsoles.yaml`
* `crd-docs/cr/kiali.io_v1alpha1_kiali.yaml`
* `crd-docs/cr/kiali.io_v1alpha1_ossmconsole.yaml`

jmazzitelli:
These existed in v1.73 that have since been removed:

- `api`
- `api.namespaces`
- `api.namespaces.exclude`
- `api.namespaces.include`
- `api.namespaces.label_selector_exclude`
- `api.namespaces.label_selector_include`
- `auth.openshift.auth_timeout`
- `auth.openshift.client_id_prefix`
- `deployment.accessible_namespaces`
- `deployment.verbose_mode`
- `external_services.grafana.in_cluster_url`
- `external_services.grafana.url`
- `external_services.istio.gateway_api_class_name`
- `external_services.tracing.in_cluster_url`
- `external_services.tracing.url`
- `kiali_feature_flags.certificates_information_indicators`
- `kiali_feature_flags.certificates_information_indicators.enabled`
- `kiali_feature_flags.certificates_information_indicators.secrets`
- `kiali_feature_flags.clustering`
- `kiali_feature_flags.clustering.autodetect_secrets`
- `kiali_feature_flags.clustering.autodetect_secrets.enabled`
- `kiali_feature_flags.clustering.autodetect_secrets.label`
- `kiali_feature_flags.clustering.clusters`
- `kiali_feature_flags.clustering.clusters[].name`
- `kiali_feature_flags.clustering.clusters[].secret_name`
- `kiali_feature_flags.clustering.kiali_urls`
- `kiali_feature_flags.clustering.kiali_urls[].cluster_name`
- `kiali_feature_flags.clustering.kiali_urls[].instance_name`
- `kiali_feature_flags.clustering.kiali_urls[].namespace`
- `kiali_feature_flags.clustering.kiali_urls[].url`

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -121,18 +155,21 @@ spec:
                         type: string
                       authentication_timeout:
                         type: integer
+                        default: 300
```
nrfox, line 158: Are we going to need to keep defaults in sync in four places by adding them here? Inside kiali server, kiali-operator, helm values, CRD? This wouldn't apply to non-kiali config on the CRD like the `deployment` options or `version`.

Makefile
```diff
@@ -121,6 +121,75 @@ validate-cr:
...
+	@echo "Synchronizing Kiali CRD from golden copy: crd-docs/crd/kiali.io_kialis.yaml"
+
+	@if [ -d "../helm-charts/kiali-operator/crds" ]; then \
+		echo "  -> helm-charts/kiali-operator/crds/crds.yaml (with YAML document separators)"; \
+		echo "---" > ../helm-charts/kiali-operator/crds/crds.yaml; \
+		cat crd-docs/crd/kiali.io_kialis.yaml >> ../helm-charts/kiali-operator/crds/crds.yaml; \
+		echo "..." >> ../helm-charts/kiali-operator/crds/crds.yaml; \
+	else \
+		echo "  -> helm-charts/kiali-operator/crds/crds.yaml (SKIPPED - directory not found)"; \
+	fi
```
nrfox, line 135: Can you make `helm-charts` dir configurable? Mine is at `kiali-helm-charts` so this command fails for me. I wonder if it'd be easier long term to move the operator and helm-charts into `kiali/kiali`?

nrfox (commented):
LGTM just some trouble with running the `sync-crds` target

Makefile
nrfox: If we don't want to make them configurable that's fine.

In practice, we always release the operator, server, and helm charts together right? The releases are all tied together and they share a lot of the same makefiles/hack scripts. Does it matter that the operator can also install older versions of the server?

crd-docs/crd/kiali.io_kialis.yaml
jmazzitelli: Ah... NOW I understand what you were asking. I did not understand. But I see you asked `keep **defaults** in sync` -- so I get it now.

Yes, those three you point out are the defaults that should be kept in sync. Those aren't the CRDs so that "make sync-crd" doesn't touch them.

We could (especially if AI is smart enough ;)) enhance the "make sync-crds" target to actually update these defaults, too, and in the `validate-crds` target confirm the defaults match the schema.

Here's an explanation of what these files are, for background:

> Don't we update: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml?

Those are the Ansible defaults as defined in the operator. Those get placed into the Kiali ConfigMap when Kiali is installed or updated during reconciliation. This yaml file does get manually updated when people add support for new settings, or change the defaults for whatever reason; this is so the defaults get set for those Kiali Server installations _installed by the operator_.

> and these: https://github.com/kiali/kiali/blob/3bc6f3bf0b02866f4c2e830f1e57d6b6078d26b4/config/config.go#L718-L987

The config/config.go defaults was always a sore point to me, but it was a necessary evil because those should only be used by the unit tests (back when we only had the operator as our true installation mechanism - more on the server helm chart below). Running the unit tests, there is no Kiali ConfigMap so the unit tests grab their base config from these config.go defaults. So we were forced to add defaults here in config.go. But it was supposed to be that these just tracked the defaults in the Ansible defaults yaml, because its the operator that configures all the settings in Kiali.

Those config.go defaults are actually overridden by settings that are set in the ConfigMap (it's how users configure Kiali - settings in the Kiali CR go into the ConfigMap which then get loaded into the server via config.go). This is why, since the operator came into existence, I've always tried to set every setting in the Ansible defaults of the operator - because that was supposed to be the source-of-truth of the actual defaults, not config/config.go. Otherwise, it was going to be confusing while debugging ("what is the default value? - is it defined in config.go or the operator?"). It was supposed to be, "go to the operator defaults, THOSE are the source-of-truth"). But, over time, the code base has drifted from this ideal and now some defaults are in config.go but not in the Ansible defaults. But THEN this wrinkle came into being:

> and these: https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml

Then the server helm chart was born (this occurred well after the operator had been published). Well, now there may be no operator when the server is installed. But unlike the operator, the values.yaml you see in the server helm charts are those that only the server helm chart actually needs to reference to do its job. That's what this comment is meant to convey:

https://github.com/kiali/helm-charts/blob/v2.13.0/kiali-server/values.yaml#L22-L25

```
#
# Settings that mimic the Kiali CR which are placed in the ConfigMap.
# Note that only those values used by the Helm Chart will be here.
#
```

Historically, the server helm chart was only meant as an unsupported convenience to users. We did not want to support it - the operator was supposed to be our only supported installation mechanism. So we didn't work diligently to get the values.yaml to exactly match all the defaults in the Ansible operator defaults. We only put those settings that the server helm chart needed, and that was it. But over time we have been forced to be more official with our support of the helm chart because so many people started using it.

So, in short, a lot of this is due to years-old legacy processes. But yes, we do have to make sure we keep the defaults (as declared in the CRD Schema) to match the actual code. AFAIK, all defaults in the schema DO match the code today.

jmazzitelli:
After a bunch of research re-remembering how CRD Schema versions work (and after confirming with Cursor :)), the best way forward to support v1.73 but also have a nice CRD Schema moving forward is to include all settings that v1.73 has in the CRD Schema, even if the settings are now removed ... this includes settings like spec.istio_namespace and spec.deployment.accessible_namespaces.

jmazzitelli:
all molecule tests pass on crc/openshift with latest code

jshaughn (commented):
The schema seems correct at this point and so I've started going through the test instructions [here](https://github.com/kiali/kiali/issues/8237#issuecomment-3029115905). 

Phase I passed with the exception of dex-related molecule tests.  These fails are tangential to the PR and likely due to some issue on my machine. Not blocking, just mentioning.

Phase II in progress...

PR #933, opened 2025-07-02
Add operator image to relatedImages section in CSV

ferhoyos:
This PR adds the Kiali operator image to the relatedImages section of the ClusterServiceVersion (CSV) manifest, ensuring proper image tracking and compliance with OpenShift/OLM requirements.

### DIFF
--- a/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
@@ -224,6 +224,8 @@ spec:
         - 'urn:alm:descriptor:com.tectonic.ui:text'
   apiservicedefinitions: {}
   relatedImages:
+  - name: operator
+    image: "${KIALI_OPERATOR}"
   - name: kiali_default
     image: "${KIALI_2_11}"
   - name: kiali_v2_11
@@ -242,8 +244,6 @@ spec:
     image: "${OSSMCONSOLE_2_4}"
   - name: ossmconsole_v1_73
     image: "${OSSMCONSOLE_1_73}"
-  - name: operator
-    image: "${KIALI_OPERATOR}"
   install:
     strategy: deployment
     spec:


manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
```diff
@@ -242,6 +242,8 @@ spec:
     image: "${OSSMCONSOLE_2_4}"
   - name: ossmconsole_v1_73
     image: "${OSSMCONSOLE_1_73}"
+  - name: operator
+    image: "${KIALI_OPERATOR}"
```
FilipB, line 246: There are two variables for the operator now, KIALI_OPERATOR_REGISTRY and KIALI_OPERATOR

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
```diff
@@ -271,7 +273,7 @@ spec:
               serviceAccountName: kiali-operator
               containers:
               - name: operator
-                image: ${KIALI_OPERATOR_REGISTRY}
```
FilipB, line 274: KIALI_OPERATOR_REGISTRY is also used in https://github.com/kiali/kiali-operator/blob/e2b3dcc6c99e89bbf165dae62bf24aff7334737a/Makefile#L85

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
jmazzitelli: Can I ask for something that probably looks trivial, but would make things a little easier in the future.

Move this new section (the operator image) all the way to the top (the first relatedImages, line 227). This is because this new section will never change in the future. But the others will eventually be rotated out. So when we check diffs of the changes as new versions come and old versions go, the position of this operator section will always be the same, never change.

This is a very trivial nit - but when making changes in this file over time, I like to see things in fixed positions when they can be (it just makes looking the diffs easier, from my experience).

jmazzitelli (changes requested):
I'm asking for a minor change

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
ferhoyos: Sure @jmazzitelli, the request makes sense :slightly_smiling_face:

jmazzitelli (approved):

PR #938, opened 2025-07-29
Fix for the update status task due to changes in kubernetes 1.33.

jmazzitelli:
This moves away from operator_sdk collection. Note that kubernetes.core versions 4 and 6 both support Kubernetes 1.24 and up (with v6 supporting more current versions), thus we should not have any backward compat issues.

The code changes in this PR were generated by Cursor/claude-4-sonnet

fixes: https://github.com/kiali/kiali/issues/8593

NOTE: this may not be needed. See comments in the issue

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -43,9 +43,7 @@ collections:
 - name: community.general
   version: 9.0.0
 - name: kubernetes.core
-  version: 4.0.0
-- name: operator_sdk.util
-  version: 0.5.0
+  version: 6.0.0
 - name: ansible.posix
   version: 1.6.2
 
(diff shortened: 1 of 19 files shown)


jmazzitelli:
the latest code looks like its working. now running molecule tests to make sure nothing is broken.

jmazzitelli:
small testing on kubernetes 1.31 shows this working. I upgraded my local minikube to 1.36.0 to pick up kubernetes 1.33.1 and am now running full molecule test suite. Preliminary tests shows it working, but I'll see what the full molecule test suite does.

jmazzitelli:
NOTE THIS MIGHT NOT BE NEEDED.

The original bug report showed `"Service Unavailable", "status": 503` as the failure - this might be a cluster environment issue, not an API issue.

jmazzitelli:
FWIW, all molecule tests just passed.

jmazzitelli:
see issue for discussion and details... this PR is not needed

PR #952, opened 2025-08-29
release workflow to validate crd sync

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/8681

this also checks the OSSMC CRD - we weren't doing that before


jmazzitelli:
I tested this in my fork and it seems to work. We now validate the CRDs so the OLM metadata is in sync.

PR #961, opened 2025-09-25
Add explicit OIDC endpoints configuration for security-hardened envs

olsib:
part of: https://github.com/kiali/kiali/issues/8777

Following Kiali development checklist for configuration changes:

✅ Set default values in defaults/main.yml
✅ Add new fields to CRD schema with descriptions
✅ Update comprehensive CR example documentation
✅ Add example usage in OpenID test CR

New fields added:
- token_endpoint: URL of provider's token endpoint
- userinfo_endpoint: URL of provider's UserInfo endpoint
- jwks_uri: URL of provider's JWK Set document

These fields enable Kiali authentication in environments where IdP restricts access to /.well-known/openid-configuration endpoint for security reasons.

When both authorization_endpoint and token_endpoint are configured, Kiali will use explicit endpoints instead of auto-discovery, supporting enterprise security policies that block discovery endpoints.

### DIFF
--- a/manifests/kiali-ossm/manifests/kiali.crd.yaml
+++ b/manifests/kiali-ossm/manifests/kiali.crd.yaml
@@ -185,15 +185,15 @@ spec:
                           authorization_endpoint:
                             type: string
                             description: "The URL of the provider's authorization endpoint."
+                          jwks_uri:
+                            type: string
+                            description: "The URL of the provider's JWK Set document."
                           token_endpoint:
                             type: string
                             description: "The URL of the provider's token endpoint."
                           userinfo_endpoint:
                             type: string
                             description: "The URL of the provider's UserInfo endpoint."
-                          jwks_uri:
-                            type: string
-                            description: "The URL of the provider's JWK Set document."
                       username_claim:
                         type: string
                         default: "sub"
--- a/manifests/kiali-upstream/2.18.0/manifests/kiali.crd.yaml
+++ b/manifests/kiali-upstream/2.18.0/manifests/kiali.crd.yaml
@@ -185,15 +185,15 @@ spec:
                           authorization_endpoint:
                             type: string
                             description: "The URL of the provider's authorization endpoint."
+                          jwks_uri:
+                            type: string
+                            description: "The URL of the provider's JWK Set document."
                           token_endpoint:
                             type: string
                             description: "The URL of the provider's token endpoint."
                           userinfo_endpoint:
                             type: string
                             description: "The URL of the provider's UserInfo endpoint."
-                          jwks_uri:
-                            type: string
-                            description: "The URL of the provider's JWK Set document."
                       username_claim:
                         type: string
                         default: "sub"


crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -173,10 +173,19 @@ spec:
                         default: false
                       issuer_uri:
                         type: string
+                      jwks_uri:
```
jmazzitelli, line 176: I'm thinking if we do this, we might want to combine these under a parent config. Since these all relate to overriding the discoverable `.well-known/openid-configuration` I'm thinking put it under something like:

```yaml
well-known-openid-config:
  jwks-url: ...
  token_endpoint: ...
```

Ask Cursor what it thinks is a good name of this parent section :) I think something like `well-known-openid-config` or `well-known-openid-config-override` or something.

jmazzitelli:
@olsib we need to take a step back. You just removed a setting from the CRD. The Kiali team has recently made a decision that we will not be removing settings from the CRD for several reasons (backward compat; we don't want to manage new CRD versions, etc.). So the latest change needs to be reverted.

I merely asked "what's the difference between the two?" If they are the same setting, I was going to say just leave the original as-is (or add the second new one, and deprecate the old original one and only use the original as a backup if the new one isn't set).

Let's have a discussion about what the requirements are, how we expect this to be used, and how it is going to be backwards compatible. I realize generative AI can help change code in a flash, but lots of times it isn't going to do the right thing.

Right now, here's how I think this is to go:

1. Add new settings the CRD Schema with the overrides. If set, then I assume this means Kiali is not going to use the oidc-configuration for discovery - I believe this is the entire use-case for this change.
2. We need the operator to define the new CRD Schema with the new settings with defaults. The defaults need to be set in such a way so the oidc-configuration URL is used for discovery by default in the server (I think this means all settings need to be defaulted to things like empty string). The server should use the oidc-configuration URL for discovery if the new settings are empty string.
3. I do not think there needs to be any changes to the helm chart - I haven't checked, but I don't think it needs any oidc configuration changes. BUT! you will need to create a helm-chart PR with the CRD synced (any changes to the CRD now requires a helm-chart PR with minimum the CRDs synced - see the "make sync-crds" target in the operator).
4. Backward compatibility requires that any existing settings remain as-is. You can deprecate them (meaning, just have a deprecation notice in the CRD Schema description) but they cannot be deleted. Typically, we would want to use the deprecated settings in server code - people will still have them set in production and we don't want to break them. So if the old setting is set, and the new setting is not, the old setting takes effect.

I should probably update the dev checklist to include the new CRD sync process.

I say all of this with the caveat I know very little about OIDC and the OpenID integration in Kiali. Edgar (@israel-hdez) is no longer on the Kiali project so he won't be a person we can tap for help.

jmazzitelli:
please rebase this on master - there are come CI workflows that we want to pick up and run in new PRs

jmazzitelli:
While the new stuff here doesn't have molecule testing, I can see the existing auth-related molecule tests don't break. So this is good.

Running this locally:

```
hack/k8s-minikube.sh --hydra-enabled true -mp ci start && \
hack/k8s-minikube.sh -mp ci istio && \
make CLUSTER_TYPE=minikube MINIKUBE_PROFILE=ci clean build-ui build test cluster-push && \
hack/run-molecule-tests.sh --client-exe "$(which kubectl)" --cluster-type minikube --minikube-profile ci -udi true -hcrp false -at "default openid-test header-auth-test token-test"
```

results in:

```
=====================
=== TEST RESULTS: ===
=====================

                             openid-test... success [3m 32s]
                        header-auth-test... success [2m 31s]
```

_(Note: we are unable to run these two auth-related tests in CI today so they must be run manually)._

jshaughn:
@olsib Still interesting in this effort?

PR #964, opened 2025-09-29
K8s GW API Inference Extensions to v1

hhovsepy:
Core PR https://github.com/kiali/kiali/pull/8783
Issue https://github.com/kiali/kiali/issues/8782

### DIFF
--- a/manifests/kiali-upstream/2.17.0/manifests/kiali.v2.17.0.clusterserviceversion.yaml
+++ b/manifests/kiali-upstream/2.17.0/manifests/kiali.v2.17.0.clusterserviceversion.yaml
@@ -521,7 +521,7 @@ spec:
           - extensions.istio.io
           - telemetry.istio.io
           - gateway.networking.k8s.io
-          - inference.networking.x-k8s.io
+          - inference.networking.k8s.io
           resources: ["*"]
           verbs:
           - get


hack/verify-kiali-server-permissions.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+!/bin/bash
```
jmazzitelli, line 1: Is this right? I've never see this without the comment # prefix

jmazzitelli (changes requested):
Did this x-k8s.io permission go out in a downstream production release?

manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml
```diff
@@ -645,7 +645,7 @@ spec:
           - extensions.istio.io
           - telemetry.istio.io
           - gateway.networking.k8s.io
-          - inference.networking.x-k8s.io
+          - inference.networking.k8s.io
```
jmazzitelli, line 648: If x-k8s.io is not supported in any production release downstream, this is OK. If we had this in any 2.x release prior, we have to keep the old one and add the new one.

hack/verify-kiali-server-permissions.sh
hhovsepy: seems a typo

jmazzitelli (approved):

PR #966, opened 2025-09-30
Describe the `custom_workload_types` feature

martinbaillie:
Reflects changes from https://github.com/kiali/kiali/pull/8780


jmazzitelli:
yes, we should put something in config-values-test molecule test just for completeness. Just add it  to the test Kiali CR so we can see it get processed properly.

martinbaillie:
wrt. placement, i did have a bit of a think before implementing this and I couldn't readily find a better place than your feature flags section. 

I had started in the `kubernetes_config` stanza but realized it was not a great home as it is more for kube client tuning and what to exclude from server-side queries whereas this feature is purely UI.

PR #968, opened 2025-10-01
[ci] run some molecule tests on operator PRs

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8785

When changing a molecule test or any ansible code in the operator's default role, a small set of molecule tests are run just to make sure things are OK. By default "config-values-test" will always run. If the PR changes any molecule tests, those molecule tests are run too. The exceptions are the molecule tests that need OIDC (the molecule tests that need OIDC cannot run in CI today) - this includues openid-test, header-auth-test, token-test. OpenShift molecule tests are also excluded as they, too, cannot run in CI today

### DIFF
--- a/.github/workflows/molecule-tests.yml
+++ b/.github/workflows/molecule-tests.yml
@@ -23,22 +23,22 @@ jobs:
       id: determine-tests
       run: |
         cd kiali-operator
-        
+
         # Always run config-values-test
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
This is a good idea. Just to make sure we don't footgun ourselves.

PR #970, opened 2025-10-02
enable hydra so molecule tests that need oidc can pass

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/8789


jshaughn (approved):
This looks OK to me, approving, but see my comment in the server PR.

PR #1025, opened 2026-04-16
Add kiali_feature_flags.authz.require_namespace_get config option

jshaughn:
Part-of: https://github.com/kiali/kiali/issues/7125

## Summary
- Adds a new `kiali_feature_flags.authz.require_namespace_get` configuration option
- When enabled, Kiali requires GET permission on each namespace for it to be visible, even if the user has LIST permission
- Supports multi-tenant environments where LIST is granted broadly but GET is restricted per namespace
- Defaults to `false` to preserve backward compatibility
- Updates CRD golden copy, synced CRD copies, example CR, and operator defaults

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -481,7 +481,6 @@ spec:
 
   kiali_feature_flags:
     authz:
-      # default: require_namespace_get is false
       require_namespace_get: false
     clustering:
       enable_exec_provider: false
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -480,6 +480,9 @@ spec:
     version_label_name: ""
 
   kiali_feature_flags:
+    authz:
+      # default: require_namespace_get is false
+      require_namespace_get: false
     clustering:
       enable_exec_provider: false
     # default: custom_workload_types is an empty list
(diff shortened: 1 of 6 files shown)


crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -480,6 +480,9 @@ spec:
     version_label_name: ""
 
   kiali_feature_flags:
+    authz:
+      # default: require_namespace_get is false
```
jmazzitelli: no need for this comment - it is superflous. All values in this example CR *are* the defaults. You only comment what the default is if the example value shown is NOT the default.

```suggestion
```

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1730,6 +1730,17 @@ spec:
...
+                    description: "Authorization-related feature flags."
+                    type: object
+                    properties:
+                      require_namespace_get:
+                        description: |
+                          When true, a user must have GET permission on a namespace for it to be visible,
+                          even if the user has LIST permission. When false (the default), a successful LIST
+                          is trusted without per-namespace GET checks. Enable this in multi-tenant environments
+                          where LIST permission is granted broadly but GET is restricted per namespace.
+                        type: boolean
```
jmazzitelli, line 1743: we can add a default: here to enforce this at the CR-creation level

```suggestion
                        type: boolean
                        default: false
```

jmazzitelli (changes requested):

jmazzitelli (dismissed):

jmazzitelli (approved):

PR #1029, opened 2026-04-23
Add PodDisruptionBudget support to the Kiali operator

### LINKED ISSUE #9590: Add pod disruption budget template to kiali chart
Want Add pod disruption budget template to kiali chart
It helps to maintain service availability during the cluster upgrade/patching activities when deployed at kubernetes envs

### What is the current behaviour?
kiali doesn't have PDB templates, so if service is deployed in multi replica structure and cluster goes under patching/upgrade activities abrupt disruption of kiali pods happens and it results in service disruption

### What is the new behaviour?
Adding PDB would ensure **minavailable** pods are always running irrespective of any patching or upgrade activity where kiali is deployed in multi-replica

jmazzitelli:
Add opt-in PodDisruptionBudget (PDB) support to the Kiali operator, mirroring the existing HorizontalPodAutoscaler pattern. Users can configure a PDB via the new `deployment.pod_disruption_budget` CR field, which accepts an `api_version` (defaults to "policy/v1" when left empty) and a `spec` passthrough. When `spec` is empty no PDB is created; when populated the operator creates, updates, and removes the PDB resource as part of the normal deploy and remove lifecycle.

Fixes: https://github.com/kiali/kiali/issues/9590

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -234,6 +234,11 @@ spec:
     # default: pod_annotations is empty
     pod_annotations:
       proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'
+    pod_disruption_budget:
+      api_version: ""
+      # default: spec is empty
+      spec:
+        minAvailable: 1
     # default: pod_labels is empty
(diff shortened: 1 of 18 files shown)


jshaughn (approved):
This looks good to me, and AI review comments are nominal.

PR #1031, opened 2026-04-30
Add ability to enable/disable Prometheus via external_services.prometheus.enabled

jmazzitelli:
Introduce a new `external_services.prometheus.enabled` configuration field (default true) that allows Kiali to run without a Prometheus metrics store. When disabled, a no-op Prometheus client is injected so the server starts without a Prometheus connection. Metrics-dependent features degrade gracefully: the graph page and mini-graph cards show an empty state with an explanation, health badges fall back to Kubernetes-only replica status, metrics and traffic tabs are hidden from detail pages, mesh target panels display a disabled message, and MCP/AI metric tools return an informative error. Non-metrics features (workload/service/app listing, Istio config, mesh topology, logs, traces) continue to work normally.

Changes span the Kiali server (config, noop client, handler guards, mesh generator, addon status, version probe), operator (CRD schema, defaults, conditional URL defaulting and secret mounts), Helm charts (credential gating, values, template tests), frontend (config type, empty states, tab hiding, test data), molecule tests (baseline and disabled assertions in metrics-test and config-values-test), and kiali.io documentation.

Part of: https://github.com/kiali/kiali/issues/8654

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -404,6 +404,7 @@ spec:
       # default: custom_headers is empty
       custom_headers:
         customHeader1: "customHeader1Value"
+      enabled: true
       health_check_url: ""
       is_core: true
       # default: query_scope is empty
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -1482,6 +1482,10 @@ spec:
                         type: object
                         additionalProperties:
                           type: string
+                      enabled:
+                        description: "When true, Kiali will connect to the Prometheus instance for Istio telemetry metrics. When false, Kiali will run without metrics support - the graph, health based on request rates, and metrics displays will be unavailable."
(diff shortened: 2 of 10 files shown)


jshaughn (approved):
LG!

jmazzitelli:
Here's what the molecule tests confirm:

**config-values-test:**
- `prometheus.enabled` defaults to `true` in the operator-generated ConfigMap

**metrics-test (baseline — Prometheus enabled):**
- `/api/config` reports `prometheus.enabled: true`
- Graph endpoint returns HTTP 200
- Prometheus reported as healthy in `/api/istio/status`
- Prometheus infra node present in `/api/mesh/graph`

**metrics-test (after disabling via CR update):**
- `/api/config` reports `prometheus.enabled: false`
- Graph endpoint returns HTTP 200 with empty graph (graceful degradation, no 503)
- Non-metrics endpoint (`/api/namespaces`) still returns HTTP 200 (non-metrics features work)
- Prometheus not reported as unhealthy in `/api/istio/status` (it's disabled, not failing)
- Prometheus infra node absent from `/api/mesh/graph`

PR #1039, opened 2026-05-27
feat: add OAuth2 client_credentials support to operator CRD and secret scanning

emanuelbesliu:
## Summary

Adds operator support for OAuth2 `client_credentials` authentication on all external service auth blocks.

### CRD Schema (`roles/default/kiali-deploy/defaults/main.yml`)
- Added `oauth2` sub-object under all 5 auth blocks (prometheus, grafana, tracing, perses, custom_dashboards.prometheus)
- Fields: `client_id`, `client_secret`, `token_url`, `scopes`, `audience`, `auth_style`

### Secret Volume Scanning (`roles/default/kiali-deploy/tasks/main.yml`)
- Added OAuth2 `client_id` and `client_secret` secret extraction for all 5 services
- Uses the same `secret:<name>:<key>` pattern as existing credentials

### CR Example (`roles/default/kiali-deploy/kiali-deploy-example-cr.yaml`)
- Added commented `oauth2` section in Prometheus auth block showing usage

## Related PRs
- Kiali core: https://github.com/kiali/kiali/pull/9751
- Helm Charts: emanuelbesliu/helm-charts#feat/oauth2-client-credentials-auth

Assisted-by: Claude (Anthropic)


crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -298,8 +298,15 @@ spec:
           insecure_skip_verify: false
           password: ""
           token: ""
-          type: "none"
+          type: "none" # Valid values: none, basic, bearer, oauth2
           use_kiali_token: false
+          oauth2:
```
jmazzitelli, line 300: `oauth2` is not in alphabetical order - must appear prior to `password` and after `insecure_skip_verify`.

This is mentioned in a few places in our coding standards/agents docs:
* https://github.com/kiali/kiali/blob/master/AGENTS.md?plain=1#L84-L85
* https://github.com/kiali/kiali/blob/master/AGENTS.md?plain=1#L1601
* https://github.com/kiali/kiali/blob/master/AGENTS.md?plain=1#L1998
* https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc?plain=1#L210

roles/default/kiali-deploy/defaults/main.yml
```diff
@@ -160,6 +160,13 @@ kiali_defaults:
           token: ""
           type: "none"
           use_kiali_token: false
+          oauth2:
```
jmazzitelli, line 160: same critique about alphabetizing

jmazzitelli (changes requested):
quick minor thing needs fixing here

jmazzitelli:
I'm not sure why CI failed here - might be a problem with the CI script that runs - I'll look into that. This does not seem valid for this PR - not sure why it thinks these fields were deleted:
```
Breaking changes detected:
  ✗ Field removed: external_services.custom_dashboards.prometheus.auth.key_file
  ✗ Field removed: external_services.grafana.auth.key_file
  ✗ Field removed: external_services.perses.auth.key_file
  ✗ Field removed: external_services.prometheus.auth.key_file
  ✗ Field removed: external_services.tracing.auth.key_file
```

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
emanuelbesliu: Fixed in 8813495 — moved `oauth2` to appear after `insecure_skip_verify` and before `password` in both the CR example and defaults file.

roles/default/kiali-deploy/defaults/main.yml
emanuelbesliu: Fixed in the same commit (8813495).

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1064,6 +1064,30 @@ spec:
...
+                                    type: string
+                                  scopes:
+                                    description: "A list of scopes to request with the token."
+                                    type: array
+                                    items:
+                                      type: string
+                                  token_url:
+                                    description: "The URL of the OAuth2 token endpoint."
+                                    type: string
                                 type: string
```
jmazzitelli: This is why CI is failing - you injected the oauth2 field BETWEEN key_file.description and key_file.type - thus this CRD is actually invalid. So the CI test failure was correctly detecting a problem that needs fixing.

jmazzitelli (changes requested):

crd-docs/crd/kiali.io_kialis.yaml
jmazzitelli: Add `type: string` in the correct location (back to where it was) to all 5 key_file entries in crd-docs/crd/kiali.io_kialis.yaml, then re-run `make sync-crds` to propagate the fix to the manifests and helm chart.

jmazzitelli (changes requested):
Thanks for the operator changes — I quickly compared against [the plan I posted in the discussion](https://github.com/kiali/kiali/discussions/9735#discussioncomment-17065061) and found several gaps.

## `client_id` should not use `secret:` notation

The secret volume scanning in `tasks/main.yml` scans `oauth2.client_id` for the `secret:<name>:<key>` pattern across all five services, in addition to `oauth2.client_secret`. This is incorrect. `client_id` is a public identifier in the OAuth2 spec — it has no confidentiality requirement and does not belong in a Kubernetes Secret. No other Kiali auth config treats `client_id` as a secret (e.g. the existing OpenID `client_id` is a plain string with no secret mounting support). The `client_id` scanning blocks should be removed; only `client_secret` warrants this treatment.

## CRD schema: `auth_style` missing enum constraint and default

For all five auth blocks, `oauth2.auth_style` is defined as a bare string with no `enum` constraint and no `default` value. The plan specifies:

```yaml
auth_style:
  enum: ["autodetect", "header", "params"]
  default: "autodetect"
```

Without the enum, invalid values are silently accepted by the CRD. Without the default, users who omit the field get no default behavior documented at the schema level.

## CRD schema: tracing gRPC limitation not documented

The description for `external_services.tracing.auth.type` does not mention that `oauth2` is only supported when `use_grpc` is `false`. This should be noted in the field description so users see it when inspecting the CRD.

## Ansible validation: `use_kiali_token` conflict check incomplete

The `fail` tasks validating that `oauth2` and `use_kiali_token` cannot be used together only cover prometheus, tracing, and grafana. **Perses and custom_dashboards.prometheus are missing from this check.** A user who sets `use_kiali_token: true` with `auth.type: oauth2` on either of those services will not get an operator-level error — the conflict passes through silently.

jmazzitelli (changes requested):
Before posting your next update, please verify your own work against the plan using an AI tool with this prompt — this will tell you if you missed anything and is basically what I've been doing. It's best if you do this prior to requesting review:

> Read the implementation plan at https://github.com/kiali/kiali/discussions/9735#discussioncomment-17065061 in full. Then audit the code in this repo against every requirement in that plan. Report only what is wrong, missing, or incomplete — not what is correct. Be specific: file paths, line numbers, and what exactly is missing or broken. The full implementation spans three repos: kiali, kiali-operator, and helm-charts. Check all three.

Once the code fully implements the plan, I'll do a deeper review for correctness and security — but I can't do that until the plan is completely implemented.

---

I quickly compared the latest changes against [the plan I posted in the discussion](https://github.com/kiali/kiali/discussions/9735#discussioncomment-17065061). Two items are still incomplete.

## CRD schema: `auth_style` still missing enum constraint and default

For all five auth blocks, `oauth2.auth_style` remains a bare `type: string` with no `enum` constraint and no `default` value. The plan specifies:

```yaml
enum: ["autodetect", "header", "params"]
default: "autodetect"
```

Affected lines: 1076 (custom_dashboards), 1186 (grafana), 1423 (perses), 1538 (prometheus), 1652 (tracing).

## CRD schema: tracing gRPC note still missing

The description for `external_services.tracing.auth.type` (line 1675) now mentions `oauth2` as a valid option but still does not note that it requires `use_grpc: false`. Users inspecting the CRD will not see this constraint documented on the field.

crd-docs/crd/kiali.io_kialis.yaml
```diff
@@ -1065,14 +1065,40 @@ spec:
...
+                                  audience:
+                                    description: "The audience parameter sent with the token request. Some providers require this to scope the token."
+                                    type: string
+                                  auth_style:
+                                    description: "How to send client credentials to the token endpoint. Use `params` to send in the request body, `header` to send as HTTP Basic Auth, or `autodetect` to let the library determine the method."
+                                    type: string
+                                    default: "autodetect"
+                                    enum: ["autodetect", "header", "params"]
+                                  client_id:
+                                    description: "The OAuth2 client ID. May refer to a secret using the pattern `secret:<secretName>:<secretKey>`."
```
jmazzitelli: **BUG-1: `client_id` description claims `secret:` support that doesn't exist**

The `client_id` field description reads: _"May refer to a secret using the pattern `secret:<secretName>:<secretKey>`."_ However, the operator only performs volume scanning for `client_secret` — there is no Jinja2 block in `tasks/main.yml` that mounts a secret for `client_id`. If an operator follows the documentation and sets `client_id: secret:mysecret:mykey`, that string will be written verbatim into the Kiali ConfigMap and used as the literal client ID sent to the identity provider, causing OAuth2 token requests to fail at runtime with no operator-level error.

More importantly, this documentation is wrong by design: RFC 6749 §2.2 (Client Identifier) explicitly states that `client_id` _"is not a secret; it is exposed to the resource owner and MUST NOT be used alone for client authentication."_ The `client_id` is a public identifier and must not be treated as a credential. It does not belong in a Kubernetes Secret and should not support the `secret:` pattern.

Remove the "May refer to a secret" sentence from the `client_id` description in all five auth sections (lines 1081, 1193, 1432, 1549, 1665). After editing `crd-docs/crd/kiali.io_kialis.yaml`, run `make sync-crds` to propagate the change to the generated copies.

**Severity: Important**

roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -728,6 +839,26 @@
       {% set d = d | combine({'perses-key': {'secret_name': kiali_vars.external_services.perses.auth.key_file | regex_replace('secret:(.+):.+', '\\1'), 'secret_key': kiali_vars.external_services.perses.auth.key_file | regex_replace('secret:.+:(.+)', '\\1'), 'path': kiali_vars.external_services.perses.auth.key_file | regex_replace('secret:.+:(.+)', '\\1') }}) %}
       {% endif %}
 
+      {# Prepare the secret volume for prometheus oauth2 client_secret #}
+      {% if kiali_vars is defined and kiali_vars.external_services is defined and kiali_vars.external_services.prometheus is defined and kiali_vars.external_services.prometheus.enabled is defined and kiali_vars.external_services.prometheus.enabled|bool == True and kiali_vars.external_services.prometheus.auth is defined and kiali_vars.external_services.prometheus.auth.oauth2 is defined and (kiali_vars.external_services.prometheus.auth.oauth2.client_secret | default('')) | regex_search('secret:.+:.+') %}
```
jmazzitelli: **BUG-2: Secret volumes mounted regardless of `auth.type`**

The Jinja2 guards for `oauth2.client_secret` volume scanning do not check `auth.type == 'oauth2'` before mounting the secret. The guard only checks that the `oauth2` sub-object is defined and that `client_secret` matches the `secret:` pattern. If a user sets `auth.type: basic` but also has `oauth2.client_secret: secret:x:y` (e.g., pre-staged or leftover config), the operator will mount the secret volume and expose it to the Kiali pod unnecessarily.

Fix — add `auth.type == 'oauth2'` guard to each of the five oauth2 volume scanning blocks (lines 843, 848, 853, 858, 892). For example for prometheus:

```jinja2
{% if ... and kiali_vars.external_services.prometheus.auth.type | default('') == 'oauth2' and kiali_vars.external_services.prometheus.auth.oauth2 is defined and (kiali_vars.external_services.prometheus.auth.oauth2.client_secret | default('')) | regex_search('secret:.+:.+') %}
```

**Severity: Important**

roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -508,6 +508,117 @@
   - can_i_create_clusterrolebindings.result.status.allowed is defined
   - can_i_create_clusterrolebindings.result.status.allowed == False
 
+- name: Validate oauth2 auth config for prometheus
```
jmazzitelli, line 511: **TST-1: No molecule test coverage for any OAuth2 configuration**

This PR modifies `roles/default/kiali-deploy/defaults/main.yml` and `roles/default/kiali-deploy/tasks/main.yml` — both of which trigger `config-values-test` in CI on every PR. Per the testing practices, `config-values-test` must be updated to cover new configuration settings. Zero changes were made to `molecule/config-values-test/`. The `kiali-cr-all-yaml` file contains no `auth.type: oauth2` or `oauth2:` sub-block for any external service.

At minimum, add prometheus and tracing to `kiali-cr-all-yaml` with OAuth2 config. Use plain-text dummy values — the token URL is never contacted during a molecule test since OAuth2 token fetch is lazy:

```yaml
prometheus:
  auth:
    type: oauth2
    oauth2:
      audience: ""
      auth_style: "header"
      client_id: "test-client-id"
      client_secret: "test-client-secret"
      scopes: []
      token_url: "https://login.example.com/token"

tracing:
  use_grpc: false
  auth:
    type: oauth2
    oauth2:
      audience: ""
      auth_style: "header"
      client_id: "test-client-id"
      client_secret: "test-client-secret"
      scopes: []
      token_url: "https://login.example.com/token"
```

This should be sufficient for the test to pass — OAuth2 token fetch is lazy and the URL is never contacted at startup. If for any reason the Kiali server fails to start with a non-reachable OAuth2 URL, this specific test case can be skipped with a comment explaining why.

**Severity: Critical**

roles/default/kiali-deploy/tasks/main.yml
jmazzitelli: **TST-2: OAuth2 `client_secret` volume mounting not tested**

The `external-auth-rotation-test` scenario validates `secret:name:key` volume mounting for Grafana `username`, `password`, `cert_file`, and `key_file`. This PR adds the identical mechanism for `oauth2.client_secret` across all five services but none of the new volume entries are tested with a `secret:` reference. The Jinja2 regex logic for these entries is non-trivial and untested — a regression in a condition guard would go undetected.

Extend the existing `external-auth-rotation-test` scenario to set `client_secret: "secret:prometheus-oauth2-secret:client_secret"` for prometheus (following the existing pattern in that scenario) and assert that volume `prometheus-oauth2-client-secret` appears in the pod spec.

**Severity: Important**

jmazzitelli (changes requested):

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -296,9 +296,16 @@ spec:
       prometheus:
         auth:
           insecure_skip_verify: false
+          oauth2:
+            audience: ""
+            auth_style: "autodetect"
```
jmazzitelli:
```suggestion
            auth_style: "header"
```

This is what the CI error is flagging - this CR example file now has the incorrect default. This is no longer a valid value - set this to "header" which is the new default. Do this change in the 5 new sections.


```
The TestKiali "kiali" is invalid: 
* spec.external_services.custom_dashboards.prometheus.auth.oauth2.auth_style: Unsupported value: "autodetect": supported values: "header", "params"
* spec.external_services.grafana.auth.oauth2.auth_style: Unsupported value: "autodetect": supported values: "header", "params"
* spec.external_services.perses.auth.oauth2.auth_style: Unsupported value: "autodetect": supported values: "header", "params"
* spec.external_services.prometheus.auth.oauth2.auth_style: Unsupported value: "autodetect": supported values: "header", "params"
* spec.external_services.tracing.auth.oauth2.auth_style: Unsupported value: "autodetect": supported values: "header", "params"
----------
ERROR! Validation failed for Kiali CR [/home/runner/work/kiali-operator/kiali-operator/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml]
make: *** [Makefile:109: validate-cr] Error 1
```

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -363,8 +377,15 @@ spec:
     perses:
       auth:
         insecure_skip_verify: false
+        oauth2:
+          audience: ""
+          auth_style: "autodetect"
```
jmazzitelli:
```suggestion
          auth_style: "header"
```

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -394,9 +415,16 @@ spec:
     prometheus:
       auth:
         insecure_skip_verify: false
+        oauth2:
+          audience: ""
+          auth_style: "autodetect"
```
jmazzitelli:
```suggestion
          auth_style: "header"
```

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -420,9 +448,16 @@ spec:
     tracing:
       auth:
         insecure_skip_verify: false
+        oauth2:
+          audience: ""
+          auth_style: "autodetect"
```
jmazzitelli:
```suggestion
          auth_style: "header"
```

crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
```diff
@@ -321,9 +328,16 @@ spec:
     grafana:
       auth:
         insecure_skip_verify: false
+        oauth2:
+          audience: ""
+          auth_style: "autodetect"
```
jmazzitelli:
```suggestion
          auth_style: "header"
```

jmazzitelli (dismissed):
fix CR example file - should show the correct defaults

jmazzitelli (approved):

PR #1040, opened 2026-05-28
update workflows

jmazzitelli:
This PR hardens the GitHub Actions workflows against several known vulnerability classes.

**Supply chain risk (unpinned actions):** All third-party actions were using floating version tags (e.g. `@v4`, `@v5`). These are mutable references — a compromised action maintainer can silently retag them to malicious code that executes in CI with access to repository secrets. All action references have been pinned to their full immutable commit SHA, with the version tag retained as a comment.

**Unverified binary download (yq):** Multiple workflows downloaded the `yq` binary from GitHub releases and executed it without verifying a checksum. If the `mikefarah/yq` release infrastructure were compromised, a malicious binary would be executed with `sudo` privileges. Checksum verification has been added using the SHA-256 value from the official `checksums` file published alongside each release.

**Missing `permissions:` blocks:** None of the workflows declared explicit token permissions, meaning they inherited the repository default (typically broad write access). All workflows now declare `permissions: read-all` at the top level, with `contents: write` and/or `pull-requests: write` granted only to the specific jobs that actually need them (release, tag creation, version bumping).

Generated-by: Claude

### DIFF
--- a/.github/workflows/bump-release-version.yml
+++ b/.github/workflows/bump-release-version.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout branch
-      uses: actions/checkout@v5
+      uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5
       with:
         ref: ${{ github.event.inputs.tag_branch }}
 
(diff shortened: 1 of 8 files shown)


jshaughn (commented):
Looks fine. Wait for CI to complete. Not sure what dependabot will do here because I think typically that is what performs the updates. Add a description before merge.

jshaughn (dismissed):

jshaughn (approved):

PR #1049, opened 2026-06-03
Added new config option insecure_skip_verify for chat_ai

### LINKED ISSUE #9736: [AI] Make LightSpeed TLS verification configurable
The new LightSpeed provider currently creates its client with `WithInsecureSkipTLS(true)` unconditionally in `ai/providers/lightspeed/lightspeed_provider.go`.
That means TLS certificate verification is always disabled for LightSpeed HTTPS endpoints, even in normal production setups where the certificate should be validated.

### What is the new behavior?
TLS verification should be enabled by default.

If insecure TLS is needed, it should be controlled explicitly through configuration, for example:

a provider-specific insecure_skip_tls_verify option, or
reuse of an existing shared TLS/auth configuration pattern if there is one.

it would be more consistent to reuse Kiali's existing external-service auth/TLS model.

Kiali already supports outbound TLS settings for other integrations through config.Auth, including:

- insecure_skip_verify
- cert_file
- key_file

and builds the HTTP transport through the shared httputil.CreateTransport(...) helpers.

hhovsepy:
Core PR https://github.com/kiali/kiali/pull/9780
Fixes https://github.com/kiali/kiali/issues/9736

### DIFF
--- a/hack/verify-crd-defaults.sh
+++ b/hack/verify-crd-defaults.sh
@@ -62,6 +62,7 @@ PROPERTIES_TO_IGNORE=(
     "external_services.istio.istio_sidecar_annotation"
 
     # Array item defaults that don't have direct Ansible equivalents
+    "chat_ai.providers.items.insecure_skip_verify"
     "deployment.custom_secrets.items.optional"
     "extensions.items.enabled"
     "external_services.istio.component_status.components.items.is_core"
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -253,6 +253,10 @@ spec:
                         endpoint:
                           description: "The endpoint of the provider. This is only used if a model has no endpoint specified."
                           type: string
+                        insecure_skip_verify:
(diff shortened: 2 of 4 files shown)


hack/verify-crd-defaults.sh
```diff
@@ -62,6 +62,7 @@ PROPERTIES_TO_IGNORE=(
     "external_services.istio.istio_sidecar_annotation"
 
     # Array item defaults that don't have direct Ansible equivalents
+    "chat_ai.providers.items.insecure_skip_verify"
```
josunect, line 65: This is required because this field is no added into  `roles/default/kiali-deploy/defaults/main.yml` ?

I wonder if it would be more useful to add a providers entry in the docs as example?

hhovsepy: very valid point, keeping the providers: [] empty here as for other properties,
added a kiali.io docs https://github.com/kiali/kiali.io/pull/982

josunect: For what I know, I think this kiali file is generated using the operator, so probably should be added here?

josunect (approved):
LGTM

PR #1061, opened 2026-06-12
add option to add custom labels to the Service only

### LINKED ISSUE #9858: Add an option to add labels only to the Kiali service (i.e. via a new parameter service_labels)
The Kiali CR supports setting custom annotations on the Kiali Service via the `service_annotations` parameter. It would be useful to also have a parameter that allows you to add custom labels only to the Kiali Service and not any of the other resources. For example, a `service_labels` parameter.

Our use case is when we start using Solo's Istio distribution we need to label our services with custom solo labels for multi-cluster support.

### What is the current behavior?
It is currently possible to (ab)use the `.spec.deployment.extra_labels` parameter to do this, but it will label all resources and those labels do not make much sense there (although they do not cause any issues either).

### What is the new behavior?
With the extra parameter, custom labels can only be set on the Service resource.

rolandkool:
To fix https://github.com/kiali/kiali/issues/9858


roles/default/kiali-deploy/templates/kubernetes/service.yaml
```diff
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   name: {{ kiali_vars.deployment.instance_name }}
   namespace: "{{ kiali_vars.deployment.namespace }}"
-  labels: {{ kiali_resource_metadata_labels }}
+  labels: {{ kiali_service_metadata_labels }}
```
jmazzitelli: Correct me if I'm wrong, but this is basically overwrites all the resource labels with just the custom service labels (i.e. it is replacing, not adding). This means this will lose all the `kubernets.io/XXX` labels that all Kiali resources need to have - which is the wrong thing to do.

roles/default/kiali-deploy/templates/openshift/service.yaml
```diff
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   name: {{ kiali_vars.deployment.instance_name }}
   namespace: "{{ kiali_vars.deployment.namespace }}"
-  labels: {{ kiali_resource_metadata_labels }}
+  labels: {{ kiali_service_metadata_labels }}
```
jmazzitelli: Correct me if I'm wrong, but this is basically overwrites all the resource labels with just the custom service labels (i.e. it is replacing, not adding). This means this will lose all the `kubernets.io/XXX` labels that all Kiali resources need to have - which is the wrong thing to do.

jmazzitelli (changes requested):

roles/default/kiali-deploy/templates/kubernetes/service.yaml
rolandkool: I'm not very familiar with Ansible, so I could misunderstand how this works. I thought my change to main.yml would merge the service_labels with the kubernetes_resource_labels fact (which contain all the other labels).

jmazzitelli:
(I will be submitted a PR to this PR shortly to fix some bugs - stay tuned)

roles/default/kiali-deploy/templates/kubernetes/service.yaml
jmazzitelli: This is fixed in my followup PR

roles/default/kiali-deploy/templates/openshift/service.yaml
jmazzitelli: This is fixed in my followup PR

jmazzitelli (changes requested):
This PR requires the changes that I submitted here:

* https://github.com/rolandkool/kiali-operator/pull/1

rolandkool:
@jmazzitelli could you check if this is OK now?

jmazzitelli:
> could you check if this is OK now?

CRD has changed (we just did the 2.29 release yesterday) so you need to do a "make sync-crds" again. See [CI failure](https://github.com/kiali/kiali-operator/actions/runs/29314006364/job/87069038843?pr=1061).

jmazzitelli:
Something bad happened - this task was removed in my PR: https://github.com/rolandkool/kiali-operator/commit/04caff36059a417435ceddbbb8f94ccfcf8df2ea#diff-fdfdd3a1972d101a6ad24c7cebcc48ae8cb539d555d4413549c7e4cac23c99f5

But you added it back here: https://github.com/kiali/kiali-operator/pull/1061/changes/4a9b820796c0ccf48a3faec6787ee0e9ac91aeca

That has to be removed - this is dead code:

```
- name: Determine the extra custom labels to apply to the service resource
  set_fact:
    kiali_service_metadata_labels: "{{ kiali_vars.deployment.service_labels | combine(kiali_resource_metadata_labels) }}"
  when:
  - kiali_vars.deployment.service_labels is defined
  - kiali_vars.deployment.service_labels | length > 0
```

rolandkool:
Looks like I'm only creating a bigger mess... Maybe abandon and create a fresh one?

jmazzitelli:
> Looks like I'm only creating a bigger mess... Maybe abandon and create a fresh one?

Pull my branches from operator and helm here - if you reset your PR branches to these, it should be good:

* operator - https://github.com/jmazzitelli/kiali-operator/tree/rolandkool-new_param_service_labels
* helm - https://github.com/jmazzitelli/helm-charts/tree/rolandkool-new_param_service_labels

To update your branches, run (note the last command - you want to push to your fork so your PR updates - it might be origin or it might be some other remote name you gave it):

```bash
# Operator
cd <your kiali-operator local git repo>
git remote add jmazzitelli https://github.com/jmazzitelli/kiali-operator.git
git fetch jmazzitelli rolandkool-new_param_service_labels
git checkout new_param_service_labels
git reset --hard jmazzitelli/rolandkool-new_param_service_labels
git push --force-with-lease origin new_param_service_labels  # replace "origin" with whatever remote points to your fork

# Helm charts
cd <your kiali helm-charts local git repo>
git remote add jmazzitelli https://github.com/jmazzitelli/helm-charts.git
git fetch jmazzitelli rolandkool-new_param_service_labels
git checkout new_param_service_labels
git reset --hard jmazzitelli/rolandkool-new_param_service_labels
git push --force-with-lease origin new_param_service_labels  # replace "origin" with whatever remote points to your fork

jshaughn:
Parent issue closed.

PR #1071, opened 2026-06-29
If Kiali returns 401 for /api during OSSMC install, skip the version check

jmazzitelli:
When `server.require_auth` is set to `true` in the Kiali CR, the /api endpoint requires authentication. The operator calls this endpoint without credentials to check the Kiali version before deploying the OSSMC plugin. Previously this would cause the operator to retry for 5 minutes and then block OSSMC installation entirely.

This fix stops retrying immediately on a 401 response, logs a clear warning explaining that server.require_auth is likely set to true, and automatically skips the version check so that OSSMC installation can proceed. The fix is applied to the default, v2.22, and v2.27 roles.

Fixes: https://github.com/kiali/kiali/issues/9964
Generated-by: Cursor

Need to apply the fixes to the v2.22 and v2.27 playbooks/roles and backported to the v2.27 branch

PR #1072, opened 2026-06-29
[v2.27] If Kiali returns 401 for /api during OSSMC install, skip the version check

jmazzitelli:
(cherry pick of https://github.com/kiali/kiali-operator/pull/1071 )

When server.require_auth is set to true in the Kiali CR, the /api endpoint requires authentication. The operator calls this endpoint without credentials to check the Kiali version before deploying the OSSMC plugin. Previously this would cause the operator to retry for 5 minutes and then block OSSMC installation entirely.

This fix stops retrying immediately on a 401 response, logs a clear warning explaining that server.require_auth is likely set to true, and automatically skips the version check so that OSSMC installation can proceed. The fix is applied to the default, v2.22, and v2.27 roles.

Fixes: https://github.com/kiali/kiali/issues/9964
Generated-by: Cursor


jshaughn (approved):
Looks like a clean backport.

PR #1073, opened 2026-06-30
[CI] Fix race condition in set_kiali_cr task

pbajjuri20:
Add retry logic with 10 retries and 5 second delay to handle resource conflict errors (HTTP 409) when patching Kiali CR.

The operator can modify the CR between read and update operations, causing 'the object has been modified' errors. This retry mechanism resolves the race condition.

Fixes os-console-links-test intermittent failures.

### DIFF
--- a/molecule/common/set_kiali_cr.yml
+++ b/molecule/common/set_kiali_cr.yml
@@ -2,3 +2,7 @@
   k8s:
     namespace: '{{ cr_namespace }}'
     definition: "{{ new_kiali_cr }}"
+  retries: 10
+  delay: 5
+  register: kiali_cr_patch_result
+  until: kiali_cr_patch_result is succeeded


pbajjuri20:
Tested that single test case, it passed as expected.
Now running  all the test cases.

kiali/test-jobs/kiali-operator:#90 - SUCCESS
- OCP URL: https://api.user-rhos-d-3.servicemesh.rhqeaws.com:6443/
- Kiali branch: master
- Job URL: https://jenkins-csb-servicemesh-master.dno.corp.redhat.com/job/kiali/job/test-jobs/job/kiali-operator/90/
- Tests run: 1, Failures: 0, Skipped: 0

jmazzitelli (approved):
This LGTM but keep in mind this task is used for probably every molecule test - once merged, we need to keep an eye on the molecule test reports to make sure this doesn't have any odd side-effects.

PR #1075, opened 2026-07-02
Increase retry count for set_kiali_cr to handle longer reconciliation

pbajjuri20:
The existing retry logic (10 retries × 5s = 50s) was insufficient when the operator's reconciliation takes longer, causing os-console-links-test to fail with 409 Conflict errors.

This commit increases retries from 10 to 60 (5 minutes total) to allow enough time for the operator to complete reconciliation before the patch succeeds.

Fixes: 409 Conflict error 'Operation cannot be fulfilled on kialis.kiali.io'


ScriptingShrimp (approved):
lgtm

PR #1077, opened 2026-07-10
Strip resourceVersion from set_kiali_cr to fix 409 Conflict race

### LINKED ISSUE #10030: [Test] set_kiali_cr 409 Conflict race condition — external-auth-rotation-test, os-console-links-test / Jenkins nightly
### Describe the bug

Molecule test failures observed on Jenkins nightly. Both tests fail at the same Ansible task with identical error.

**Affected tests:**
- `external-auth-rotation-test`
- `os-console-links-test`

**Failing task:** `Change Kiali CR with new Kiali CR` (`molecule/common/set_kiali_cr.yml`)
**Classification:** test-bug
**Confidence:** high

**Error:**
```
Failed to patch object: Operation cannot be fulfilled on kialis.kiali.io "kiali":
the object has been modified; please apply your changes to the latest version and try again
(Conflict, 409)
```

All 60 retries (5 minutes) exhausted. Every retry sends the same stale `resourceVersion`, so 409 is guaranteed as long as the operator reconciler is active.

### Root cause

`molecule/common/set_kiali_cr.yml` uses the `k8s` Ansible module with `retries: 60` and `delay: 5`. On each retry, the module re-sends the same patch with the original `resourceVersion`. Because the operator continuously reconciles and updates the CR, the `resourceVersion` advances on every reconciliation cycle, and every retry hits 409.

**Prior fix attempts (insufficient):**
- PR #1073 — added retry logic (10 retries × 5s)
- PR #1075 — increased to 60 retries

Neither fixes the core issue: the task must **re-read the CR** to obtain a fresh `resourceVersion` before each patch attempt.

### Expected Behavior

The `set_kiali_cr` task should successfully patch the Kiali CR even when the operator is actively reconciling.

### What are the steps to reproduce this bug?

1. Ensure a minikube cluster suitable for molecule tests is running (profile `ci`)
2. Run the affected molecule tests:
   ```bash
   <kiali-git-repo>/hack/run-molecule-tests.sh \
     --client-exe "$(which kubectl)" \
     --cluster-type minikube \
     --minikube-profile ci \
     -udi true -hcrp false \
     -at "external-auth-rotation-test os-console-links-test"
   ```

### Environment

- **Kiali version:** unknown
- **Istio version:** not specified
- **Kubernetes impl:** OpenShift
- **OCP version:** 4.22.3
- **Build URL:** https://jenkins-csb-servicemesh-master.dno.corp.redhat.com/job/kiali/job/test-jobs/job/kiali-operator/103/

### Suggested fix

Replace the simple retry loop with a read-then-patch pattern:

1. Read the current Kiali CR (getting fresh `resourceVersion`)
2. Merge desired spec changes into the current CR
3. Apply the patch
4. On 409, go back to step 1

This can be implemented as an Ansible task block with `k8s_info` + `k8s` inside a `until` loop.

ScriptingShrimp:
## Summary

- Strip `metadata.resourceVersion` from `new_kiali_cr` before patching in `molecule/common/set_kiali_cr.yml`
- Prevents 409 Conflict when the operator reconciler modifies the CR between read and patch
- Reduces retries from 60 to 10 (409 was the sole reason for high retry count)

## Root cause

Callers build `new_kiali_cr` via `kiali_cr | combine(...)` where `kiali_cr` was read earlier and carries a stale `resourceVersion`. The `k8s` module sends this version on every retry attempt, and Kubernetes rejects all of them with 409 because the operator's reconciler continuously advances the version.

Prior fix attempts (PR #1073, #1075) added retries but didn't address the core issue — retrying the same stale version can never succeed while the reconciler is active.

## Fix

Strip `resourceVersion` from metadata before patching. Without it, Kubernetes applies the patch unconditionally (no optimistic locking). This is safe for test automation where the test owns the desired spec state.

Fixes https://github.com/kiali/kiali/issues/10030

## Test plan

- [ ] Verify `external-auth-rotation-test` passes on Jenkins nightly
- [ ] Verify `os-console-links-test` passes on Jenkins nightly
- [ ] Confirm no regressions in other molecule tests that use `set_kiali_cr.yml`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/molecule/common/set_kiali_cr.yml
+++ b/molecule/common/set_kiali_cr.yml
@@ -1,8 +1,13 @@
+# Callers build new_kiali_cr from a previously-read CR that includes a stale
+# metadata.resourceVersion. Stripping it lets the patch apply unconditionally
+# instead of racing with the operator reconciler (409 Conflict).
 - name: Change Kiali CR with new Kiali CR
+  vars:
+    cr_metadata_clean: "{{ new_kiali_cr.metadata | dict2items | rejectattr('key', 'equalto', 'resourceVersion') | list | items2dict }}"
   k8s:
     namespace: '{{ cr_namespace }}'
-    definition: "{{ new_kiali_cr }}"
-  retries: 60
+    definition: "{{ new_kiali_cr | combine({'metadata': cr_metadata_clean}) }}"
+  retries: 10
   delay: 5
   register: kiali_cr_patch_result
   until: kiali_cr_patch_result is succeeded


jmazzitelli (approved):
nice fix

jshaughn:
@ScriptingShrimp Should we backport to 2.27?

ScriptingShrimp:
> @ScriptingShrimp Should we backport to 2.27?

Yes, I think we should backport into all productized versions which will be regression tested in private pipelines. 

If you agree, I will create subtasks in original issue before closing it.

jshaughn:
Confirm with @jmazzitelli , but I think only 2.27 is needed as that is the only recommended Operator for all release versions.

ScriptingShrimp:
@jmazzitelli is backport into v2.27 sufficient?

jmazzitelli:
Yes, only 2.27 backport would be necessary.

jshaughn:
> Yes, only 2.27 backport would be necessary.

Just a note that 2.27 is in code-freeze until the z-stream goes out...

PR #1078, opened 2026-07-10
Add opt-in OpenShift impersonation mode for multi-cluster auth

### LINKED ISSUE #10038: OpenShift Impersonation for Multi-Cluster Auth
## Problem

Kiali's multi-cluster support on OpenShift currently requires users to individually authenticate to each remote cluster via the Kiali UI's user profile dropdown. Each cluster login triggers a separate OpenShift OAuth redirect flow. This design is acceptable for small deployments (2-3 clusters) but becomes completely impractical at scale. An organization with 10, 50, or 100+ clusters cannot reasonably ask users to manually log into each one — the experience is so laborious that it effectively renders multi-cluster Kiali unusable for large fleets.

The root cause is that OpenShift OAuth does not provide SSO across clusters. Each cluster has its own OAuth server, its own tokens, and its own session. Kiali faithfully models this by maintaining per-cluster session cookies and requiring per-cluster authentication.

## Scope: OpenShift Only

Impersonation mode is scoped to the openshift auth strategy. Non-OpenShift environments already solve the multi-cluster single-login problem via the openid strategy with a shared OIDC provider — a single OIDC token is valid across all clusters, so there is no per-cluster login burden. Impersonation addresses the gap specific to OpenShift, where each cluster has its own independent OAuth server and there is no built-in cross-cluster SSO.

## Proposal

Add an opt-in impersonation mode for the OpenShift auth strategy. When enabled, the user authenticates once to the home cluster (where Kiali is deployed) via the normal OpenShift OAuth flow. For all cluster API calls (including the home cluster), Kiali then uses its own Service Account credentials (already provisioned on remote clusters via the existing remote cluster secret mechanism) and attaches Kubernetes impersonation headers (Impersonate-User, Impersonate-Group) carrying the authenticated user's identity. The remote cluster's API server authenticates the Kiali SA, verifies it has permission to impersonate, and then evaluates RBAC against the impersonated user — preserving per-user access control without requiring per-user tokens.

This reduces the user login experience from O(n) to O(1) regardless of cluster count, while maintaining the same security boundary: each user can only see and do what their RBAC on each remote cluster permits.

The feature is:

* Opt-in via a new Kiali CR configuration field (`spec.auth.openshift.impersonation.enabled`)
* OpenShift-only (non-OpenShift environments already solve this via the OpenID strategy with a shared OIDC provider)
* Backward-compatible (existing per-cluster OAuth flow remains the default)
* Security-conscious (impersonation privileges are only granted when explicitly opted in; the impersonated identity is derived server-side from the verified home cluster session, never from user-supplied input)

Note: For architectural consistency, impersonation is applied to all clusters including the home cluster. The user's OAuth token is used only for the one-time identity verification; all subsequent API calls use SA + impersonation.

PRs:
* server:  https://github.com/kiali/kiali/pull/10039
* operator: https://github.com/kiali/kiali-operator/pull/1078
* helm-charts: https://github.com/kiali/helm-charts/pull/426
* kiali.io docs: https://github.com/kiali/kiali.io/pull/992

jmazzitelli:
Fixes: https://github.com/kiali/kiali/issues/10038

When spec.auth.openshift.impersonation.enabled is true, users authenticate once to the home cluster and Kiali uses its SA credentials with Kubernetes impersonation headers for all cluster API calls. This eliminates the need for per-cluster OAuth login flows that don't scale beyond a few clusters.

Generated-by: Cursor

### DIFF
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -42,6 +42,8 @@ spec:
         userinfo_endpoint: ""
     openshift:
       impersonation:
+        allowed_groups: []
+        allowed_users: []
         enabled: false
       #redirect_uris:
       #token_inactivity_timeout:
--- a/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
+++ b/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
@@ -41,6 +41,8 @@ spec:
         token_endpoint: ""
         userinfo_endpoint: ""
     openshift:
+      impersonation:
+        enabled: false
       #redirect_uris:
       #token_inactivity_timeout:
       #token_max_age:
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -211,6 +211,16 @@ spec:
                         description: "Settings for Kubernetes API impersonation. When enabled, users authenticate once to the home cluster and Kiali uses its service account to impersonate the user on all clusters (including home)."
                         type: object
                         properties:
+                          allowed_groups:
+                            description: "When set, only these groups will be included in impersonation headers (in addition to system:authenticated and system:authenticated:oauth which are always included). If empty, all of the user's groups are impersonated. This also restricts the Kiali SA's ClusterRole via resourceNames. Must not contain system: prefixed groups. For full RBAC defense-in-depth, configure both allowed_users and allowed_groups."
+                            type: array
+                            items:
+                              type: string
+                          allowed_users:
+                            description: "When set, only users in this list are permitted to use Kiali with impersonation. Users not in this list will be denied access (HTTP 403). This also restricts the Kiali SA's ClusterRole via resourceNames, limiting which identities the SA can impersonate even at the Kubernetes RBAC level. Must not contain system: prefixed identities. For full RBAC defense-in-depth, configure both allowed_users and allowed_groups."
+                            type: array
+                            items:
+                              type: string
                           enabled:
                             description: "When true, Kiali will use Kubernetes API impersonation for accessing all clusters instead of requiring per-cluster OAuth login. The user logs in once to the home cluster, and Kiali uses its service account credentials with impersonation headers carrying the user's identity. Requires the Kiali service account on each cluster to have ClusterRole permissions to impersonate users and groups. SECURITY NOTE: Enabling this grants the Kiali service account the ability to impersonate any user. Protect the service account token accordingly."
                             type: boolean
--- a/crd-docs/crd/kiali.io_kialis.yaml
+++ b/crd-docs/crd/kiali.io_kialis.yaml
@@ -207,6 +207,14 @@ spec:
                       client_id_prefix:
                         description: "DEPRECATED AFTER v1.73: A prefix that will be applied to the OpenShift OAuth client identifier."
                         type: string
+                      impersonation:
+                        description: "Settings for Kubernetes API impersonation. When enabled, users authenticate once to the home cluster and Kiali uses its service account to impersonate the user on all clusters (including home)."
+                        type: object
+                        properties:
+                          enabled:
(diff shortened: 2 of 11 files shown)


jshaughn (commented):
Visually looks fine to me. AI review comments seem valid:

The main flag is that the operator ClusterRole gets impersonate permission unconditionally (even when no Kiali CR enables the feature), which is by design due to Kubernetes RBAC escalation rules but will draw attention from security auditors. Two checklist files are missing: the example CR (crd-docs/cr/kiali.io_v1alpha1_kiali.yaml) and the comprehensive molecule CR (molecule/config-values-test/kiali-cr-all-yaml) both need the new auth.openshift.impersonation entry.

One more, minor:

The CRD description doesn't mention the OpenShift strategy requirement. It says:

"When true, Kiali will use Kubernetes API impersonation for accessing all clusters..."

It doesn't say "requires auth.strategy=openshift." Someone reading the CRD alone wouldn't know.

jmazzitelli:
> The main flag is that the operator ClusterRole gets impersonate permission unconditionally (even when no Kiali CR enables the feature), which is by design due to Kubernetes RBAC escalation rules

To be expected - like all other permissions, the operator has to have permissions in order to grant them. We don't know when/if Kiali CRs will use impersonation: true so the operator has to have it. It can't NOT have it or it simply will never be able to support this feature should a Kiali CR want to enable it.

> the example CR (crd-docs/cr/kiali.io_v1alpha1_kiali.yaml) and the comprehensive molecule CR (molecule/config-values-test/kiali-cr-all-yaml) both need the new auth.openshift.impersonation entry.

Done.

> The CRD description doesn't mention the OpenShift strategy requirement. It says: "When true, Kiali will use Kubernetes API impersonation for accessing all clusters..." It doesn't say "requires auth.strategy=openshift." Someone reading the CRD alone wouldn't know.

Ignoring. That setting is INSIDE the openshift auth section itself! So obviously, that setting is only for openshift auth - no need to explicitly say this (othewise, we'd want to say it for ALL the openshift auth settings). In fact, if we do this kind of notification, we would want to say this for EVERY specific auth setting (openid settings for example - none of them say "this requires auth.strategy=openid" -- it's simply expected and implied that the settings that are under that specific auth section require that auth strategy for it to take effect.

hhovsepy:
## Manual test results — Operator install (OpenShift multi-primary + impersonation)

Tested on two OCP 4.22.5 clusters (east home / west remote) with this operator PR image (`kiali-operator:dev`), helm chart from https://github.com/kiali/helm-charts/pull/426 (`supportImpersonation=true`), and server image from https://github.com/kiali/kiali/pull/10039.

Full UX/RBAC/allowlist testing was done earlier via Helm and reported on https://github.com/kiali/kiali/pull/10039#issuecomment-5093067663 (with screenshots). This comment covers the **Operator install path**.

### Setup
| Cluster | Operator | Kiali CR |
|---------|----------|----------|
| east | PR operator, `supportImpersonation=true`, `allowAdHocKialiImage=true` | Full server: `auth.openshift.impersonation.enabled=true`, custom server image |
| west | Same operator image | `deployment.remote_cluster_resources_only: true` + `impersonation.enabled=true` |

### Verified
| Check | Result |
|-------|--------|
| Operator reconciles east CR successfully | ✅ `Successful=True` |
| Operator reconciles west remote-only CR successfully | ✅ `Successful=True` |
| Operator ClusterRole has `impersonate` (via `supportImpersonation`) | ✅ |
| East ConfigMap / `/api/auth/info` → `impersonationEnabled: true` | ✅ |
| East discovers west (`clusters=2`, multi-cluster secret mounted) | ✅ |
| East + west server ClusterRoles have `impersonate` on users/groups | ✅ |
| West: no Kiali Deployment, no OAuthClient | ✅ |

Related: https://github.com/kiali/kiali/pull/10039 · https://github.com/kiali/helm-charts/pull/426 · https://github.com/kiali/kiali.io/pull/992

jmazzitelli:
@hhovsepy so all your tests are good? Can you click the "approve" link?

hhovsepy (approved):

jshaughn (approved):
I don't see anything blocking this option, it seems valid and potentially useful for users that want to use it.

PR #1082, opened 2026-07-29
Support installing OSSM Console without a Kiali server present ("lite mode")

jmazzitelli:
The operator now allows the OSSMConsole CR to reconcile successfully even when no Kiali installation can be found, installing the plugin without Kiali integration instead of failing the deployment. If the user explicitly configures any of spec.kiali.serviceName/serviceNamespace/ servicePort, that is treated as a firm requirement and reconciliation still hard-fails when no matching Kiali service can be found or its version does not match, since that indicates a misconfiguration rather than an absent Kiali. Status reporting gains a kiali.available field, the ConsolePlugin's proxy entry is omitted when Kiali is unavailable, and a new ossmconsole-lite-test molecule scenario covers both the lite install and the hard-failure path.

related to: https://github.com/kiali/openshift-servicemesh-plugin/pull/789

Generated-by: Cursor

This will require a change to the kiali repo's `run-molecule-tests.sh` script so non-OpenShift clusters skip this new test:

```diff
diff --git a/hack/run-molecule-tests.sh b/hack/run-molecule-tests.sh
index df7752359..2725ce715 100755
--- a/hack/run-molecule-tests.sh
+++ b/hack/run-molecule-tests.sh
@@ -181,9 +181,9 @@ ALL_TESTS=${ALL_TESTS:-$(cd "${KIALI_SRC_HOME}/operator/molecule" || exit; ls -d
 if [ "${CLUSTER_TYPE}" == "openshift" ]; then
   SKIP_TESTS="${SKIP_TESTS:-header-auth-test openid-test}"
 elif [ "${CLUSTER_TYPE}" == "minikube" ]; then
-  SKIP_TESTS="${SKIP_TESTS:-openid-openshift-test os-console-links-test ossmconsole-config-values-test tls-profile-test}"
+  SKIP_TESTS="${SKIP_TESTS:-openid-openshift-test os-console-links-test ossmconsole-config-values-test ossmconsole-lite-test tls-profile-test}"
 elif [ "${CLUSTER_TYPE}" == "kind" ]; then
-  SKIP_TESTS="${SKIP_TESTS:-openid-openshift-test os-console-links-test ossmconsole-config-values-test tls-profile-test}"
+  SKIP_TESTS="${SKIP_TESTS:-openid-openshift-test os-console-links-test ossmconsole-config-values-test ossmconsole-lite-test tls-profile-test}"
 fi
```

That change and another minor change in helm chart repo is needed if we elect to do this (very minor - CRD description changes in helm charts repo as well as that small change to the run-molecule-tests.sh script in server repo). I didn't create PRs over there (server and helm) to avoid the noise if we don't want this - those missing PRs are why the CI tests are red.

* server: https://github.com/kiali/kiali/compare/master...jmazzitelli:ossmc-lite?expand=1
* helm-charts: https://github.com/kiali/helm-charts/compare/master...jmazzitelli:ossmc-lite?expand=1


jmazzitelli:
as of august 14, 12:30pm EDT this PR is ready to go, barring additional reviews/request-for-changes from anyone else.

jmazzitelli:
molecule tests pass:
```
=====================
=== TEST RESULTS: ===
=====================

                     ossmconsole-default... success [2m 55s]
          ossmconsole-config-values-test... success [3m 24s]
                   ossmconsole-lite-test... success [8m 44s]
```

crd-docs/crd/kiali.io_ossmconsoles.yaml
```diff
@@ -91,15 +91,22 @@ spec:
...
+                description: "Unstructured section for internal testing and debugging features. Not officially supported - fields here may change or be removed without notice."
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
               kiali:
                 type: object
                 properties:
+                  autoDiscover:
+                    description: "Controls whether the operator will opportunistically auto-discover a Kiali installation via its OpenShift Route when none of `kiali.serviceName`, `kiali.serviceNamespace`, and `kiali.servicePort` are set. If those three settings are all empty and this is set to `false`, OSSM Console is installed in lite mode (without Kiali integration) rather than attempting discovery. This setting has no effect if all three of `kiali.serviceName`, `kiali.serviceNamespace`, and `kiali.servicePort` are explicitly set (auto-discovery is already skipped in that case, and the explicitly configured Kiali installation is used). Setting this to `false` while only some (not all, and not none) of those three settings are specified is invalid and will fail reconciliation. Defaults to `true`."
```
jshaughn:
```suggestion
                    description: "Controls whether the operator will opportunistically auto-discover a Kiali installation via its OpenShift Route when none of `kiali.serviceName`, `kiali.serviceNamespace`, and `kiali.servicePort` are set. If those three settings are all empty and this is set to `false`, OSSM Console is installed without Kiali integration, rather than attempting discovery. This setting has no effect if all three of `kiali.serviceName`, `kiali.serviceNamespace`, and `kiali.servicePort` are explicitly set (auto-discovery is already skipped in that case, and the explicitly configured Kiali installation is used). Setting this to `false` while only some (not all, and not none) of those three settings are specified is invalid and will fail reconciliation. Defaults to `true`."
```

again, can we nix "lite"?  Here we may just be able to omit it. Or, if we can come up with a term that doesn't make it sound like a different version/flavor of the plugin. "detached", "disconnected", "unconfigured"...

jshaughn: This applies to several places in the PR...

jmazzitelli: see latest commit for lite-removal changes

jshaughn (approved):
LG, the "lite" terminology has been removed from the user-facing descriptions.

PR #1086, opened 2026-08-04
[molecule] set impersonate setting so test can pass

jmazzitelli:
should have been part of the [impersonation work](https://github.com/kiali/kiali-operator/pull/1078/)


jshaughn (dismissed):
I blame Mary.

PR #1087, opened 2026-08-05
[molecule] Fix OIDC readiness race in openid-openshift-test

jmazzitelli:
Harden the openid-openshift-test molecule test to protect against race condition that causes flaky test failures.

The kube-apiserver can report Available via ClusterOperator before its OIDC authenticator is fully initialized, causing a transient 401 when Kiali exchanges the authorization code. This replaces the arbitrary 30s pause with a check that all apiserver nodes are at the latest static pod revision, and wraps the OIDC handshaking flow in block/rescue retries (up to 3 attempts) so the real auth flow rides out any remaining initialization window.

Generated-by: Cursor

### DIFF
--- a/molecule/openid-openshift-test/converge.yml
+++ b/molecule/openid-openshift-test/converge.yml
@@ -246,8 +246,29 @@
     debug:
       msg: "Testing OAuth2 authentication flow using auto-discovery (/.well-known/openid-configuration)"
 
+  # The OIDC handshaking is wrapped in block/rescue to handle a known race
+  # condition: after the kube-apiserver rolls out with OIDC config, there is a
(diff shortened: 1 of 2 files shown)


jmazzitelli:
Start CRC/OpenShift then install Istio init:
```
./hack/crc-openshift.sh start -he true -p <your pull secret>
./hack/istio/install-istio-via-istioctl.sh -c oc
```

Then run molecule test:
```
./hack/run-molecule-tests.sh --client-exe "$(which oc)" --cluster-type openshift -at openid-openshift-test
```

See results:
```
=====================
=== TEST RESULTS: ===
=====================

                   openid-openshift-test... success [11m 36s]
```

PR #1088, opened 2026-08-11
Mark CR spec values as unsafe to prevent Jinja2 template injection

jmazzitelli:
Add markUnsafe: True to every watch entry in all watches files. Without this setting, the ansible-operator base image defaults markUnsafe to false, passing CR spec string values to Ansible playbooks without the __ansible_unsafe wrapper. This means Jinja2 template expressions in CR spec fields would be evaluated inside the operator pod, potentially allowing code execution as the operator ServiceAccount.

The kiali-operator playbooks do not intentionally evaluate CR spec values as Jinja2 templates, so this change is safe.

### DIFF
--- a/watches-k8s-ns.yaml
+++ b/watches-k8s-ns.yaml
@@ -5,6 +5,7 @@
   group: kiali.io
   kind: Kiali
   playbook: playbooks/kiali-deploy.yml
+  markUnsafe: True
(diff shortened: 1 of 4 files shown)


nrfox (approved):

jshaughn:
This should be backported to v2.27.

PR #1092, opened 2026-08-25
revert timestamp from old metadata csv

jmazzitelli:
Trivial change.

We need to revert this timestamp, otherwise the operatorhub.io metadata gets screwed up and our operator hub update PRs won't be mergeable.

This is due to a bad PR merge we did in the past.

The real timestamp was a few weeks earlier (2026-07-12T09:23:43Z); this has no practical effects if we revert this.


jmazzitelli:
Ignore the CI build failure - it correctly identified that we are changing a past version metadata, but this is on purpose.

jshaughn (commented):
LOK

jshaughn (approved):

PR #1093, opened 2026-08-26
[AI] Refactor configuration and allowed_users

aljesusg:
Refactor configuration and allowed_users

https://redhat.atlassian.net/browse/OSSM-15741
https://redhat.atlassian.net/browse/OSSM-15742

UI+Back: https://github.com/kiali/kiali/pull/10256
Helm: https://github.com/kiali/helm-charts/pull/434

### DIFF
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -17,7 +17,7 @@ kiali_defaults:
     icon_annotation: "kiali.io/api-type"
 
   ai:
-    chat_ai:
+    chat:
       allowed_users: []
       default_provider: ""
       enabled: false
--- a/roles/default/kiali-deploy/defaults/main.yml
+++ b/roles/default/kiali-deploy/defaults/main.yml
@@ -16,6 +16,22 @@ kiali_defaults:
     annotation: "kiali.io/api-spec"
     icon_annotation: "kiali.io/api-type"
 
+  ai:
+    chat_ai:
+      allowed_users: []
+      default_provider: ""
+      enabled: false
+      max_tool_iterations: 5
+      providers: []
+      store_config:
+        enabled: true
+        inactivity_timeout: "30m"
+        max_cache_memory_mb: 1024
+        reduce_threshold: 15
+        reduce_with_ai: false
+      tools: {}
+    enabled: false
+
   auth:
     openid:
       additional_request_params: {}
@@ -50,19 +66,6 @@ kiali_defaults:
       #token_inactivity_timeout:
       #token_max_age:
     strategy: ""
-
-  chat_ai:
-    default_provider: ""
-    enabled: false
-    max_tool_iterations: 5
-    providers: []
-    store_config:
-      enabled: true
-      inactivity_timeout: "30m"
-      max_cache_memory_mb: 1024
-      reduce_threshold: 15
-      reduce_with_ai: false
-    tools: {}
 
   clustering:
     autodetect_secrets:
--- a/roles/default/kiali-deploy/tasks/main.yml
+++ b/roles/default/kiali-deploy/tasks/main.yml
@@ -824,7 +824,7 @@
 
       {# Prepare the secret volumes for ai.chat provider and model keys #}
       {# Note: provider.enabled and model.enabled default to false if not specified #}
-      {% if kiali_vars is defined and kiali_vars.ai is defined and kiali_vars.ai.chat is defined and kiali_vars.ai.chat.enabled is defined and kiali_vars.ai.chat.enabled|bool and kiali_vars.ai.chat.providers is defined and kiali_vars.ai.chat.providers|length > 0 %}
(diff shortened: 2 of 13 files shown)


roles/default/kiali-deploy/defaults/main.yml
jmazzitelli: you can probably delete this now from here - no need to set the defaults for the old obsolete chat_ai fields. These aren't used by the operator and if the server never looks at these chat_ai fields anymore, no need for them to go into the ConfigMap.

If the server DOES look at these settings (perhaps for backwards compatibility?) then leave these here.

aljesusg: As we discussed, if there's any config in chat_ai, I'm moving it to ai in the server config.

jmazzitelli: Yes, I see over in the server PR all references to the config items are now referring to this main AI section. So we should remove lines 70 thru 81 - they are obsolete now and can be deleted.

Leave the CRD settings as you already have them - those are fine. In your CRD schema description, you mentioned the DEPRECATED note which is correct. Those settings are just now going to be ignored totally.

jmazzitelli (changes requested):
There are two things that need to change.

1. There are molecule tests which test these configuration settings. The molecule tests need to be updated. See [github query here](https://github.com/search?q=repo%3Akiali%2Fkiali-operator+chat_ai+path%3Amolecule&type=code).

To test that the molecule tests pass, you can run this (but once you put these files in your PR, the GitHub CI Actions should run the molecule tests for you and give you green  when they pass):

(BTW: just tell Cursor or Claude to read AGENTS.md in kiali/kiali repo and tell it to "run the molecule tests config-values-test and external-auth-rotation-test" and it will do it for you ;)

From kiali/kiali repo:
```bash
./hack/k8s-minikube.sh --hydra-enabled true start

make CLUSTER_TYPE=minikube build-ui build test cluster-push

./hack/run-molecule-tests.sh --client-exe "$(which kubectl)" --cluster-type minikube -hcrp false -udi true -at "config-values-test external-auth-rotation-test"
```

2. There are also AI agent docs that need to be updated:

* `docs/agents/crd-and-api.md`:| `chat_ai` | AI assistant integration (providers, models, caching) |
* `docs/agents/operator-architecture.md`:- `chat_ai.providers[*].key` and `chat_ai.providers[*].models[*].key` (only when `enabled: true`)

jmazzitelli (changes requested):
There is a third thing that needs to be updated... the example CR which is used to produce the kiali.io docs:

This has to be moved/renamed:

https://github.com/kiali/kiali-operator/blob/master/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L54-L65

roles/default/kiali-deploy/tasks/main.yml
```diff
@@ -822,10 +822,10 @@
       {% set d = d | combine({'customdashboards-prometheus-oauth2-client-secret': {'secret_name': kiali_vars.external_services.custom_dashboards.prometheus.auth.oauth2.client_secret | regex_replace('secret:(.+):.+', '\\1'), 'secret_key': kiali_vars.external_services.custom_dashboards.prometheus.auth.oauth2.client_secret | regex_replace('secret:.+:(.+)', '\\1') }}) %}
       {% endif %}
 
-      {# Prepare the secret volumes for chat_ai provider and model keys #}
+      {# Prepare the secret volumes for ai.chat provider and model keys #}
       {# Note: provider.enabled and model.enabled default to false if not specified #}
-      {% if kiali_vars is defined and kiali_vars.chat_ai is defined and kiali_vars.chat_ai.enabled is defined and kiali_vars.chat_ai.enabled|bool and kiali_vars.chat_ai.providers is defined and kiali_vars.chat_ai.providers|length > 0 %}
-      {% for provider in kiali_vars.chat_ai.providers %}
+      {% if kiali_vars is defined and kiali_vars.ai is defined and kiali_vars.ai.chat is defined and kiali_vars.ai.chat.enabled is defined and kiali_vars.ai.chat.enabled|bool and kiali_vars.ai.chat.providers is defined and kiali_vars.ai.chat.providers|length > 0 %}
```
jmazzitelli: The condition here checks `kiali_vars.ai.chat.enabled` but does not check the parent `kiali_vars.ai.enabled` flag. This means that if someone sets `ai.enabled: false` (intending to disable the entire AI feature) but leaves `ai.chat.enabled: true`, the operator will still mount chat AI provider/model secrets into the pod, which is not what we want.

If `ai.enabled` is meant to be a top-level gate that suppresses all AI behavior (and I think that is true), then this condition should also check `kiali_vars.ai.enabled|bool`:

```suggestion
      {% if kiali_vars is defined and kiali_vars.ai is defined and kiali_vars.ai.enabled is defined and kiali_vars.ai.enabled|bool and kiali_vars.ai.chat is defined and kiali_vars.ai.chat.enabled is defined and kiali_vars.ai.chat.enabled|bool and kiali_vars.ai.chat.providers is defined and kiali_vars.ai.chat.providers|length > 0 %}
```

jmazzitelli (changes requested):
The recent changes are all good, and the molecule tests pass.

However, I found a bug that needs to be fixed in the changes made in the tasks/main.yml.

jmazzitelli (approved):
Approve. Wait for the molecule tests CI job to finish - if it is green, this is good to merge.

